PR #2, opened 2015-04-18
Swift 1.2

1ec5:
Includes upgrading to SwiftyJSON 2.2.0.

/cc @incanus


1ec5:
I’m leaning towards `internal(set) var`, which would keep access to the setters internal to `MBRouteStep` rather than exposing them to the entire app’s module.

incanus:
Cool, yeah I ran with that as well. I also cleaned up some stuff around optionals some more and built out a basic playground that we can use to run through a simple use case. We can build it up from there. I symlinked the requisite files (`SwiftyJSON.swift` and `MapboxDirections.swift`) into the playground, which gets a little weird during dev, but works fine for demo purposes.

incanus:
Low priority right now, but we can make a framework (https://github.com/mapbox/MapboxDirections.swift/issues/4) to bring back Objective-C support.

PR #22, opened 2016-03-03
Add Objective-C support and example; reorganize project [WIP]

friedbunny:
**Changes**
- Adds `NSObject` to every public class in MapboxDirections.swift.
- Adds a new ObjC target to the Xcode project and moves the original Swift example to its own target.
- Adds a MapboxDirections.framework target that the two example targets bring in.
- Reorganizes the repo — moves MapboxDirections.swift to the MBDirections folder.
- Asserts that the access token has been changed (if not properly set).

**Issues**
- The initializer for `MBDirectionsRequest` will not automatically bridge to ObjC.
 -  `waypointCoordinates: [CLLocationCoordinate2D]` is the problem.
 - Temporarily created another initializer that ignores waypoints.
 - ... but this new initializer needs to be differently named, so that the compiler won't complain about ambiguity when you leave out a property when you're initializing in Swift.
- Logging output to console is meh, but acceptable for now.
- Tests need to be updated for the new project structure.

Will need to solve the initializer bridging problem and minimize disruption before this can be merged.

/cc @1ec5

### DIFF
--- a/Directions Example.xcodeproj/project.pbxproj
+++ b/Directions Example.xcodeproj/project.pbxproj
@@ -8,13 +8,37 @@
 
 /* Begin PBXBuildFile section */
 		6A0C780296F2AD5A7682A87D /* Pods_Unit_Tests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 371BFE7855E3EA8B44CA2B5F /* Pods_Unit_Tests.framework */; };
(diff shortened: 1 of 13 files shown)


1ec5:
Given that we expect keep the API surface area of this library fairly minimal, maybe we should have two language-specific libraries, to get around issues like the waypoint parameter.

In the meantime, though, the solution to this ambiguity would be to remove the default value from the waypoint parameter.

friedbunny:
Putting this particular attempt aside, for now.

PR #23, opened 2016-03-24
Directions API v5

### LINKED ISSUE #4: create framework
We should create a `MapboxDirections.framework` that can be used by Objective-C apps as well.

### LINKED ISSUE #9: add tests

1ec5:
Rewrote most of the implementation for compatibility with Mapbox Directions API v5. The primary differences are an intermediate MBRouteLeg object that contains any number of MBRouteSteps, more granular properties for route steps, and geometries segmented by step. As before, I’ve kept the API as close to MapKit API as possible. In fact, that meant renaming a couple of properties that previously differed.

Along the way, I also refactored the response logic and unwrapped a number of previously optional properties. It’s reasonable for this library to assert or crash if the server returns a malformed response.

If we want to support the v4 API concurrently with the v5 API, we could maintain it on a separate api-v4 branch, risking a loss of parity. But I’d like to investigate writing shims for the v4 API and turning the public-facing classes into umbrella classes. Even though this approach would incur some maintenance overhead, it’ll set us up well for the future when we might have more variety in the endpoints we have to support. This approach only makes sense if we intend to continue to make the API look like MapKit’s directions API.

* [x] Reformulate MBRoute
* [x] Reformulate MBRouteStep
* [x] Rewrite request logic based on RequestKit
* [x] Expose API to choose REST API version
* [x] Shim for v4 MBRoute
* [x] Shim for v4 MBRouteStep
* [x] Add podspec
* [x] Convert to framework
* [x] Added basic functional tests

Fixes #4 and fixes #9.

/cc @friedbunny @bsudekum @zugaldia


zugaldia:
> Even though this approach would incur some maintenance overhead, it’ll set us up well for the future when we might have more variety in the endpoints we have to support.

I'm also investigating that possibility. For now, I'm leaning towards a separate Java package (think if namespace) inside the submodule to support different versions of APIs, not just directions.

1ec5:
Now that the server response distinguishes between a _maneuver type_ and a _maneuver direction_, we should create an MBManeuver struct type (or perhaps class for Objective-C compatibility) to coordinate all the aspects of a maneuver, as opposed to overall step data like duration and distance.

1ec5:
I rewrote the HTTP request logic based on the excellent RequestKit. The full set of v4 and v5 options are supported, but so far no additional options have been exposed in the public API, and there isn’t a way to choose the v4 endpoint.

I did have to make one modification to RequestKit, which is to expose the NSURLSessionDataTask that gets created, so that `MBDirections.cancel()` works. So this PR depends on nerdishbynature/RequestKit#14. I’ve pointed the branch to a forked podspec for now.

1ec5:
The v4 response format is once again supported via a modicum of translation code. RequestKit was particularly handy for keeping the two request formats straight, while some switch statements were unavoidable in the response parsing code. I considered adopting a class cluster architecture, but it’s [rather more involved](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20151207/000925.html) in Swift than in Objective-C.

This PR also adds a podspec so that the library and its dependency, RequestKit, can be installed via CocoaPods. For now, while nerdishbynature/RequestKit#14 is outstanding, I’ve avoided qualifying the dependency on any particular version of RequestKit. For now, downstream users need to explicitly override the RequestKit dependency in their podfiles to specify the patched RequestKit:

```podspec
pod 'NBNRequestKit', :podspec => 'https://raw.githubusercontent.com/1ec5/RequestKit/mapbox-podspec/NBNRequestKit.podspec'
```

Once this PR lands, I need to undo 9449b865890a0227909da3e52fc95f2fb4fd7a73.

@friedbunny, can you review?

friedbunny:
Eyeballing now.

MBDirectionsRouter.swift
```diff
@@ -0,0 +1,131 @@
...
+                params["steps"] = String(includeSteps)
+            }
+            return params
+            
+        case .V5(let config, _, let waypoints, let includeAlternative, let geometryFormat, let overviewGranularity, let includeSteps, let allowPointUTurns):
+            var params: [String: String] = ["access_token": config.accessToken!]
+            if let includeAlternative = includeAlternative {
+                params["alternative"] = String(includeAlternative)
+            }
+            let hasHeadings = !waypoints.flatMap { $0.heading }.isEmpty
```
friedbunny: Took me a second to see that the exclamation point is negating the `.isEmpty` — can we remove the space before the first curly brace in these cases?

MapboxDirections.swift
```diff
@@ -21,106 +21,197 @@ public class MBPoint {
 
 }
 
+private func CLLocationCoordinate2DFromJSONArray(array: [Double]) -> CLLocationCoordinate2D? {
+    guard array.count == 2 else {
+        return nil
```
friedbunny: Can the API give us bad coordinates?

MapboxDirections.swift
```diff
@@ -129,47 +220,52 @@ public class MBRoute {
...
-        if let jsonGeometry = json["geometry"] as? JSON, coordinates = jsonGeometry["coordinates"] as? [[Double]] {
-            geometry = coordinates.map {
-                CLLocationCoordinate2D(latitude: $0[1], longitude: $0[0])
+        
+        switch version {
+        case .Four:
+            legs = [MBRouteLeg(source: source, destination: destination, json: json, profileIdentifier: profileIdentifier, version: version)]
+        case .Five:
+            // Associate each leg JSON with an origin and destination. The sequence
+            // of destinations is offset by one from the sequence of origins.
```
friedbunny: Thanks for this comment, heh.

MapboxDirections.swift
1ec5: It shouldn’t, so I made this an assertion instead. At least until the v5 endpoint stabilizes, it’s better for us to fail early and predictably than to fail gracefully in the face of a malformed response.

MBDirectionsRouter.swift
1ec5: I added parentheses to make the intention clear. It may look strange, but I use a space here for consistency with the case where a closure appears on multiple lines.

1ec5:
Aaaand we have a real functional test finally! I swapped out OHHTTPStubs in favor of Nocilla and implemented a basic test of `MBDirections.calculateDirectionsWithCompletionHandler(_:)` against the v4 API. (We’ll add v5 tests once the v5 endpoint’s profiles are live.) I also moved the Swift files into a proper framework.

PR #30, opened 2016-04-06
Switch to Alamofire. Fixes #29 [skip ci]

### LINKED ISSUE #29: Switch to Alamofire for requests
It's not possible to cleanly set a user agent header with RequestKit, and there are likely other methods that we'll need down the line, so let's switch to the battle-tested Alamofire.

tmcw:


tmcw:
It passes tests! @1ec5 care to review?

Main point of "huh" is the working around dispatch_sync https://github.com/mapbox/MapboxDirections.swift/pull/30/files#diff-e76be193f7c380a503afe0e47c1cd829L132 - keeping these calls in would hang the main thread, and _think_ that's because Alamofire is _already on_ the main thread so maybe like it's waiting for itself? Really I have no idea. But I removed the calls willy-nilly and am not sure if it works, only sure that it makes tests pass.

MapboxDirections/MBDirections.swift
```diff
@@ -36,10 +37,7 @@ public class MBDirections: NSObject {
 
     private let request: MBDirectionsRequest
     private let configuration: MBDirectionsConfiguration
-    private var task: NSURLSessionDataTask?
-    public var calculating: Bool {
```
1ec5: We should keep this property for compatibility with MapKit. It's how client code knows that the object is already waiting for a request. Perhaps not the most elegant approach, but we should be deliberate about the places we diverge from MapKit.

MapboxDirections/MBDirectionsRouter.swift
```diff
@@ -48,83 +49,65 @@ internal enum MBDirectionsRouter: Router {
...
-                params["alternative"] = String(includeAlternative)
-            }
-            let hasHeadings = !(waypoints.flatMap { $0.heading }.isEmpty)
-            if hasHeadings {
-                params["bearings"] = waypoints.map {
-                    return $0.heading != nil ? "\($0.heading!)" : ""
-                }.joinWithSeparator(";")
+    var URLRequest: NSMutableURLRequest {
+        let result: (configuration: MBDirectionsConfiguration, path: String, parameters: [String: AnyObject]) = {
+            switch self {
```
1ec5: Instead of this massive assignment block, declare `result` with `let` but don't assign anything to it on that line. Then write to the variable in every case of a switch statement before ever reading from the variable.

1ec5: The use of a tuple here also strikes me as a JavaScriptism, the equivalent of a `var` list. Instead, declare multiple `let` variables.

MapboxDirections/MBDirectionsRouter.swift
```diff
@@ -48,83 +49,65 @@ internal enum MBDirectionsRouter: Router {
...
-    }
-    
-    var configuration: Configuration {
-        switch self {
-        case .V4(let config, _, _, _, _, _, _): return config
-        case .V5(let config, _, _, _, _, _, _, _): return config
-        }
-    }
+    case V4(MBDirectionsConfiguration, String, [MBDirectionsWaypoint], Bool?, InstructionFormat?, GeometryFormatV4?, Bool?)
+    case V5(MBDirectionsConfiguration, String, [MBDirectionsWaypoint], Bool?, GeometryFormatV5?, OverviewGranularity?, Bool?, Bool?)
```
1ec5: I'm not wedded to the original approach of structuring the router around a v4/v5 enum and switch statement. The lack of named parameters was very tricky as I originally wrote this, because the types are totally divorced from the names, much as they might be in idiomatic C code.

MapboxDirectionsTests/MapboxDirectionsTests.swift
```diff
@@ -26,29 +26,29 @@ class MapboxDirectionsTests: XCTestCase {
...
+            XCTAssertEqual(routes.first!.legs.first!.steps.count, 136)
+            XCTAssertEqual(routes.first!.legs.first!.steps[24].distance, 106_258)
+            XCTAssertEqual(routes.first!.legs.first!.steps[24].duration, 3_929)
+            XCTAssertEqual(routes.first!.legs.first!.steps[24].name, "I 80")
+            
             expectation.fulfill()
         }
-        waitForExpectationsWithTimeout(2) { (error) in
+        
+        waitForExpectationsWithTimeout(10) { error in
```
1ec5: Does the timeout need to be 10 seconds long? If so, it might be possible to reduce the time by moving the assertions back out of the completion handler.

1ec5:
From a quick glance at the Alamofire codebase, I think you're right that it already handles dispatching to the main queue. RequestKit did not, so we had to dispatch to the main queue ourselves; otherwise, client code would end up executing on the URL session delegate queue.

MapboxDirections/MBDirectionsRouter.swift
tmcw: This was taken from the [alamofire example for this pattern](https://github.com/Alamofire/Alamofire#api-parameter-abstraction), so I assumed it was idiomatic

1ec5: Sort of? It doesn’t scale well, in any case.

MapboxDirections/MBDirectionsRouter.swift
tmcw: Yeah, it's odd to switch v4/v5 everywhere - it seems weird to support more than one api version at a time, or if we do, we could treat them as separate services with totally different code paths?

1ec5: Right now both versions need to be supported until the three default profiles are available for v5. But I think we could also find ourselves in a situation down the road where multiple endpoints need to be supported simultaneously. Originally, I wanted to keep completely separate code paths but expose them as one unified API using class clusters. Unfortunately, Swift currently lacks language support for Objective-C-style class clusters.

MapboxDirections/MBDirections.swift
tmcw: Per voice, given the async'ness of requests this bit of compatibility with MapKit isn't worthwhile: our directions SDK doesn't have any good reason will permit concurrent requests and any kind of serialization would happen on the app level.

tmcw:
:+1: got rid of the big switch statement

I'm going to do the user-agent header addition in this branch, but feel like a v4/v5 strategy refactor is better left for another day, since it's pretty involved and also would diverge from the original API even more.

tmcw:
Testing for the User-Agent header is tricky because Nocilla doesn't support expectations on requests: https://github.com/luisobo/Nocilla/issues/4

1ec5:
I switched from the `OHHTTPStubs/Swift` pod to Nocilla because the latter seemed to be cleaner and more idiomatic, but if the former has the functionality you're looking for, we could switch back relatively painlessly.

1ec5:
I ended up switching back to OHHTTPStubs/Swift in #47, along with rewriting the library in a way that addresses the weirdness above but also makes it more difficult to integrate Alamofire into the library.

1ec5:
Closing per https://github.com/mapbox/MapboxDirections.swift/issues/29#issuecomment-232201726.

PR #35, opened 2016-04-13
Convert Podfile to individual targets for CocoaPods 1.0.0

friedbunny:
Moves the Podfile to individually named targets, as defined in the Xcode project. This is required for [CocoaPods 1.0.0](http://blog.cocoapods.org/CocoaPods-1.0/).

/cc @1ec5

### DIFF
--- a/MapboxDirections.xcodeproj/project.pbxproj
+++ b/MapboxDirections.xcodeproj/project.pbxproj
@@ -7,10 +7,11 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


friedbunny:
When CocoaPods 1.0.0 actually drops, we’ll want to switch to the new [`abstract_target`](https://guides.cocoapods.org/syntax/podfile.html#abstract_target) (instead of `shared_pods`).

```ruby
platform :ios, '8.0'
use_frameworks!

abstract_target 'Directions' do
  pod 'NBNRequestKit', :git => 'https://github.com/1ec5/RequestKit.git', :branch => 'mapbox-podspec'
  pod 'Polyline', '~> 3.0'

  target 'MapboxDirections'
  target 'Directions Example'

  target 'MapboxDirectionsTests' do
    pod 'Nocilla'
  end
end
```

PR #40, opened 2016-04-29
Update v5 API

1ec5:
This PR reacts to various changes to the v5 API made over the past few weeks:

* Updated `ManeuverType` values.
* A request has only one transport type, corresponding to the profile, but for every request, each step may have a different, more granular transport type. So now an `MBRouteStep`’s `transportType` has a distinct datatype. The `transportType` property matters particularly when the `maneuverType` is `HeedWarning`.
* Updated v5 URL parameters.
* v5 route legs currently lack route summaries, so here I’m synthesizing them using logic similar to what would be done on the server side. This code can be very expensive, so I hope to remove it as soon as possible.

/cc @karenzshea @bsudekum @freenerd


1ec5:
97378caaf86fcff92be21051d9fa4e8da10021d9 follows up by inverting the logic for when to synthesize a route name.

Issue #20, opened 2016-02-18
Send appropriate user agent with directions requests


friedbunny:
As with https://github.com/mapbox/MapboxGeocoder.swift/issues/25, we should provide an identifiable user agent with directions API requests.

/cc @1ec5

PR #47, opened 2016-05-31
Rewrite the library with full Objective-C support and API documentation

### LINKED ISSUE #41: taskWithRouter causes errors
When attempting to incorporate the Directions files into my application, the error

 ``` Cannot convert call result type '()' to expected type 'NSURLSessionDataTask?' ```

caused by 

``return router.loadJSON(JSON.self) { [weak self] (json, error) in```... in MBDirections.swift

halts compilation of the application when using the example code. I've built the example app provided and ran the tests, which all pass.

### LINKED ISSUE #43: Add test cases for v5 API
Now that the v5 API is publicly accessible, we should add some tests and test fixtures for it.

/cc @tmcw

### LINKED ISSUE #44: Read access token from Info.plist
As in mapbox/MapboxStatic.swift#19 and MapboxGeocoder.swift#40, the access token should be optional. When no access token is explicitly passed into MBDirections’ initializer, this library should get it from the `MGLMapboxAccessToken` key in the main bundle’s Info.plist, just like the Mapbox iOS SDK does.

/cc @friedbunny

1ec5:
This PR rewrites the entire library to fully bridge to Objective-C while feeling more natural in Swift, handle errors more robustly, and provide complete API documentation. This PR, which closely parallels mapbox/MapboxGeocoder.swift#41 and mapbox/MapboxStatic.swift#19, is the last major step before consolidating the three libraries into one library that can eventually connect to all of Mapbox’s server APIs.

The goal is no longer to be a (poor) drop-in replacement for MapKit’s MKDirections API but rather to be as idiomatic as possible in Swift and Objective-C simultaneously, with a preference for Swift. Symbol names generally conform to platform and language conventions, but in some cases, symbol names eschew MapKit terminology in favor of Directions API terminology, so that api-documentation makes sense in the context of this library. The `MB` class prefix has been removed from Swift but is kept for Objective-C.

Directions API v4 support has been reimplemented as a series of subclasses of the options and model objects. The previous approach of inserting switch statements everywhere proved less than ideal. Use `RouteOptionsV4` instead of `RouteOptions` to connect to the v4 API.

## Requests

Removed the dependency on RequestKit. Once we consolidate the three Mapbox Swift libraries, it’ll be straightforward for each of the API handlers to share the simple URL request and response code. This library chooses reasonable default networking behavior; if you need anything more sophisticated, you can access the URL request information to use with a custom NSURLSession. `Directions` no longer needs to be strongly held in order for the request to finish. Instead, the request is made against the shared URL session; to use a custom URL session, make the request yourself using the URL returned by the `URLForCalculatingDirections(options:)` method. Removed the `cancel()` method; instead, directly cancel the NSURLSessionDataTask returned by `calculateDirections(options:completionHandler:)`.

Added a shared (singleton) Directions object. Use the shared object if you’ve set your Mapbox access token in the `MGLMapboxAccessToken` key of your application’s Info.plist file. Otherwise, create a `Directions` object with the access token explicitly. A single directions object can handle multiple requests concurrently (since it just passes the requests off to the shared URL session).

Eliminated the separate APIs for calculating ETAs. Instead, all the options supported by the Directions API, including flags for including steps and the overview geometry, are now exposed through a new `RouteOptions` class. Effectively, this change adds support for the `geometries`, `overview`, `radiuses`, `steps`, and `continue_straight` query parameters and allows the heading accuracy of any waypoint to be customized. However, note that steps are no longer returned by default, and the overview geometry is simplified by default. If this turns out to be a major inconvenience or surprise, we can add specialized options classes for either use case. The `access_token` parameter is now the last URL parameter instead of the first. Broke out `MBDirectionsRequest.TransportType` into three profile identifier constants. You always specify a profile identifier, not a transport type. That makes the desired mode of transportation more open-ended.

## Responses

Removed the `MBDirectionsResponse` class in favor of passing the waypoints and routes from the response directly into the completion handler. Renamed `Route.geometry` to `Route.coordinates`. For Objective-C compatibility, there are additional methods that work with C arrays of coordinates. Also, all enums are backed by integer types instead of String, but they’re `CustomStringConvertible` to allow for the same expressiveness we’re used to in Swift.

Error handling is much more robust now. Various error conditions returned by the API cause the localized failure reason and recovery suggestion to be set in the NSError object that is passed into the completion handler.

## Documentation and other changes

Documented every public symbol. Some of the more finicky options get lots more documentation, discussing why you would use the option, when you avoid using the option, etc. There’s little chance this voluminous documentation will head off developer error, but it along with the improved error messages will help developers find their way when things do go wrong.

Updated test fixtures and expectations for the v5 server-side API and the revamped client-side API, and updated examples.

Fixes #41, fixes #43, fixes #44.

/cc @mapbox/directions @tmcw @friedbunny


MapboxDirections/MBRouteOptions.swift
```diff
@@ -0,0 +1,276 @@
...
+     */
+    public var waypoints: [Waypoint]
+    
+    /**
+     A Boolean value that indicates whether a returned route may require a point U-turn at an intermediate waypoint.
+     
+     If the value of this property is `true`, a returned route may require an immediate U-turn at an intermediate waypoint. At an intermediate waypoint, if the value of this property is `false`, each returned route may continue straight ahead or turn to either side but may not U-turn. This property has no effect if only two waypoints are specified.
+     
+     Set this property to `true` if you expect the user to traverse each leg of the trip separately. For example, it would be quite easy for the user to effectively “U-turn” at a waypoint if the user first parks the car and patronizes a restaurant there before embarking on the next leg of the trip. Set this property to `false` if you expect the user to proceed to the next waypoint immediately upon arrival. For example, if the user only needs to drop off a passenger or package at the waypoint before continuing, it would be inconvenient to perform a U-turn at that location.
```
1ec5: Here’s the documentation corresponding to the `continue_straight` query parameter.

MapboxDirections/MBWaypoint.swift
```diff
@@ -0,0 +1,127 @@
...
+     
+     For driving directions, this property can be useful for avoiding a route that begins by going in the direction opposite the current direction of travel. For example, if you know the user is moving eastwardly and the first waypoint is the user’s current location, specifying a heading of 90 degrees and a heading accuracy of 90 degrees for the first waypoint avoids a route that begins with a “head west” instruction.
+     
+     You should be certain that the user is in motion before specifying a heading and heading accuracy; otherwise, you may be unnecessarily filter out the best route. For example, suppose the user is sitting in a car parked in a driveway, facing due north, with the garage in front and the street to the rear. In that case, specifying a heading of 0 degrees and a heading accuracy of 90 degrees may result in a route that begins on the back alley or, worse, no route at all. For this reason, it is recommended that you only specify a heading and heading accuracy when automatically recalculating directions due to the user deviating from the route.
```
1ec5: Here’s the documentation corresponding to the `bearings` parameter.

1ec5:
This PR goes in a very different direction than #30 does. After seeing how MapboxStatic.swift gets away without much in the way of networking features, I think this library can get away with something similar. One big benefit with this approach is that there’s no object to keep a strong reference to. On the other hand, #20 becomes the developer’s responsibility.

Issue #62, opened 2016-08-01
Serialize destinations field


1ec5:
We should add a new destinations property to RouteStep using the destinations field in the JSON.

Issue #31, opened 2016-04-07
Add test cases for failures


tmcw:
We should have test cases covering
- HTTP failure, like a 500 with gibberish content
- Request failure, like a 403 with a JSON message property

1ec5:
This largely happened in #103.

PR #63, opened 2016-08-01
Parse destinations

bsudekum:
Closes: https://github.com/mapbox/MapboxDirections.swift/issues/62

@1ec5 Would it be worth changing the test case that uses a destination? Am I missing anything else?


MapboxDirections/MBRouteStep.swift
```diff
@@ -506,10 +506,17 @@ public class RouteStep: NSObject {
      */
     public let transportType: TransportType?
     
+    
+    /**
+     Destinations signs for the current way.
```
1ec5: The documentation is written for an audience that may not be familiar with routers. How about something like this:

> Destinations, such as [control cities](https://en.wikipedia.org/wiki/Control_city), that appear on guide signage for the road identified in the `name` property.
> 
> This property is typically available in steps leading to or from a freeway or expressway.

1ec5:
> Would it be worth changing the test case that uses a destination? Am I missing anything else?

Please add additional test cases. We’ve been wanting to do that for a while (e.g., #31).

bsudekum:
Updated description and added `destinations` test.

MapboxDirections/MBRouteStep.swift
```diff
@@ -506,10 +506,18 @@ public class RouteStep: NSObject {
      */
     public let transportType: TransportType?
     
+    /**
+     Destinations, such as [control cities](https://en.wikipedia.org/wiki/Control_city), that appear on guide signage for the road identified in the `name` property.
+     
+     This property is typically available in steps leading to or from a freeway or expressway.
+     */
+    public let destinations: String?
```
1ec5: Developers will expect this property to be an array if we name it `destinations`, plural. I think it could be called `destination`, by analogy with `name`, which is singular despite the fact that it can sometimes contain multiple names in one string.

One thing I'm unsure about is whether it's a problem that "destination" already conceptually refers to the destination waypoint. But let's go with `destination` and revisit if there's any confusion. The documentation is pretty clear in any case.

1ec5: Heh, then again, we have `instructions`, plural. Alright, your call. 😅

bsudekum: Always a fan of parity, I'd like to leave as is.

PR #65, opened 2016-08-12
Migrate to Swift 2.3

1ec5:
Also added a dummy watchOS target that embeds the watchOS framework to work around CocoaPods/CocoaPods#5687. We can remove the watchOS application and extension once CocoaPods/CocoaPods#5733 is released.

/cc @superpeteblaze @boundsj


superpeteblaze:
Awesome, 👍 from me once Xcode 8 is GA. (I'm guessing once Xcode 8 is GA, this will be merged and version tagged. And them we can do the same with the Swift 3 branch.)

1ec5:
This branch maintains compatibility with Xcode 7.3 and Swift 2.2. In fact, it passed on Bitrise using the Xcode 7.3 stack. Any reason to wait until Xcode 8 GMs?

superpeteblaze:
Very good question...I was basing the strategy on [this](http://www.jessesquires.com/migrating-to-swift-3/). Thoughts?

1ec5:
I definitely see the value in waiting until the GM to merge the swift3 branch. 😄 For this particular library, no source changes were needed between Swift 2.2 and Swift 2.3, other than the CocoaPods workaround I mentioned. So it’s just a matter of adding a build setting that Xcode 7.3 ignores anyways.

I suspect the strategy in that blog post was meant for situations where source changes are required. (Although you could work around them with `#if swift(>=2.3)` constructs, assuming no one is using Xcode 7.2.)

superpeteblaze:
Cool, good insight and I agree. 👍

PR #64, opened 2016-08-03
Migrate to Swift 3

superpeteblaze:
## What changed
- Update directions target SWIFT_VERSION and test targets 
- Run Swift 3 migration on directions framework
- Convert example to swift 3
## Todos
- [x] Integrate with a Polyline swift3 branch
- [x] Swiftify the API
- [x] Document
## Notify

@1ec5 
## Note
- This of course won't build without a swift3 version of Polyline

PR #71, opened 2016-09-15
Upgrade project to Xcode 8.1

1ec5:
This change upgrades project settings to Xcode 8.1, upgrades CocoaPods to 1.2.0.beta.1, and upgrades dependencies. It also fixes some build warnings.

Along the lines of mapbox/MapboxGeocoder.swift#66, this change sets the `BITCODE_GENERATION_MODE` build setting so that applications that link against this library can archive.

Depends on raphaelmor/Polyline#34.

/cc @boundsj


1ec5:
Archiving the library still fails with:

```
ld: bitcode bundle could not be generated because '/Users/mxn/Library/Developer/Xcode/DerivedData/MapboxDirections-edgcxadoaxriyleblulvzecsblrz/Build/Products/Debug-watchos/Polyline-watchOS/Polyline.framework/Polyline' was built without full bitcode. All frameworks and dylibs for bitcode must be generated from Xcode Archive or Install build for architecture armv7k
```

1ec5:
@boundsj, any idea why setting `BITCODE_GENERATION_MODE = bitcode` in both MapboxDirections and Polyline causes both MapboxDirections and any downstream application to fail to link with the error in https://github.com/mapbox/MapboxDirections.swift/pull/71#issuecomment-257108388?

1ec5:
As usual, I forgot that CocoaPods puts together a different project than the one that’s checked into the Polyline repository. So raphaelmor/Polyline#34 had no effect. We probably have to introduce `BITCODE_GENERATION_MODE = bitcode` in the podspec.

1ec5:
I see no instances of `s.pod_target_xcconfig = { 'BITCODE_GENERATION_MODE' => 'bitcode' }` in podspecs on GitHub, so that can’t be the right answer either.

1ec5:
Now the watchOS archive step is failing on Bitrise with:

```
No mobileprovision_path found
```

Archiving locally works fine.

1ec5:
Bitrise suggests flipping `SKIP_INSTALL`, but I think `YES` is the appropriate value for that setting, since this is a framework rather than an application. I reverted #85 and modified the watchOS scheme to analyze and archive WatchExample instead of a bare watchOS framework.

1ec5:
To get the watchOS tests passing again, I'll probably replace the existing analyze step with a Bash script step that runs `xcodebuild build`, since Bitrise doesn't provide a prepackaged step to do that.

Eldorado234:
I guess this PR hast to be updated to reflect the changes made to the project setup in #102 (now merged into #64)

1ec5:
#102 is only on the Swift 3 branch. I was holding out hope that we could publish a final Swift 2.3 release before cutting over to Swift 3, but that’s going to be less important if this drags on past the Xcode 8.2 release (the last release that supports Swift 2.3).

1ec5:
Superseded by #108.

PR #72, opened 2016-09-16
Avoid deserializing non-JSON response data

1ec5:
Fixed an assertion failure that could occur (instead of an error being passed to the completion handler) if the response data was not the expected JSON format. This can happen, for instance, on a captive network.

Normally, one might also consider checking the response’s status code at this point. However, we actually do need to parse the JSON response for certain error codes like 404. So that error handling happens in the subsequent call to `Directions.descriptiveError(_:response:underlyingError:)`.

/cc @bsudekum @willwhite


superpeteblaze:
👍

PR #73, opened 2016-09-27
Bump polyline

frederoni:


frederoni:
The examples depends on
```
pod 'Polyline', :git => 'https://github.com/superpeteblaze/Polyline', :branch => 'swift3'
```
Should we bump that to 4.0 as well?

superpeteblaze:
@frederoni I don't think it is needed. The examples can prob switch to the original Polyline reference

Issue #79, opened 2016-10-17
Use-lane, rotary maneuver types


1ec5:
ManeuverType should have `useLane` and `traverseRotary` values that correspond to the Directions API’s `use lane` and `rotary` types, respectively. We should also update the documentation for `RouteStep.exitIndex` to indicate that it applies to rotaries as well as roundabouts.

/cc @bsudekum @mayagao

1ec5:
Fixed in #93.

PR #80, opened 2016-10-17
Intersections and lanes

bsudekum:
This starts work for adding a lanes class to each step. Currently, this adds an `Intersections` class where the lanes class will live. 

todo:
- [x] document
- [x] add lanes class

/cc @1ec5


bsudekum:
@1ec5 Added lanes, want to take a look?

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+
+public class Intersection: NSObject, NSSecureCoding {
```
1ec5: Use `@objc` to pretend the MB class prefix for Objective-C usage.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,56 @@
...
+    public var inIndex: Int?
```
1ec5: This is an awkward place for a preposition. How about `entranceIndex` and `exitIndex`?

Also, does `[Bool].Index` work as the type for this property?

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,56 @@
...
+    public var outIndex: Int?
+    public var entry: [Bool]
```
1ec5: Instead of an array of Booleans, this property should be an IndexSet called `outlets` that contains the indices of the roads through which the user can exit the intersection.

MapboxDirections/MBIntersection.swift
1ec5: Actually, to avoid confusion with the existing `exitIndex` property on RouteStep, how about `approachIndex` and `outletIndex`, or `ingressIndex` and `egressIndex`?

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
+
+public enum LaneIndicationType: Int, CustomStringConvertible {
+    case Left
```
1ec5: In Swift 3, enum values start with lowercase letters.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,83 @@
...
+    
+    case SharpLeft
+    
+    case SharpRight
+    
+    case SlightLeft
+    
+    case SlightRight
+    
+    case Straight
```
1ec5: straightAhead for consistency with ManeuverDirection.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,83 @@
...
+            return "slight right"
+        case .Straight:
+            return "straight"
+        case .Uturn:
+            return "uturn"
+        }
+    }
+}
+
+public class Lane: NSObject {
```
1ec5: `@objc(MBLane)`

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,83 @@
...
+    public var validTurn: Bool
```
1ec5: Whether a lane can be used for a particular maneuver is not an intrinsic property of a lane. Remove this property and add to Intersection a `usableLanes` property of type IndexSet or `Set<Lane>`, your pick.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
+
+public enum LaneIndicationType: Int, CustomStringConvertible {
```
1ec5: Just LaneIndication is fine. But remember to use `@objc` here too.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,83 @@
...
+            return "straight"
+        case .Uturn:
+            return "uturn"
+        }
+    }
+}
+
+public class Lane: NSObject {
+    public var validTurn: Bool
+    public var indications: [LaneIndicationType]
```
1ec5: An array of enums won't bridge to Objective-C. We'll need a helper property that converts it to an array of NSValues containing these enums.

MapboxDirections/MBIntersection.swift
1ec5: Does this property need to be mutable? If not, use `let`.

MapboxDirections/MBRouteStep.swift
```diff
@@ -513,6 +513,11 @@ public class RouteStep: NSObject, NSSecureCoding {
      */
     public let destinations: String?
     
+    /**
+     Array of interersections
```
1ec5: Typo. Also, explain that these are all the intersections along the step, starting with the one at the maneuver and ending with the one just before the next step's maneuver.

1ec5: And mention the relationship between the size of this array and the next step's `exitIndex` property.

MapboxDirections/MBLane.swift
1ec5: I forgot this is Swift 2.3. Just keep in mind that we'll need to lowercase the first letters when porting to Swift 3.

MapboxDirections/MBLane.swift
bsudekum: I'm unclear why this property would live on intersection. It is a specific value for a specific lane in the lanes array.

1ec5: I know that’s how it’s exposed in the API. But the API doesn’t use classes at all, so that’s comparing apples and oranges.

When you create a class, you make it possible for the developer to have an instance of that class all by itself. If the developer has a Lane object but not the Intersection or RouteStep object that contains it, `validTurn` is meaningless or misleading. It’s saying that a physical lane on a road can always or never be used for all routes that may contain the lane, which is unrealistic.

Semantically, the information you’re trying to store must live in a parallel data structure. Specifically, Intersection should have a `usableLanes` property that _contains_ lanes or lane indices, and this `validTurn` property should be removed.

MapboxDirections/MBIntersection.swift
```diff
@@ -1,38 +1,63 @@
...
 
+@objc(MBIntersection)
 public class Intersection: NSObject, NSSecureCoding {
-    public var inIndex: Int?
-    public var outIndex: Int?
+    
+    /**
+     Index into bearings/entry array.
+     
+     Used to calculate the bearing just before the turn. Namely, the clockwise angle from true north to the direction of travel immediately before the maneuver/passing the intersection. Bearings are given relative to the intersection. To get the bearing in the direction of driving, the bearing has to be rotated by a value of 180. The value is not supplied for depart maneuvers.
```
1ec5: Is it always 180°?

MapboxDirections/MBLane.swift
```diff
@@ -56,27 +89,48 @@ public enum LaneIndicationType: Int, CustomStringConvertible {
...
+    /**
+     An indication (e.g. marking on the road) specifying the turn lane.
+     
+     A road can have multiple indications (e.g. an arrow pointing straight and left). The indications are given in an array, each containing one of the following types. Further indications might be added on without an API version change.
+    */
+    public let indications: [LaneIndication]
+    
+    
+    /**
+     Objectice-c helper function for indication
```
1ec5: `indications` won’t even appear to an Objective-C developer. You’ll have to document this property as if `indications` doesn’t even exist.

MapboxDirections/MBRouteStep.swift
```diff
@@ -514,7 +514,9 @@ public class RouteStep: NSObject, NSSecureCoding {
     public let destinations: String?
     
     /**
-     Array of interersections
+     An intersection gives a full representation of any cross-way the path passes bay.
```
1ec5: You mean “of any road crossing the route”?

bsudekum:
@1ec5 updated with `usableLanes`

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+
+@objc(MBIntersection)
+public class Intersection: NSObject, NSSecureCoding {
+    
+    /**
+     Index into bearings/entry array.
+     
+     Used to calculate the bearing just before the turn. Namely, the clockwise angle from true north to the direction of travel immediately before the maneuver/passing the intersection. Bearings are given relative to the intersection. To get the bearing in the direction of driving, the bearing has to be rotated by a value of 180. The value is not supplied for depart maneuvers.
+    */
+    public let approachIndex: Int?
```
1ec5: Unfortunately, `Int?` won’t bridge to Objective-C, so this will have to be `Int` instead. Set it to `-1` by default (for the v4 case). Unfortunate, but this is one of the tradeoffs of supporting Objective-C.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+     Used to extract the bearing just after the turn. Namely, The clockwise angle from true north to the direction of travel immediately after the maneuver/passing the intersection. The value is not supplied for arrive maneuvers.
+    */
+    public let outletIndex: Int?
+    
+    /**
+     An array of booleans, corresponding in a 1:1 relationship to the bearings.
+     
+     A value of true indicates that the respective road could be entered on a valid route. false indicates that the turn onto the respective road would violate a restriction.
+     */
+    public var entry: [Bool]
```
1ec5: Does this property need to be mutable?

1ec5: https://github.com/mapbox/MapboxDirections.swift/pull/80#discussion_r83768884

Instead of an array of Booleans, this property should be an IndexSet called `outlets` that contains the indices of the roads through which the user can exit the intersection. These would be indices into the `headings` array. The index set would be equal in size to or smaller than the `headings` array.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+    
+    /**
+     Array of Lane objects that denote the available turn lanes at the intersection.
+     
+     If no lane information is available for an intersection, the lanes property will not be present.
+    */
+    public var lanes: [Lane]?
+    
+    /**
+     Set of Lane objects that have a valid turn.
```
1ec5: Also mention that this is a subset of the Lane objects in the `lanes` property.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+    
+    /**
+     An array of CLLocationDirection values that are available at the intersection.
+     
+     The bearings describe all available roads at the intersection.
+    */
+    public var headings: [CLLocationDirection]
+    
+    /**
+     Array of Lane objects that denote the available turn lanes at the intersection.
```
1ec5: This isn’t just for turn lanes.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+    
+    /**
+     CLLocationCoordinate2D representing the location of the intersection
+    */
+    public var location: CLLocationCoordinate2D
+    
+    /**
+     An array of CLLocationDirection values that are available at the intersection.
+     
+     The bearings describe all available roads at the intersection.
```
1ec5: s/bearings/headings/

> An array of `CLLocationDirection` values corresponding to the headings of the roads that lead in or out of the intersection. Each heading is measured in degrees clockwise from true north.

Actually, from the discussion above, it sounds like these headings aren’t measured from true north, in which case we should either transform them for the developer or stop using `CLLocationDirection`, which is documented to have true north as its reference point.

1ec5: Can we say anything about how these values are arranged? Are they sorted clockwise? Do they start on the driver’s left-hand side, or at 12:00, or at true north?

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+    
+    /**
+     An array of booleans, corresponding in a 1:1 relationship to the bearings.
+     
+     A value of true indicates that the respective road could be entered on a valid route. false indicates that the turn onto the respective road would violate a restriction.
+     */
+    public var entry: [Bool]
+    
+    /**
+     CLLocationCoordinate2D representing the location of the intersection
```
1ec5: Missing a period. Also no need to say the type, since the type is already declared:

> The geographic coordinates at the center of the intersection.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
+import Foundation
+
+@objc(MBIntersection)
+public class Intersection: NSObject, NSSecureCoding {
+    
+    /**
+     Index into bearings/entry array.
```
1ec5: > The index of the item in the `headings` array that corresponds to the road that the containing route step uses to approach the intersection.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+     
+     Used to calculate the bearing just before the turn. Namely, the clockwise angle from true north to the direction of travel immediately before the maneuver/passing the intersection. Bearings are given relative to the intersection. To get the bearing in the direction of driving, the bearing has to be rotated by a value of 180. The value is not supplied for depart maneuvers.
+    */
+    public let approachIndex: Int?
+    
+    /**
+     Index into the bearings/entry array.
```
1ec5: > The index of the item in the `headings` array that corresponds to the road that the containing route step uses to leave the intersection.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+        let outletIndex = json["out"] as? Int
+        let entry = json["entry"] as! [Bool]
+        let locationArray = json["location"] as! [Double]
+        let location = CLLocationCoordinate2D(latitude: locationArray[0], longitude: locationArray[1])
+        let headings = json["bearings"] as! [CLLocationDirection]
+        let lanesJSON = json["lanes"] as? [JSONDictionary]
+        var lanes = [Lane]()
+        var usableLanes = Set<Lane>()
+        
+        lanesJSON?.forEach({ (laneJSON) in
```
1ec5: In Swift, only use `forEach` if you intend to pass in a closure or function declared somewhere else. Since here you’re using a closure inline, use a for-in loop instead. For loops have exactly the same performance characteristics as `forEach`, but with less punctuation.

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
...
+        self.headings = headings
+        self.lanes = lanes
+        self.usableLanes = usableLanes
+    }
+    
+    internal convenience init(json: JSONDictionary) {
+        let approachIndex = json["in"] as? Int
+        let outletIndex = json["out"] as? Int
+        let entry = json["entry"] as! [Bool]
+        let locationArray = json["location"] as! [Double]
```
1ec5: Create a `CLLocationCoordinate2D` directly from this array using [this initializer](https://github.com/mapbox/MapboxDirections.swift/blob/eb2be1fe05361f8e586a9d523114a066cb60b07d/MapboxDirections/MBDirections.swift#L58).

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,108 @@
+import Foundation
+
+@objc(MBIntersection)
```
1ec5: This class needs a documentation comment.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,121 @@
+import Foundation
+
+@objc(MBLaneIndication)
```
1ec5: This type needs a documentation comment.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,121 @@
...
+public enum LaneIndication: Int, CustomStringConvertible {
+    
+    /**
+     An indication indicating a turn to the left.
```
1ec5: > Indicates a turn to the left.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,121 @@
...
+        case .None:
+            return "None"
+        }
+    }
+}
+
+@objc(MBLane)
+public class Lane: NSObject {
+    
+    public let indications: [LaneIndication]
```
1ec5: It occurs to me that we need to treat this as an option set rather than an array of enum values. After all, a lane can’t be `[.Uturn, .Uturn]`. Option sets are really nice in Objective-C and really nice in Swift, but unfortunately Swift option sets don’t bridge to Objective-C. So you’ll have to implement LaneIndication in Objective-C. Use MapboxGeocoder.swift’s [`MBPlacemarkScope`](https://github.com/mapbox/MapboxGeocoder.swift/blob/master/MapboxGeocoder/MBPlacemarkScope.h) option set and [`PlacemarkScope`](https://github.com/mapbox/MapboxGeocoder.swift/blob/master/MapboxGeocoder/MBPlacemarkScope.swift) extension as a model.

MapboxDirections/MBRouteStep.swift
```diff
@@ -513,6 +513,13 @@ public class RouteStep: NSObject, NSSecureCoding {
      */
     public let destinations: String?
     
+    /**
+     An intersection gives a full representation of any cross-way the path passes bay.
```
1ec5: > An array of intersections along the step.

> Each item in the array corresponds to a cross street, starting with the intersection at the maneuver location indicated by the `coordinates` property and continuing with each cross street along the step.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -86,6 +86,18 @@ class V5Tests: XCTestCase {
         XCTAssertEqual(round(coordinate.longitude), -122)
         
         XCTAssertEqual(leg.steps[18].name, "Sycamore Valley Road West")
+        
+        let intersection = step.intersections![0]
```
1ec5: You can also say `step.intersections.first!`. I find that easier to type because of code completion.

MapboxDirections/MBLane.swift
bsudekum: @1ec5 are you saying use a set because the type cannot be repeated?

MapboxDirections/MBIntersection.swift
bsudekum: I believe so

MapboxDirections/MBLane.swift
1ec5: Not a set, but an option set ([`NS_OPTIONS`](http://nshipster.com/ns_enum-ns_options/) in Objective-C). Not only because a type can’t be repeated, but also because using an `indicationValues` property (of type `[NSValue]`) in Objective-C would look like this:

```objc
BOOL isStraightAhead;
for (NSValue *indicationValue in lane.indicationValues) {
    MBLaneIndication laneIndication;
    [indicationValues.firstObject getValue:&laneIndication objCType:@encode(MBLaneIndication)];
    if (laneIndication == MBLaneIndicationStraightAhead) {
        isStraightAhead = YES;
        break;
    }
}
if (isStraightAhead) …

```

and there’s no clean way to get _all_ the indications.

If we implement `MBLaneIndication` using `NS_OPTIONS` in Objective-C, then the Objective-C code becomes a lot simpler:

```objc
MBLaneIndication indications = lane.indications;
if (indication & MBLaneIndicationStraightAhead) …
```

and the Swift code stays the same:

```swift
if lane.indications.contains(.straightAhead) …
```

MapboxDirections/MBIntersection.swift
```diff
@@ -1,54 +1,55 @@
...
 public class Intersection: NSObject, NSSecureCoding {
     
     /**
      Index into bearings/entry array.
      
-     Used to calculate the bearing just before the turn. Namely, the clockwise angle from true north to the direction of travel immediately before the maneuver/passing the intersection. Bearings are given relative to the intersection. To get the bearing in the direction of driving, the bearing has to be rotated by a value of 180. The value is not supplied for depart maneuvers.
+     The index of the item in the headings array that corresponds to the road that the containing route step uses to approach the intersection.
     */
-    public let approachIndex: Int?
+    public var approachIndex: Int = -1
```
1ec5: We probably don’t need to allow clients to modify these values. So you can either change this back to `public let` and set it to `-1` in the initializer (as you seem to do anyways), or you can say `public private(set) var`.

MapboxDirections/MBIntersection.swift
```diff
@@ -1,54 +1,55 @@
...
+     The geographic coordinates at the center of the intersection.
     */
     public var headings: [CLLocationDirection]
     
     /**
-     Array of Lane objects that denote the available turn lanes at the intersection.
+     Array of Lane objects.
      
-     If no lane information is available for an intersection, the lanes property will not be present.
+     If no lane information is available for an intersection, the lanes property will not be present. The first lane represents the left most lane, while the last represents the right most lane.
```
1ec5: 👍

Is this ordering the same for countries that drive on the left?

MapboxDirections/MBIntersection.swift
```diff
@@ -59,30 +60,30 @@ public class Intersection: NSObject, NSSecureCoding {
...
     internal convenience init(json: JSONDictionary) {
-        let approachIndex = json["in"] as? Int
-        let outletIndex = json["out"] as? Int
+        let approachIndex = json["in"] as? Int ?? -1
+        let outletIndex = json["out"] as? Int ?? -1
         let entry = json["entry"] as! [Bool]
-        let locationArray = json["location"] as! [Double]
-        let location = CLLocationCoordinate2D(latitude: locationArray[0], longitude: locationArray[1])
+        let coords = json["location"] as! [Double]
+        let location = CLLocationCoordinate2D.init(geoJSON: coords)
```
1ec5: Nit: drop `.init`.

MapboxDirections/MBIntersection.swift
```diff
@@ -59,30 +60,30 @@ public class Intersection: NSObject, NSSecureCoding {
...
+        }
         
         self.init(approachIndex: approachIndex, outletIndex: outletIndex, entry: entry, location: location, headings: headings, lanes: lanes, usableLanes: usableLanes)
     }
     
     public required init?(coder decoder: NSCoder) {
-        approachIndex = decoder.decodeObjectForKey("approachIndex") as? Int
-        outletIndex = decoder.decodeObjectForKey("outletIndex") as? Int
+        approachIndex = decoder.decodeObjectForKey("approachIndex") as! Int
+        outletIndex = decoder.decodeObjectForKey("outletIndex") as! Int
```
1ec5: #81 correctly points out that we need to use `decodeObjectOfClass(_:forKey:)` to be secure. If #81 lands first, we should update the added classes in this PR to use `decodeObjectOfClass(_:forKey:)` too.

MapboxDirections/MBIntersection.swift
bsudekum: They should line up with what's printed on the road if you were standing on it looking at the arrows

MapboxDirections/MBIntersection.swift
```diff
@@ -0,0 +1,109 @@
...
+     The index of the item in the headings array that corresponds to the road that the containing route step uses to leave the intersection.
+    */
+    public var outletIndex: Int
+    
+    /**
+     An array of booleans, corresponding in a 1:1 relationship to the headings.
+     
+     A value of true indicates that the respective road could be entered on a valid route. false indicates that the turn onto the respective road would violate a restriction.
+     */
+    public let entry: [Bool]
```
1ec5: Still waiting for this method to be renamed to `outlets` and changed to be an IndexSet.

MapboxDirections/MBLaneIndication.h
```diff
@@ -0,0 +1,29 @@
+
+typedef NS_OPTIONS(NSUInteger, MBLaneIndication) {
+    // Indicates a sharp turn to the right.
+    SharpRight = (1 << 4),
```
1ec5: Prefix each of these values with `MBLaneIndication`. In Swift, `MBLaneIndication` will be removed automatically.

MapboxDirections/MBLane.swift
```diff
@@ -1,125 +1,84 @@
...
+        if contains(LaneIndication.SlightRight) {
+            descriptions.append("slight right")
+        }
+        if contains(LaneIndication.StraightAhead) {
+            descriptions.append("straight")
+        }
+        if contains(LaneIndication.Uturn) {
+            descriptions.append("uturn")
+        }
+        if contains(LaneIndication.None) {
```
daniel-j-h: These are the only values we return:

https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#properties-5

`||right` and `none|none|right` are the same for OSRM and will be converted to the fixed set of return types in the spec above.

MapboxDirections/MBLane.swift
```diff
@@ -0,0 +1,86 @@
...
+            descriptions.append("slight right")
+        }
+        if contains(LaneIndication.StraightAhead) {
+            descriptions.append("straight")
+        }
+        if contains(LaneIndication.SlightLeft) {
+            descriptions.append("slight left")
+        }
+        if contains(LaneIndication.Left) {
+            descriptions.append("Left")
```
1ec5: s/Left/left/

MapboxDirections/MBLane.swift
1ec5: The other question was whether `none;right` for a particular lane would turn into `["none", "right"]` or `["straight", "right"]` or just `["right"]`. I’m inclined to remove the `None` value and use `[]` to represent `["none"]`. If `["none", right"]` ever comes out of the API, we can ticket out turning that into `[.StraightAhead, .Right]`.

1ec5:
I’ve addressed all the feedback above and will merge this PR once most of the builds pass. Then I’ll update the swift3 branch. I’m going to let the watchOS build go red on master while we await #71.

1ec5:
Note that this PR does not add support for “use lane” instructions. That’s tracked in #79.

bsudekum:
🎉

1ec5:
Merged to the swift3 branch in 67b8a95a1bd8f306db45a37f69fbe7117de4be9b.

PR #81, opened 2016-10-18
Use decodeObjectOfClass(es) methods to conform to NSSecureCoding

zorgiepoo:
Hi all,

I was looking over the commit history and noticed that the recent addition of conforming to NSSecureCoding was not decoding the objects securely, so consider this a friendly contribution :).

--
Checking the data type after the data has been unarchived is not sufficient because arbitrary code could have already been executed by that point. For decoding collections, a whitelist of classes that are okay to decode must be specified.

The RouteStepTests have changed to require secure coding when archiving/unarchiving the data. Tests for other coding classes don't yet exist, but could be considered as a future todo.


zorgiepoo:
Updated to remove force unwrapping (!) when decoding.

PR #86, opened 2016-10-22
Add driving-traffic profile.

willwhite:
Adds a constant for the `driving-traffic` profile.

### DIFF
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -10,7 +10,7 @@ public let MBDirectionsProfileIdentifierAutomobile = "mapbox/driving"
 /**
  The returned directions are appropriate for driving and incorporates current traffic conditions when determining the fastest route.
  */
-public let MBDirectionsProfileIdentifierAutomobileTraffic = "mapbox/driving-traffic"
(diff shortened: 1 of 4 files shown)


MapboxDirections/MBRouteOptions.swift
```diff
@@ -8,6 +8,11 @@
 public let MBDirectionsProfileIdentifierAutomobile = "mapbox/driving"
 
 /**
+ The returned directions are appropriate for driving and incorporates current traffic conditions when determining the fastest route.
```
1ec5: Point out that this profile is experimental, like we do in MapboxGeocoder.swift for the [`locale`](https://github.com/mapbox/MapboxGeocoder.swift/blob/042bf7dcad23b84c9309f4b034b8ce5c905e59f9/MapboxGeocoder/MBGeocodeOptions.swift#L58) option.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -8,6 +8,11 @@
...
+ */
+public let MBDirectionsProfileIdentifierAutomobileTraffic = "mapbox/driving-traffic"
```
1ec5: How about `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`?

Issue #67, opened 2016-09-07
Serialize ref field


1ec5:
Once Project-OSRM/osrm-backend#2857 makes it into production and the Directions API exposes a separate ref field for each step, we'll need to add a route reference property to each RouteStep, similar to the one for destinations. Assuming that v4 doesn't add this separate field, we'll fall back to a regular expression that parses the ref out of the way name.

Better yet, we could implement a Road struct that bundles a road's name with its route designation. Eventually that struct could grow to include other information about the road, such as its classification.

/cc @bsudekum

superpeteblaze:
Road struct that bundles the name & designation gets my vote 😄

1ec5:
The struct will need to be implemented in Objective-C, so it'll bridge to Swift. If we implement it in Swift, it won't bridge to Objective-C.

1ec5:
Here are two options for the implementation:

* Add a `roadDesignation` property to `RouteStep`.
* Replace the `name` property with a `road` property of type `Road`. `Road` has two fields, `name` and `designation`. `Road` is either a Swift class with `String`-typed fields or a C struct with `char *`-typed fields, but not a Swift struct.

We should avoid the term “ref”, because that’s OpenStreetMap jargon (really just an OSM tag name). A better term is “[road designation](https://en.wikipedia.org/wiki/Road_designation_or_abbreviation)”.

Based on https://github.com/mapbox/MapboxDirections.swift/issues/77#issue-181192421, we’ll also need to trim the parenthesized designation from the `name` property in the response object, because apparently the Directions API is including that information redundantly in both `ref` and `name`.

bsudekum:
> Add a roadDesignation property to RouteStep.

I'm for this. This allows the developer to decide if and when to use a `ref` vs a `name`.

1ec5:
Both options let the developer decide if and when to use a `ref` versus a `name`. With the `Road` class, you’d say `road.ref` or `road.name`.

1ec5:
To avoid a proliferation of properties on RouteStep, we’re definitely going to need a `Road` class or struct. `Road` will be useful for #78 as well as the `rotary_name` and `rotary_pronunciation` fields, which we would simply serialize as a `rotary` property of type `Road`.

simplyira:
Hey! I came across this discussion, because "road designation" sounds like the exact property I need. :) Can't find documentation yet. Any way to access it? The obvious .road or .roadDesignation on step are not working. Thanks

[...]

PR #89, opened 2016-11-08
Serialize none indication as empty indication

1ec5:
When the `indication` field coming from the API is `none`, convert that to an empty `LaneIndication` value instead of `nil`.

/cc @bsudekum


1ec5:
I’ll update the tests and test fixtures as part of implementing #67.

PR #91, opened 2016-11-09
Structured road names, codes, destinations

### LINKED ISSUE #67: Serialize ref field

1ec5:
Changed RouteStep’s `name` and `destinations` properties into arrays, so that client code doesn’t have to do any splitting or parsing on its own. Removed redundant refs from the name property. Added a `codes` property that contains the route numbers. Though a Route struct type is used internally during parsing, all the new properties live directly on RouteStep for easier Objective-C bridging.

Also updated test fixtures and tests.

Fixes #67.

/cc @bsudekum @freenerd


MapboxDirections/MBRouteStep.swift
```diff
@@ -371,6 +371,60 @@ public enum ManeuverDirection: Int, CustomStringConvertible {
...
+            // OSRM v5.5 encodes the ref separately from the name but redundantly includes the ref in the name for backwards compatibility. Remove the ref from the name.
+            let parenthetical = "(\(ref))"
+            if names == ref {
+                self.names = nil
+            } else {
+                self.names = names.stringByReplacingOccurrencesOfString(parenthetical, withString: "").tagValuesSeparatedByString(";")
+            }
+            codes = ref.tagValuesSeparatedByString(";")
+        } else if let names = name, let codesRange = names.rangeOfString("\\(.+?\\)$", options: .RegularExpressionSearch, range: names.startIndex..<names.endIndex) {
+            // OSRM v5.0 encodes the ref inside a parenthetical. Remove the ref from the name.
```
freenerd: I think we are good dropping support for older OSRM versions

1ec5: Let’s tackle removing Directions API v4 support in a separate PR.

MapboxDirections/MBRouteStep.swift
```diff
@@ -371,6 +371,60 @@ public enum ManeuverDirection: Int, CustomStringConvertible {
...
+ */
+struct Road {
+    let names: [String]?
+    let codes: [String]?
+    let destinations: [String]?
+    let destinationCodes: [String]?
+    
+    init(name: String?, ref: String?, destination: String?) {
+        var codes: [String]?
+        if let names = name, let ref = ref ?? destination {
```
freenerd: `let ref = ref ?? destination`

I think we should not put destination into ref, since these are difference concepts. We also have extra destination handling below, so I think `?? destination` can be removed here.

freenerd: Note that `name` will never be `undefined/nil` from OSRM, but an empty string "". So effectively this `if` only tests for `ref` being present.

1ec5: `ref` is a misnomer here: what I really meant is “whatever is inside the parentheses”, which can be either a ref or a destination, due to munging that either the Directions API or OSRM performs (not sure which).

MapboxDirections/MBRouteStep.swift
```diff
@@ -382,7 +436,11 @@ public class RouteStep: NSObject, NSSecureCoding {
     
     internal init(finalHeading: CLLocationDirection?, maneuverType: ManeuverType?, maneuverDirection: ManeuverDirection?, maneuverLocation: CLLocationCoordinate2D, name: String?, coordinates: [CLLocationCoordinate2D]?, json: JSONDictionary) {
         transportType = TransportType(description: json["mode"] as! String)
-        destinations = json["destinations"] as? String
+        
+        let road = Road(name: name, ref: json["ref"] as? String, destination: json["destinations"] as? String)
+        names = road.names
+        codes = road.codes ?? road.destinationCodes
+        destinations = road.destinations
```
freenerd: I assume these are properties on `self.`? Either these should be set that way, or below assignments should not have `self.=` either (like `self.intersections =` -> `intersections =`)

MapboxDirections/MBRouteStep.swift
```diff
@@ -371,6 +371,60 @@ public enum ManeuverDirection: Int, CustomStringConvertible {
...
+ */
+struct Road {
+    let names: [String]?
+    let codes: [String]?
+    let destinations: [String]?
+    let destinationCodes: [String]?
+    
+    init(name: String?, ref: String?, destination: String?) {
+        var codes: [String]?
+        if let names = name, let gloss = ref ?? destination {
```
freenerd: I don't understand the semantic meaning of `gloss` and the dictionary did not help me understand. The case that in Mapbox Directions v5 `destination` is mixed into `name` can not exist. Likewise it makes no sense to me that `codes` would ever have the contents of `destination` in them.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -54,7 +54,7 @@ class V5Tests: XCTestCase {
         
         XCTAssertNotNil(route)
         XCTAssertNotNil(route!.coordinates)
-        XCTAssertEqual(route!.coordinates!.count, 842)
+        XCTAssertEqual(route!.coordinates!.count, 28_442)
```
freenerd: What's the semantics of `28_442`? Is this a range or math exponent syntax?

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -63,40 +63,47 @@ class V5Tests: XCTestCase {
...
+        XCTAssertEqual(intersection?.headings ?? [], [15, 90, 195, 270])
+        XCTAssertNotNil(intersection?.location.latitude)
+        XCTAssertNotNil(intersection?.location.longitude)
+        XCTAssertEqual(intersection?.usableApproachLanes ?? [], NSIndexSet(indexesInRange: NSRange(location: 1, length: 3)))
         
-        let lane = intersection.approachLanes?.first
+        let lane = intersection?.approachLanes?.first
         let indications = lane?.indications
         XCTAssertNotNil(indications)
         XCTAssertTrue(indications!.contains(.Left))
```
freenerd: it might make sense to add a test for `step.names` being populated.

MapboxDirections/MBRouteStep.swift
```diff
@@ -388,17 +388,17 @@ struct Road {
     
     init(name: String?, ref: String?, destination: String?) {
         var codes: [String]?
-        if let names = name, let ref = ref ?? destination {
-            // Mapbox Directions API v5 encodes the ref separately from the name but redundantly includes the ref in the name for backwards compatibility. Remove the ref from the name.
-            let parenthetical = "(\(ref))"
-            if names == ref {
+        if let names = name, let gloss = ref ?? destination {
+            // Mapbox Directions API v5 encodes the ref separately from the name but redundantly includes the ref or destination in the name for backwards compatibility. Remove the ref or destination from the name.
```
freenerd: To clarify: This is not true, the `direction` is never mixed into the `name`, but only the `ref`.

MapboxDirections/MBRouteStep.swift
```diff
@@ -382,7 +436,11 @@ public class RouteStep: NSObject, NSSecureCoding {
     
     internal init(finalHeading: CLLocationDirection?, maneuverType: ManeuverType?, maneuverDirection: ManeuverDirection?, maneuverLocation: CLLocationCoordinate2D, name: String?, coordinates: [CLLocationCoordinate2D]?, json: JSONDictionary) {
         transportType = TransportType(description: json["mode"] as! String)
-        destinations = json["destinations"] as? String
+        
+        let road = Road(name: name, ref: json["ref"] as? String, destination: json["destinations"] as? String)
+        names = road.names
+        codes = road.codes ?? road.destinationCodes
```
1ec5: The idea here was to make it easier for applications to display a route shield for an upcoming ramp based on `codes` without having to consult a separate property. We’ve since moved away from the notion that MapboxDirections.swift should perform any sort of munging, so I’ll remove this fallback.

I’ve gone back and forth quite a bit about whether destination codes should be conflated with `destination` or live in a separate property. In OSM, `destination:ref` usually follows a different format and encodes more information than `ref`. For example, if I-80 travels over a ramp to go from one motorway to another, the `highway=motorway_link` way is tagged `ref=I 80` but may also be tagged `destination:ref=I-80 West`. The format is loose enough that I don’t feel confident making `destinationCodes` look like `codes`. But I guess it can’t hurt much.

MapboxDirections/MBRouteStep.swift
1ec5: Turns out Directions API v5 _used_ to conflate `destination` with `ref` in the parenthetical, but it no longer does so, much to my relief. I’ll gladly remove the fallbacks.

MapboxDirections/MBRouteStep.swift
1ec5: I’ll remove `self.` from `self.intersections`, since there isn’t a conflict with a local variable named `intersections`.

MapboxDirections/MBRouteStep.swift
1ec5: Fixed and fixed.

MapboxDirections/MBRouteStep.swift
1ec5: Fixed. `destinationCodes` is a separate property now.

MapboxDirectionsTests/V5Tests.swift
1ec5: Added a test for this case.

PR #92, opened 2016-11-09
Make RouteStep json initialize public

freenerd:
This allows using RouteStep easier in tests

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -378,148 +378,6 @@ public enum ManeuverDirection: Int, CustomStringConvertible {
  */
 @objc(MBRouteStep)
 public class RouteStep: NSObject, NSSecureCoding {
-    // MARK: Getting the Step Geometry
-    
-    /**
-     An array of geographic coordinates defining the path of the route step from the location of the maneuver to the location of the next step’s maneuver.
-     
-     The value of this property may be `nil`, for example when the maneuver type is `Arrive`.
-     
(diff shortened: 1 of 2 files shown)


MapboxDirections/MBRouteStep.swift
```diff
@@ -547,7 +547,7 @@ public class RouteStep: NSObject, NSSecureCoding {
         self.coordinates = coordinates
     }
     
-    internal convenience init(json: JSONDictionary) {
+    public convenience init(json: [ String: Any ]) {
```
1ec5: Change `[String: Any]` back to `JSONDictionary`: in this library, `JSONDictionary` is a typedef for `[String: AnyObject]`. JSON objects are represented as `[String: AnyObject]` in Swift 2.3 but as `[String: Any]` in Swift 3.

This is why all the builds are failing.

1ec5: All public symbols should have documentation comments.

freenerd: `JSONDictionary` is a private typealias in MBDirections:

> MBRouteStep.swift:570:24: Initializer cannot be declared public because its parameter uses an internal type

Using `init(json: JSONDictionary` in external tests is prohibited by that. Not sure how to make `typealias` public (they should be by default?), but the easier solution was to change the function signature without the typealias

1ec5:
For consistency, the JSON initializers for Route and RouteLeg should be exposed publicly too.

freenerd:
@1ec5 `Route` already had a public json initializer. `RouteLeg` now has as well. Please feel free to adjust the inline comments.

gregoireLem:
Is it possible to make Route conform to Decodable ?

gregoireLem:
What's the best way to parse JSON object into Route object ?

1ec5:
> Is it possible to make Route conform to Decodable ?

#221

Issue #60, opened 2016-07-24
Persist original JSON response


1ec5:
An application may want to add its own request/response logging functionality. Currently, if it uses the built-in URL request functionality, it has to transform `Route` objects and so forth into a string-convertible format. We should add the original serialized JSON to the completion handler just in case the application needs it.

/cc @willwhite

1ec5:
We could consider storing the original JSON response in the RouteResponse class that was made public in #393 and would be documented as part of #391.

Storing the original JSON would incur some additional memory usage, but it would avoid the potentially expensive roundtripping from JSON to Swift model objects back to JSON that the navigation SDK has to perform for MapboxNavigationNative. That roundtripping has also proven fairly brittle, because MapboxNavigationNative is a lot stricter than this library has historically been about the JSON that it accepts.

1ec5:
> Currently, if it uses the built-in URL request functionality, it has to transform `Route` objects and so forth into a string-convertible format.

This issue was originally written against a version of the library that supported Objective-C and didn’t use Codable, so round-tripping was a laborious, error-prone affair. Migrating to Codable eliminated a large class of errors and made us feel much more confident about relying on round-tripping.

> Storing the original JSON would incur some additional memory usage, but it would avoid the potentially expensive roundtripping from JSON to Swift model objects back to JSON that the navigation SDK has to perform for MapboxNavigationNative.

JSON conversions no longer figure as large in performance profiles as they did when we relied on JSONSerialization directly, and the performance problems we initially experienced when investigating Codable have not turned out to be a major problem in practice.

> That roundtripping has also proven fairly brittle, because MapboxNavigationNative is a lot stricter than this library has historically been about the JSON that it accepts.

#637 proposes storing each JSON object’s unrecognized keys in a loosely typed property and round-tripping it when encoding back to JSON. That approach effectively solves this problem, because clients such as MapboxNavigationNative only need the JSON to be equivalent in value; any differences in whitespace formatting and key order shouldn’t matter.

PR #93, opened 2016-11-09
Add rotary and use lane types

freenerd:
These two types were not present yet.

@1ec5 What other changes are required to bring these types in?


1ec5:
This fixes #79.

1ec5:
We’ll need to add support for the `rotary_name` type as well. Let’s land #91 first; then we can either add a `junctionNames` property on RouteStep or add a heuristic for rotaries that puts `name` in the `destinations` property and `rotary_name` in the `names` property, for consistency with ramps.

freenerd:
> we can either add a junctionNames property on RouteStep 
> add a heuristic for rotaries that puts name in the destinations property and rotary_name in the names property, for consistency with ramps.

For integrating osrm-text-instructions it will be more useful to have direct access to the `rotary_name` property. I'd rather not have heuristics that mangle with properties, but instead retain the original response structure in the object and add helper functions that return heuristically-derived results

1ec5:
> For integrating osrm-text-instructions it will be more useful to have direct access to the rotary_name property.

Direct equivalence to JSON properties is a non-goal for this library, but #60 would keep the original serialized JSON around for anyone who finds it useful. The tradeoff in that case is higher memory usage.

Once #91 lands, the logic I’m proposing for RouteStep would look something like this:

```swift
if maneuverType == .rotary {
    destinations = road.names ?? road.destinations
    names = road.rotaryNames
} else {
    destinations = road.destinations
    names = road.rotaryNames
}
```

I don’t consider this to be “mangling”, since you can easily reverse this logic:

```swift
if maneuverType == .rotary {
    rotaryName = step.destinations.componentsJoined(by: ";")
    name = step.names.componentsJoined(by: ";")
} else {
    destination = step.destinations.componentsJoined(by: ";")
    name = step.names.componentsJoined(by: ";")
}
```

The only problematic case would be for a rotary tagged with `destination`, but the `destination` tag seems ill-suited to rotaries anyways.

You’ll probably have a lot more difficulty reversing the logic in #91 that deals with `ref`, but again, the point was never to match the Directions API response one-for-one.

freenerd:
`destinations = road.names ?? road.destinations`

This will be problematic for instructions, since we loose semantic if `destinations` is names or destinations. For the difference, see [this example](https://github.com/Project-OSRM/osrm-text-instructions/blob/53c52c1de0489901e0492ce6d1b129e76c61f9a4/instructions/en.json#L315-L316):

```
                   "name": "Enter {rotary_name} and exit onto {way_name}",
                    "destination": "Enter {rotary_name} and exit towards {destination}"
```

Take this example:

```js
{
  "rotary_name": "Dupont Circle",
  "name":  "6th Avenue",
}
```

If we only expose `destination` then the correct instruction `Enter Dupont Circle and exit onto 6th Avenue` would become `Enter Dupont Circle and exit towards 6th Avenue`.

1ec5:
> If we only expose `destination` then the correct instruction `Enter Dupont Circle and exit onto 6th Avenue` would become `Enter Dupont Circle and exit towards 6th Avenue`.

As I pointed out in https://github.com/mapbox/MapboxDirections.swift/pull/93#issuecomment-259525631, I think you can safely ignore the destination for rotary instructions.

1ec5:
An alternative would be to put `rotary_name` into the `name` property and `name` into a new `exitName` property (so named for consistency with `exitIndex`). If we go that route, we should consider putting `name` into `exitName` and leaving `name` empty for roundabout instructions.

freenerd:
@1ec5 This sounds good to me, since all information is carried through and no semantics are lost. To paraphrase:

```swift
if type == .rotary || type == .roundabout {
  step.name = json.rotary_name
  step.destinations = json.destinations
  step.exitName = step.name // empty on .roundabout
}
```

1ec5:
Yes, that’s the idea. My only reservation is that there’s already a proliferation of road attribute properties on RouteStep. Ideally we’d shunt them all into a struct, but that won’t bridge to Objective-C; meanwhile, an Objective-C class would be overkill. But if it helps us rationalize our naming logic, then so be it.

I’m assuming we don’t want this behavior for `TurnAtRoundabout` since it’s supposed to be similar to an ordinary intersection.

1ec5:
bcca7e049f3b7146168920fd885b779c10b60777 implements the approach described in https://github.com/mapbox/MapboxDirections.swift/pull/93#issuecomment-259670333. Ready for review.

MapboxDirections/MBRouteStep.swift
```diff
@@ -170,16 +170,32 @@ public enum ManeuverType: Int, CustomStringConvertible {
...
      
-     The exit index indicates the number of roundabout exits up to and including the exit that the user must take.
+     The step’s name is the name of the road to take after exiting the roundabout. The exit index indicates the number of roundabout exits up to and including the exit to take.
      */
     case TakeRoundabout
     
     /**
+     The step requires the user to enter, traverse, and exit a large, named roundabout (traffic circle or rotary).
+     
+     The step’s name is the name of the road to take after exiting the roundabout. The exit index indicates the number of rotary exits up to and including the exit that the user must take.
```
freenerd: I think this is different now with `exitNames`.

> The step's name is the name of the rotary itself. The exit name is the name of the road to take after exiting the roundabout.

PR #94, opened 2016-11-09
Make instructions optional

freenerd:
These are specific to Mapbox Diretions but it would be great
to stay compatible with vanilla OSRM responses as well

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -385,7 +385,7 @@ public class RouteStep: NSObject, NSSecureCoding {
         destinations = json["destinations"] as? String
         
         let maneuver = json["maneuver"] as! JSONDictionary
-        instructions = maneuver["instruction"] as! String
+        instructions = maneuver["instruction"] as? String
         
         distance = json["distance"] as? Double ?? 0
         expectedTravelTime = json["duration"] as? Double ?? 0
@@ -574,14 +574,14 @@ public class RouteStep: NSObject, NSSecureCoding {
     // MARK: Getting Details About the Maneuver
     
     /**
-     A string with instructions in English explaining how to perform the step’s maneuver.
+     A string with instructions explaining how to perform the step’s maneuver.
      
      You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties.
      */
-    public let instructions: String
+    public let instructions: String?
     
     public override var description: String {
-        return instructions
+        return instructions ?? maneuverType?.description ?? maneuverDirection?.description ?? "step"
     }
     
     /**
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -386,18 +386,16 @@ public class RouteStep: NSObject, NSSecureCoding {
         
         let maneuver = json["maneuver"] as! JSONDictionary
 
-        if maneuver["instruction"] != nil {
-            instructions = maneuver["instruction"] as! String
+        if let instructions = maneuver["instruction"] as? String {
+            self.instructions = instructions
+        } else if let mt = maneuverType, md = maneuverDirection {
+            instructions = "\(mt) \(md)"
+        } else if let mt = maneuverType {
+            instructions = String(mt)
+        } else if let md = maneuverDirection {
+            instructions = String(md)
         } else {
-            if let mt = maneuverType?.description, let md = maneuverDirection?.description {
-                instructions = "\(mt) \(md)"
-            } else if let mt = maneuverType?.description {
-                instructions = mt
-            } else if let md = maneuverDirection?.description {
-                instructions = md
-            } else {
-                instructions = ""
-            }
+            instructions = ""
         }
 
         distance = json["distance"] as? Double ?? 0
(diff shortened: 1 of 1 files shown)


MapboxDirections/MBRouteStep.swift
```diff
@@ -527,7 +527,7 @@ public class RouteStep: NSObject, NSSecureCoding {
         destinations = json["destinations"] as? String
         
         let maneuver = json["maneuver"] as! JSONDictionary
-        instructions = maneuver["instruction"] as! String
+        instructions = maneuver["instruction"] as? String
```
1ec5: The `instructions` property is non-optional, so this produces a compiler error.

MapboxDirections/MBRouteStep.swift
```diff
@@ -428,14 +428,14 @@ public class RouteStep: NSObject, NSSecureCoding {
...
+     A string with instructions explaining how to perform the step’s maneuver.
      
      You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties.
      */
-    public let instructions: String
+    public let instructions: String?
     
     public override var description: String {
-        return instructions
+        return instructions ?? maneuverType?.description ?? maneuverDirection?.description ?? "step"
```
freenerd: @1ec5 is `.description` used for more than debug output?

1ec5: It isn’t supposed to be, but we do use it as sort of a back-to-JSON thing in this library.

1ec5: `description` comes from CustomStringConvertible. In general, in Swift, it can be used for non-debugging purposes but shouldn’t be relied on to be stable or user-friendly. `debugDescription`, from CustomDebugStringConvertible, is purely for debugging purposes.

1ec5: In the absence of `instructions`, the description should be an English string containing basic information about the step, enough to identify it while debugging.

MapboxDirections/MBRouteStep.swift
```diff
@@ -428,14 +428,14 @@ public class RouteStep: NSObject, NSSecureCoding {
     // MARK: Getting Details About the Maneuver
     
     /**
-     A string with instructions in English explaining how to perform the step’s maneuver.
+     A string with instructions explaining how to perform the step’s maneuver.
      
      You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties.
```
1ec5: Explain that instructions that come from the Mapbox Directions API are always in English, and that osrm-text-instructions (with an HTML link) should be used for localized, user-friendly instructions.

MapboxDirections/MBRouteStep.swift
1ec5: Does this really have to be optional? That makes the library less convenient for the most common use case, which involves the Mapbox Directions API. Can non-Mapbox uses make do with an empty string or a very basic generated sentence?

freenerd:
@1ec5 https://github.com/mapbox/MapboxDirections.swift/pull/94/commits/c26d0ba12696ea10d9e97776fb4b7993d69cb64e has another option: Always populate `RouteStep.instructions` in case the json has no instruction. This works for OSRM responses, but makes downstream work easier. There are cases where there is no type or modifier, in these cases the returned string will be less descriptive (worst case: empty) but types won't break.

MapboxDirections/MBRouteStep.swift
```diff
@@ -428,11 +428,13 @@ public class RouteStep: NSObject, NSSecureCoding {
     // MARK: Getting Details About the Maneuver
     
     /**
-     A string with instructions in English explaining how to perform the step’s maneuver.
-     
-     You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties.
+     A string with instructions explaining how to perform the step’s maneuver.
+
+     You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties via https://github.com/Project-OSRM/osrm-text-instructions
```
1ec5: Make this a Markdown link, which is recognized by Quick Help in Xcode.

MapboxDirections/MBRouteStep.swift
```diff
@@ -428,11 +428,13 @@ public class RouteStep: NSObject, NSSecureCoding {
...
+
+     Note: This string is populated by the Mapbox Directions API only, if you use OSRM it will only contain the type and modifier. Use osrm-text-instruction to generate instructions then.
      */
-    public let instructions: String
+    public let instructions: String!
```
1ec5: If this property is always set, it doesn't need to be implicitly optional.

MapboxDirections/MBRouteStep.swift
```diff
@@ -428,11 +428,13 @@ public class RouteStep: NSObject, NSSecureCoding {
...
     
     /**
-     A string with instructions in English explaining how to perform the step’s maneuver.
-     
-     You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties.
+     A string with instructions explaining how to perform the step’s maneuver.
+
+     You can display this string or read it aloud to the user. The string does not include the distance to or from the maneuver. If you need to localize or otherwise customize the instructions, you can construct the instructions yourself using the step’s other properties via https://github.com/Project-OSRM/osrm-text-instructions
+
+     Note: This string is populated by the Mapbox Directions API only, if you use OSRM it will only contain the type and modifier. Use osrm-text-instruction to generate instructions then.
```
1ec5: Use `- note:`. Swift uses reStructuredText as its documentation comment format.

MapboxDirections/MBRouteStep.swift
```diff
@@ -527,7 +529,7 @@ public class RouteStep: NSObject, NSSecureCoding {
         destinations = json["destinations"] as? String
         
         let maneuver = json["maneuver"] as! JSONDictionary
-        instructions = maneuver["instruction"] as! String
+        instructions = maneuver["instruction"] as? String ?? "\(maneuverType) \(maneuverDirection)"
```
1ec5: `maneuverType` and `maneuverDirection` are optional, so string interpolation will turn them into e.g. `Optional(turn) Optional(left)`. Instead, use an `if let` statement to conditionally populate `instructions` using unwrapped values.

PR #95, opened 2016-11-09
Round-trip right and none lane indications

1ec5:
Fixed round-tripping of `right` and `none` lane indications back to the JSON string value.

/cc @bsudekum @freenerd


1ec5:
Merged to the swift3 branch in 772cd0743ecc6885cc6dd836554d8688c5b3f0bd.

PR #98, opened 2016-11-15
Swift3 updates

bsudekum:
Incorporates many recent changes done in master to swift 3 branch.

/cc @1ec5


1ec5:
What's with 599b13f05c6cefb69f2fc67aeaf275bdc31c65c4? It seems to be repeat some of the other commits. It also downgrades the CocoaPods version, but I guess that's OK.

1ec5:
Try removing 599b13f05c6cefb69f2fc67aeaf275bdc31c65c4 from this branch. I think you may’ve committed some files in the middle of resolving merge conflicts.

PR #99, opened 2016-11-16
Always retain ref if exist

freenerd:
An oversight in https://github.com/mapbox/MapboxDirections.swift/pull/91

It would actually be nice to UnitTest the `Road`class to document all cases we expect.

### DIFF
--- a/RouteStepTests.swift
+++ b/RouteStepTests.swift
@@ -1,6 +1,96 @@
 import XCTest
 @testable import MapboxDirections
 
+class RoadTests: XCTestCase {
+    func testEmpty() {
+        let r = Road(name: "", ref: nil, destination: nil, rotaryName: nil)
+        XCTAssertNil(r.names)
+        XCTAssertNil(r.codes)
+        XCTAssertNil(r.destinations)
+        XCTAssertNil(r.destinationCodes)
+        XCTAssertNil(r.rotaryNames)
+    }
+
+    func testNamesCodes() {
+        var r: Road
+
+        // Name only
+        r = Road(name: "Way Name", ref: nil, destination: nil, rotaryName: nil)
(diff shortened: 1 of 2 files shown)


freenerd:
Added a test suite because I like tests.

RouteStepTests.swift
```diff
@@ -1,6 +1,93 @@
...
+    }
+
+    func testNamesCodes() {
+        var r: Road
+
+        // Name only
+        r = Road(name: "Way Name", ref: nil, destination: nil, rotaryName: nil)
+        XCTAssertEqual(r.names!, [ "Way Name" ])
+        XCTAssertNil(r.codes)
+        r = Road(name: "Way Name 1; Way Name 2", ref: nil, destination: nil, rotaryName: nil)
```
1ec5: Add a test for values delimited by a colon but not a space, since that's the usual OSM tagging convention.

RouteStepTests.swift
```diff
@@ -1,6 +1,93 @@
...
+        XCTAssertNil(r.destinationCodes)
+        XCTAssertNil(r.rotaryNames)
+    }
+
+    func testNamesCodes() {
+        var r: Road
+
+        // Name only
+        r = Road(name: "Way Name", ref: nil, destination: nil, rotaryName: nil)
+        XCTAssertEqual(r.names!, [ "Way Name" ])
```
1ec5: Force-unwrapping names here will cause the test case to crash on this line if names is nil for some reason. Instead, fall back to an empty array so that this assertion fails but the test can continue and report about other failures.

freenerd:
@1ec5 thanks, feedback incorporated

freenerd:
This also made it into the `swift3` branch now https://github.com/mapbox/MapboxDirections.swift/pull/64/commits/ceaf58b780fc17ea44a9150041b602d017c1e567

PR #108, opened 2017-02-19
Upgrade project to Xcode 8.2

1ec5:
Another take at #71. Upgraded to Xcode 8.2, CocoaPods 1.2, iOS SDK v3.4.1, and OHHTTPStubs v5.2.3. Fixed CocoaPods warnings about `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` being overridden. Set `BITCODE_GENERATION_MODE` to `bitcode` so that applications that link against this library can archive.

/cc @frederoni


1ec5:
I got around the watchOS failures I was experiencing on Bitrise in #71 by replacing the Xcode Analyze or Xcode Test step with a script step that simply performs the `xcodebuild` invocation at the heart of the standard Xcode Analyze step but with an extra `-destination` flag.

PR #115, opened 2017-03-12
Polyline v4.1.1

1ec5:
Upgraded to a release of Polyline, now that raphaelmor/Polyline#40 has landed.

/cc @frederoni

### DIFF
--- a/MapboxDirections.swift.podspec
+++ b/MapboxDirections.swift.podspec
@@ -43,6 +43,6 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxDirections"
 
-  s.dependency "Polyline", "~> 4.0"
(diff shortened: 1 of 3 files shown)


MapboxDirections.swift.podspec
```diff
@@ -43,6 +43,6 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.module_name = "MapboxDirections"
 
-  s.dependency "Polyline", "~> 4.0"
+  s.dependency "Polyline", "~> 4.1"
```
1ec5: It’s customary to specify only major and minor version here. The only reason I’ve specified a patch version in the Cartfile is because Carthage aggressively builds for all platforms and the tvOS and watchOS builds were broken in v4.1.0.

PR #118, opened 2017-03-21
Attributes

bsudekum:
Adds optional query param `annotation`. This is an array of annotation types, allowed enum values: `.distance`, `.duration`, `.speed`, `.node`.

Still need to add documentation.

/cc @1ec5


MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,109 @@
+//
+//  MBAnnotationType.swift
+//  MapboxDirections
+//
+//  Created by Bobby Sudekum on 3/20/17.
+//  Copyright © 2017 Mapbox. All rights reserved.
```
1ec5: Omit this file header comment. This is a permissively licensed project under version control, so the comment adds little value.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +130,9 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     open let steps: [RouteStep]
     
+    
+    open let annotation: [AnnotationType:[Any]]?
```
1ec5: Annotation is a poor name for this information for the following reasons:

* It's a collection of data, so a plural word is preferable.
* The Mapbox iOS and macOS SDKs already have a concept of annotations (for consistency with MapKit) that's completely different: markers and overlays on a map. Google Maps calls them markers, but developers are required to use the Mapbox SDKs with this library anyways.
* The term doesn't say anything about the relationship between the route leg and the information indicated by the property.

Some suggestions:

* nodeAttributes (so far only for the nodeIdentifiers key) and segmentAttributes (for the other keys)
* attributes
* userInfo – Although this may sound strange, it's consistent with Cocoa APIs like NSError and NSNotification.

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,109 @@
...
+@objc(AnnotationType)
+public enum AnnotationType: Int, CustomStringConvertible {
+
+    case congestion
+    
+    case distance
+    
+    case duration
+    
+    case nodes
```
1ec5: (I'm going to call annotation types "attributes" for the rest of this review; see below.)

Let's call this `openStreetMapNodeIdentifier`, similar to how MapboxGeocoder.swift has a `wikidataItemIdentifier`. If on the other hand we expect the same underlying string to be used for non-OSM data sources, `nodeIdentifier` would be better. (I'd like to avoid using "node" by itself, because perhaps we may associate more attributes with OSM nodes in the future.)

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,109 @@
...
+//  Created by Bobby Sudekum on 3/20/17.
+//  Copyright © 2017 Mapbox. All rights reserved.
+//
+
+import Foundation
+
+@objc(AnnotationType)
+public enum AnnotationType: Int, CustomStringConvertible {
+
+    case congestion
```
1ec5: "Congestion" is a strange word to use on its own, since it's uncountable. (Imagine talking about "congestions".) Let's call it `congestionLevel`.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -31,7 +34,32 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     public convenience init(json: [String: Any], source: Waypoint, destination: Waypoint, profileIdentifier: MBDirectionsProfileIdentifier) {
         let steps = (json["steps"] as? [JSONDictionary] ?? []).map { RouteStep(json: $0) }
-        self.init(steps: steps, json: json, source: source, destination: destination, profileIdentifier: profileIdentifier)
+        
+        var annotation: Annotation = [:]
+        if json["annotation"] != nil {
+            let rawAnnoation = json["annotation"] as! [String: Any]
```
1ec5: s/annoation/annotation/

MapboxDirections/MBRouteLeg.swift
```diff
@@ -31,7 +34,32 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+            if rawAnnoation["congestion"] != nil {
+                annotation[.congestion] = (rawAnnoation["congestion"] as! [String]).map { CongestionLevel(description: $0) ?? .unknown}
```
1ec5: To avoid repeating yourself, use an `if let … as? …` above`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -256,6 +258,15 @@ open class RouteOptions: NSObject {
             params.append(URLQueryItem(name: "radiuses", value: accuracies))
         }
         
+        
+        if annotation.count > 0 {
```
1ec5: Nit: `!annotation.isEmpty`

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,8 @@ open class RouteOptions: NSObject {
      */
     open var routeShapeResolution = RouteShapeResolution.low
     
+    open var annotation: [AnnotationType] = []
```
1ec5: `attributes`

MapboxDirections/MBRouteLeg.swift
```diff
@@ -40,6 +68,9 @@ open class RouteLeg: NSObject, NSSecureCoding {
         }
         source = decodedSource
         
+
+        annotation = nil
```
1ec5: The encode method encodes the attribute data, so the decode method should decode it.

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,109 @@
...
+    
+    case distance
+    
+    case duration
```
1ec5: "Durations" can indicate lots of things. Let's call this `expectedTravelTime`, for consistency with the `expectedTravelTime` property of RouteLeg and RouteStep.

MapboxDirections/MBRouteLeg.swift
1ec5: With a split between "node attributes" and "segment attributes", we'll eventually be able to distribute the attribute values among each step (while still retaining the full, leg-level attribute structure as an option).

1ec5: A little trivia: "[segment](https://wiki.openstreetmap.org/wiki/Segment)" used to be a primitive data type in OpenStreetMap back in the day.

bsudekum:
@1ec5 updated with feedback

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -15,6 +15,8 @@
 		358D48871E2EAB4500F32A65 /* MBRouteOptions.m in Sources */ = {isa = PBXBuildFile; fileRef = 358D48851E2EAB4500F32A65 /* MBRouteOptions.m */; };
 		358D48881E2EAB4500F32A65 /* MBRouteOptions.m in Sources */ = {isa = PBXBuildFile; fileRef = 358D48851E2EAB4500F32A65 /* MBRouteOptions.m */; };
 		358D48891E2EAB4500F32A65 /* MBRouteOptions.m in Sources */ = {isa = PBXBuildFile; fileRef = 358D48851E2EAB4500F32A65 /* MBRouteOptions.m */; };
+		C51538CC1E807FF00093FF3E /* MBAnnotation.swift in Sources */ = {isa = PBXBuildFile; fileRef = C51538CB1E807FF00093FF3E /* MBAnnotation.swift */; };
```
1ec5: This file needs to be added to all four framework targets. Because it’s only in the iOS framework target, macOS, tvOS, and watchOS builds are failing with numerous errors, starting with:

```
▸ Compiling MBRouteLeg.swift

❌  /Users/vagrant/git/MapboxDirections/MBRouteLeg.swift:11:36: use of undeclared type 'SegmentAttribute'

    typealias SegmentAttributes = [SegmentAttribute:[Any]]
                                   ^~~~~~~~~~~~~~~~



❌  /Users/vagrant/git/MapboxDirections/MBRouteLeg.swift:129:34: use of undeclared type 'SegmentAttribute'

    open let segmentAttributes: [SegmentAttribute:[Any]]?
                                 ^~~~~~~~~~~~~~~~



❌  /Users/vagrant/git/MapboxDirections/MBRouteLeg.swift:173:14: cannot invoke 'RouteLegV4.init' with an argument list of type '(steps: [RouteStepV4], json: JSONDictionary, source: Waypoint, destination: Waypoint, profileIdentifier: MBDirectionsProfileIdentifier)'

        self.init(steps: steps, json: json, source: source, destination: destination, profileIdentifier: profileIdentifier)
             ^
```

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -15,6 +15,8 @@
...
+		C5247D711E818A24004B6154 /* AnnotationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C5247D701E818A24004B6154 /* AnnotationTests.swift */; };
```
1ec5: Likewise, this file needs to be added to all four unit test targets.

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+@objc(AnnotationType)
```
1ec5: This type needs a documentation comment.

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,45 @@
...
+public enum SegmentAttribute: Int, CustomStringConvertible {
+    
+    case distance
```
1ec5: Each of these cases needs a documentation comment.

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+@objc(AnnotationType)
+public enum SegmentAttribute: Int, CustomStringConvertible {
```
1ec5: This type should be called simply `Attribute`, because `openStreetMapNodeIdentifier` is a node attribute, not a segment attribute.

MapboxDirections/MBAnnotation.swift
```diff
@@ -0,0 +1,45 @@
...
+    case distance
+    
+    case expectedTravelTime
+    
+    case openStreetMapNodeIdentifier
+    
+    case speed
+    
+    
+    public init?(description: String) {
```
1ec5: Since this initializer is public, it needs a documentation comment.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -7,16 +7,19 @@ import Polyline
...
-    internal init(steps: [RouteStep], json: JSONDictionary, source: Waypoint, destination: Waypoint, profileIdentifier: MBDirectionsProfileIdentifier) {
+    internal init(steps: [RouteStep], json: JSONDictionary, source: Waypoint, destination: Waypoint, profileIdentifier: MBDirectionsProfileIdentifier, segmentAttributes: SegmentAttributes? = nil) {
         self.source = source
         self.destination = destination
         self.profileIdentifier = profileIdentifier
         self.steps = steps
         distance = json["distance"] as! Double
         expectedTravelTime = json["duration"] as! Double
         self.name = json["summary"] as! String
+        self.segmentAttributes = segmentAttributes
```
1ec5: There should be two separate properties: `nodeAttributes` and `segmentAttributes`. `nodeAttributes` should contain a `openStreetMapNodeIdentifier` key, and `segmentAttributes` should contain the other keys.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +125,9 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     open let steps: [RouteStep]
     
+    
+    open let segmentAttributes: [SegmentAttribute:[Any]]?
```
1ec5: Use the `SegmentAttributes` typealias here.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -40,6 +63,9 @@ open class RouteLeg: NSObject, NSSecureCoding {
         }
         source = decodedSource
         
+
+        segmentAttributes = nil
```
1ec5: https://github.com/mapbox/MapboxDirections.swift/pull/118#discussion_r107310563

MapboxDirections/MBRouteOptions.swift
```diff
@@ -256,6 +258,16 @@ open class RouteOptions: NSObject {
             params.append(URLQueryItem(name: "radiuses", value: accuracies))
         }
         
+        
+        if segmentAttributes.count > 0 {
```
1ec5: https://github.com/mapbox/MapboxDirections.swift/pull/118#discussion_r107310361

MapboxDirections/MBRouteOptions.swift
```diff
@@ -256,6 +258,16 @@ open class RouteOptions: NSObject {
...
+            let segmentAttributesStrings = segmentAttributes.map {
+                $0.description
```
1ec5: `String(describing: $0)`

MapboxDirectionsTests/AnnotationTests.swift
```diff
@@ -0,0 +1,74 @@
...
+        XCTAssertNotNil(route!.coordinates)
+        XCTAssertEqual(route!.coordinates!.count, 93)
+        
+        let leg = route!.legs.first!
+        let annotation = leg.segmentAttributes!
+        XCTAssertEqual(annotation.count, 4)
+        XCTAssertNotNil(annotation[.distance])
+        XCTAssertNotNil(annotation[.expectedTravelTime])
+        XCTAssertNotNil(annotation[.openStreetMapNodeIdentifier])
+        XCTAssertNotNil(annotation[.speed])
```
1ec5: These tests should also check specific attribute values.

MapboxDirectionsTests/AnnotationTests.swift
```diff
@@ -0,0 +1,74 @@
...
+        
+        let nodes = annotation[.openStreetMapNodeIdentifier]!.count
```
1ec5: Check that there are the right number of node identifier attributes.

MapboxDirections/MBRouteLeg.swift
1ec5: There should also be two separate properties in RouteOptions, even if they end up being combined into a single `annotations` parameter. It’s unfortunate that the API conflates these two concepts, but there’s no reason we need to extend the confusion to the client library.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +129,11 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     open let steps: [RouteStep]
     
+    
+    open let segmentAttributes: [Attribute:[Any]]?
```
1ec5: This property won't bridge to Objective-C, because NSDictionary's keys need to be objects, not enumeration values. (Option sets won't bridge either.) We'll probably want to use an Objective-C string enumeration (`NS_STRING_ENUM`) instead of the Swift enumeration. It would be separate from the option set used in RouteOptions.

If that's too much trouble, RouteLeg could have a separate property for each attribute. For example, `segmentSpeeds` would be of type `[CLLocationSpeed]`. That would be pretty clean, but I don't know what the story will be for any unsupported attributes we get back from the API.

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,12 @@
...
+
+    MBAttributeDistance = (1 << 1),
+    
+    MBAttributeExpectedTravelTime = (1 << 2),
+    
+    MBAttributeOpenStreetMapNodeIdentifier = (1 << 3),
+    
+    MBAttributeSpeed = (1 << 4),
+    
+    MBAttributeAll = 0x0ffffUL
```
1ec5: We probably don’t need an “all” option for attributes.

bsudekum:
@1ec5 this is ready for another round of 👀

Cartfile.resolved
```diff
@@ -1,3 +1,3 @@
-binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "3.4.2"
-github "AliSoftware/OHHTTPStubs" "57feceaabf333e72b2c637dfba6c13a7a5c49619"
+binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" "3.5.0"
+github "AliSoftware/OHHTTPStubs" "f90c2bb0fb882e43761ab963ca8869d349d2c6e3"
```
1ec5: This is fine, but in the future, let’s split out any changes that occur due to running `carthage update` into separate PRs, so that we can narrow down any regressions they cause. If you just need to get set up, use `carthage bootstrap` instead.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -10,16 +10,39 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
         self.destination = destination
         self.profileIdentifier = profileIdentifier
         self.steps = steps
         distance = json["distance"] as! Double
         expectedTravelTime = json["duration"] as! Double
         self.name = json["summary"] as! String
-        self.segmentAttributes = segmentAttributes
-        self.nodeAttributes = nodeAttributes
+        
+        var segmentDistances: [CLLocationDistance]? = nil
```
1ec5: Nit: an optional is `nil` by default; no need to set it to `nil`.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -129,10 +134,13 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
     
+    open let segmentDistances: [CLLocationDistance]?
+    
+    open let segmentExpectedTravelTimes: [TimeInterval]?
     
-    open let segmentAttributes: [Attribute:[Any]]?
+    open let segmentSpeeds: [CLLocationSpeed]?
     
-    open let nodeAttributes: [Attribute:[Any]]?
+    open let nodeOpenStreetMapNodeIdentifiers: [Int]?
```
1ec5: To avoid overflow, OSM node IDs [must be stored as 64-bit integers](http://textual.ru/64/). [`Int`](https://developer.apple.com/reference/swift/int) can be either 32-bit or 64-bit according to the device’s architecture. To be on the safe side and to clarify our intent, use `Int64`.

1ec5: “Node” is in the name twice, unnecessarily: rename this property to `openStreetMapNodeIdentifiers`.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -129,10 +134,13 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     open let steps: [RouteStep]
     
+    open let segmentDistances: [CLLocationDistance]?
+    
+    open let segmentExpectedTravelTimes: [TimeInterval]?
```
1ec5: The name of this property makes it sound like a verb, “to segment the expected travel times”. Rename this property to `expectedSegmentTravelTimes`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -260,22 +260,21 @@ open class RouteOptions: NSObject {
...
         if !nodeAttributes.isEmpty && nodeAttributes.contains(.openStreetMapNodeIdentifier) {
-            let segmentAttributesStrings = segmentAttributes.map {
+            attributeOptions = attributeOptions + nodeAttributes
+        }
+        
+        if !attributeOptions.isEmpty {
+            let attributesStrings = attributeOptions.map {
                 String(describing: $0)
-            }.joined(separator: ",")
+                }.joined(separator: ",")
```
1ec5: `Attribute.description` is [already a comma-separated list](https://github.com/mapbox/MapboxDirections.swift/blob/8fb56c4a5b09021e229e870b58726bd62c963198/MapboxDirections/MBAttribute.swift#L38), so you can set `attributesStrings` to `String(describing: segmentAttributes + .openStreetMapNodeIdentifier)` to get the same effect.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,10 @@ open class RouteOptions: NSObject {
      */
     open var routeShapeResolution = RouteShapeResolution.low
     
+    open var segmentAttributes: [Attribute] = []
```
1ec5: Now that Attribute is an OptionSet, we no longer need an array of Attributes; a single Attribute can store multiple options.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,10 @@ open class RouteOptions: NSObject {
...
+    
+    open var nodeAttributes: [Attribute] = []
```
1ec5: Compared to the situation in RouteLeg before, there’s no need to distinguish between segment and node attributes in RouteOptions – the two behave exactly the same as far as requesting a route is concerned. A single `attributeOptions` property will do.

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,10 @@
+typedef NS_OPTIONS(NSUInteger, MBAttribute) {
```
1ec5: One more rename! Now that this is an option set instead of a single attribute, rename this type to MBAttributeOptions, so it’s easier to talk about individual attributes versus the whole set of attributes.

Cartfile.resolved
bsudekum: I'll revert this

bsudekum:
@1ec5 fixed

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,10 @@ open class RouteOptions: NSObject {
      */
     open var routeShapeResolution = RouteShapeResolution.low
     
+    open var segmentAttributes: AttributeOptions = []
+    
+    open var nodeAttributes: AttributeOptions = []
```
1ec5: Combine these two properties into a single attributes property.

bsudekum:
Started documentation in https://github.com/mapbox/MapboxDirections.swift/pull/118/commits/5c2c5c03483c1743a8c8888abb82b6cb388adab7

MapboxDirections/MBAttribute.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+public typealias AttributeOptions = MBAttributeOptions
+
+extension AttributeOptions: CustomStringConvertible {
+    /**
+     Creates an AttributeOptions from the given description strings.
+     */
+    public init?(description: String) {
+         var scope: AttributeOptions = []
```
1ec5: Nit: funky indentation.

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,17 @@
+/**
+ Attributes are metadata information for a given route. The number of attributes returned will be a direct 1-1 relationship with the route's full geometry. Each type will return an ordered list of requested attributes. For `.distance`, `.expectedTrabelTime`, and `.speed` there will be one less value when compared to the route geometry. This is because these values represent the data on segment between geometry points.
+*/
+typedef NS_OPTIONS(NSUInteger, MBAttributeOptions) {
+
+    /// Segment distance.
```
1ec5: Distances are measured in meters.

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,17 @@
...
+    MBAttributeDistance = (1 << 1),
+    
+    // Segment expected travel time in seconds.
+    MBAttributeExpectedTravelTime = (1 << 2),
+    
+    // Segment current speed.
```
1ec5: Speeds are measured in meters per second.

MapboxDirections/MBAttribute.h
1ec5: Use plurals (“distances”), because this option provides the developer with multiple distance values.

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,17 @@
...
+    MBAttributeSpeed = (1 << 3),
+    
+    // Unique OpenStreetMap node.
```
1ec5: This option returns unique node _identifiers_, not the nodes themselves. Link to [this OpenStreetMap Wiki article](https://wiki.openstreetmap.org/wiki/Node) for more information. Also note that this attribute option is added automatically when any of the other attributes are added.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +134,26 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     open let steps: [RouteStep]
     
+    /**
+     An array of distances representing the distances between nodes.
```
1ec5: > An array containing the distance (measured in meters) between each coordinate in the route leg geometry.
> 
> This property is set if the `RouteOptions.attributeOptions` property contains `.distance`.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +134,26 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+     */
+    open let segmentDistances: [CLLocationDistance]?
+    
+    /**
+     An array of expected travel times for the space between each geometry node. This value is dynamic and accounts for any and all conditions that may change along current segment.
```
1ec5: > An array containing the expected travel time (measured in seconds) between each coordinate in the route leg geometry.
> 
> These values are dynamic, accounting for any conditions that may change along a segment, such as traffic congestion if the profile identifier is set to `.automobileAvoidingTraffic`.
> 
> This property is set if the `RouteOptions.attributeOptions` property contains `.expectedTravelTime`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,13 @@ open class RouteOptions: NSObject {
      */
     open var routeShapeResolution = RouteShapeResolution.low
     
+    /**
+     AttributeOptions for the route. Any combination of `.AttributeOptions` can be specified.
```
1ec5: Remove the extra dot: `.AttributeOptions` implies that there’s an `AttributeOptions.AttributeOptions`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,13 @@ open class RouteOptions: NSObject {
      */
     open var routeShapeResolution = RouteShapeResolution.low
     
+    /**
+     AttributeOptions for the route. Any combination of `.AttributeOptions` can be specified. 
+     
+     The default value of this property is `[]`.
```
1ec5: > By default, no attribute options are specified.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -211,6 +211,13 @@ open class RouteOptions: NSObject {
      */
     open var routeShapeResolution = RouteShapeResolution.low
     
+    /**
+     AttributeOptions for the route. Any combination of `.AttributeOptions` can be specified. 
+
```
1ec5: Note that this option requires the `routeShapeResolution` property to be set to `.full`. (The default is `.low`.)

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +134,26 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+     */
+    open let expectedSegmentTravelTimes: [TimeInterval]?
+    
+    /**
+     An array of current speeds along segment. This value is dynamic and does not represent the speed limit but rather the average speed for a given segment
```
1ec5: > An array containing the expected average speed (measured in meters per second) between each coordinate in the route leg geometry.
> 
> These values are dynamic; rather than speed limits, they account for the road’s classification and/or any traffic congestion (if the profile identifier is set to `.automobileAvoidingTraffic`).
> 
> This property is set if the `RouteOptions.attributeOptions` property contains `.speed`.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +134,26 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+     */
+    open let segmentSpeeds: [CLLocationSpeed]?
+    
+    /**
+     An array of OpenStreetMap nodes.
```
1ec5: > An array containing [OpenStreetMap node identifiers](https://wiki.openstreetmap.org/wiki/Node), one for each coordinate along the route geometry.
> 
> This property is set if the `RouteOptions.attributeOptions` property contains `.openStreetMapNodeIdentifier`.

MapboxDirections/MBRouteLeg.swift
1ec5: (This is a little awkward, because RouteLeg itself doesn’t contain the geometry; only Route and RouteStep do.)

MapboxDirections/MBAttribute.h
bsudekum: > Also note that this attribute option is added automatically when any of the other attributes are added.

This is not true: 
image

MapboxDirections/MBRouteOptions.swift
bsudekum: Unfortunately, this is not true. This should be ticketed upstream.

1ec5: Can you elaborate? Do the attributes still work with `.low` resolution? Do the values make any sense?

MapboxDirections/MBAttribute.h
1ec5: Ah, I thought we had baked that logic into RouteOptions, but it seems to have been removed. 👍

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,17 @@
+/**
+ Attributes are metadata information for a given route. The number of attributes returned will be a direct 1-1 relationship with the route's full geometry. The .distance, .expectedTrabelTime, and .speed attributes have one fewer value than the .openStreetMapNodeIdentifier attribute.
```
1ec5: Typo: s/expectedTrabelTime/expectedTravelTime/

1ec5: Surround references like `.distance` in backticks so jazzy can autolink them.

MapboxDirections/MBAttribute.h
```diff
@@ -0,0 +1,17 @@
...
+    /// Segment distance. Distances are measured in meters.
+    MBAttributeDistance = (1 << 1),
+    
+    // Segment expected travel time in seconds.
+    MBAttributeExpectedTravelTime = (1 << 2),
+    
+    // Segment current speed. Speeds are measured in meters per second.
+    MBAttributeSpeed = (1 << 3),
+    
+    // Unique node identifiers (https://wiki.openstreetmap.org/wiki/Node)
```
1ec5: Use Markdown syntax:

```markdown
[OpenStreetMap node identifier](https://wiki.openstreetmap.org/wiki/Node).
```

MapboxDirections/MBRouteLeg.swift
```diff
@@ -98,6 +134,38 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+     An array containing the expected average speed (measured in meters per second) between each coordinate in the route leg geometry.
+     
+     These values are dynamic; rather than speed limits, they account for the road’s classification and/or any traffic congestion (if the profile identifier is set to `.automobileAvoidingTraffic`).
+     
+     This property is set if the `RouteOptions.attributeOptions` property contains `.speed`.
+     */
+    open let segmentSpeeds: [CLLocationSpeed]?
+    
+    /**
+     An array containing OpenStreetMap node identifiers (https://wiki.openstreetmap.org/wiki/Node), one for each coordinate along the route geometry.
```
1ec5: Use Markdown syntax here too.

MapboxDirections/MBRouteOptions.swift
1ec5: If it’s the case that `.low` resolution yields the same values (which reflect the `.full` geometry), let’s recommend that the developer set `.full` to make use of these options.

bsudekum:
@1ec5 think this is ready to go

PR #119, opened 2017-03-29
Fix swift 3.1 warnings

bsudekum:
Fixes Swift 3.1 warnings when using xcode 8.3.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -557,7 +557,7 @@ open class RouteStep: NSObject, NSSecureCoding {
             return nil
         }
         maneuverType = ManeuverType(description: maneuverTypeDescription)
-        let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection") as! String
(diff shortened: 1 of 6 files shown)


MapboxDirections/MBRouteStep.swift
```diff
@@ -557,7 +557,7 @@ open class RouteStep: NSObject, NSSecureCoding {
             return nil
         }
         maneuverType = ManeuverType(description: maneuverTypeDescription)
-        let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection") as! String
+        let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection")! as String
```
1ec5: Huh, how’d this `!` get here in the first place? This line should read:

```swift
guard let maneuverDirectionDescription = … as String? else { return nil }
```

An `init?(coder:)` method should be able to fail gracefully (hence the `?`), because it could be passed absolutely any object for perfectly legitimate reasons.

PR #122, opened 2017-04-13
Make RouteOptions available on route response

bsudekum:
Instead of returning the profile identifier on the route, this replaces it with the entire RouteOptions used to create the directions request.

/cc @frederoni @1ec5

### DIFF
--- a/MapboxDirections/MBRoute.swift
+++ b/MapboxDirections/MBRoute.swift
@@ -156,9 +156,7 @@ open class Route: NSObject, NSSecureCoding {
     open let expectedTravelTime: TimeInterval
     
     /**
-     A string specifying the primary mode of transportation for the route.
-     
-     The value of this property is `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`, depending on the `profileIdentifier` property of the original `RouteOptions` object. This property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.
+     `RouteOptions` used to create the directions request.
      */
     open let routeOptions: RouteOptions
 }
--- a/MapboxDirections/MBRoute.swift
+++ b/MapboxDirections/MBRoute.swift
@@ -24,7 +24,7 @@ open class Route: NSObject, NSSecureCoding {
      
      - parameter json: A JSON dictionary representation of the route as returned by the Mapbox Directions API.
      - parameter waypoints: An array of waypoints that the route visits in chronological order.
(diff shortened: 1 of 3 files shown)


MapboxDirections/MBRoute.swift
```diff
@@ -26,12 +26,12 @@ open class Route: NSObject, NSSecureCoding {
      - parameter waypoints: An array of waypoints that the route visits in chronological order.
      - parameter profileIdentifier: The profile identifier used to request the routes.
```
1ec5: This parameter has been replaced by `routeOptions`.

MapboxDirections/MBRoute.swift
```diff
@@ -156,18 +156,16 @@ open class Route: NSObject, NSSecureCoding {
     open let expectedTravelTime: TimeInterval
     
     /**
-     A string specifying the primary mode of transportation for the route.
-     
-     The value of this property is `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`, depending on the `profileIdentifier` property of the original `RouteOptions` object. This property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.
```
1ec5: Developers may not be aware of this nuance, so we might want to leave it in:

> The route options object’s `profileIdentifier` property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -52,6 +52,13 @@ class V5Tests: XCTestCase {
...
         }
         
+        let opts = route!.routeOptions
+        
+        XCTAssertEqual(opts.profileIdentifier, .automobile)
+        XCTAssertTrue(opts.includesSteps)
+        XCTAssertTrue(opts.includesAlternativeRoutes)
+        XCTAssertEqual(opts.routeShapeResolution, .full)
+        
         XCTAssertNotNil(route)
```
1ec5: Move this assertion up above any assertions that depend on `route` being non-nil.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -52,6 +52,13 @@ class V5Tests: XCTestCase {
             XCTAssertEqual(task.state, .completed)
         }
         
+        let opts = route!.routeOptions
+        
+        XCTAssertEqual(opts.profileIdentifier, .automobile)
```
1ec5: Might as well assert that `route!.routeOptions` is equal to `options`. That way we can be sure they’re literally the same object; no need to compare individual properties on them.

PR #123, opened 2017-04-13
Attribute congestion

bsudekum:
Adds congestion attribute option.


MapboxDirections/MBCongestion.h
```diff
@@ -0,0 +1,12 @@
+typedef NS_OPTIONS(NSUInteger, MBCongestion) {
```
1ec5: This should be called MBCongestionLevel, for consistency with MBAttributeCongestionLevel.

1ec5: A particular segment can’t have two levels at the same time, so we would want an `NS_ENUM` in Objective-C (`enum` in Swift), not an `NS_OPTIONS` in Objective-C (`OptionSet` in Swift).

Since normal enumerations bridge from Swift to Objective-C just fine, we can implement this as an ordinary Swift enumeration – no need for an Objective-C header.

1ec5:
The `congestion` attribute has been deployed server-side. This PR still needs work, however.

MapboxDirections/MBAttribute.h
```diff
@@ -1,17 +1,41 @@
...
 typedef NS_OPTIONS(NSUInteger, MBAttributeOptions) {
-
-    /// Segment distance. Distances are measured in meters.
-    MBAttributeDistance = (1 << 1),
+    /**
+     [OpenStreetMap node identifier](https://wiki.openstreetmap.org/wiki/Node).
+     
+     When this attribute is specified, the `RouteLeg.openStreetMapNodeIdentifiers` property contains one value for each coordinate in the leg’s full geometry.
+     */
+    MBAttributeOpenStreetMapNodeIdentifier = (1 << 1),
```
1ec5: I’ve changed the underlying values of the various attributes for forwards-compatibility. Unfortunately, this is sort of a breaking change in that developers who rely on the raw values of these attribute options may end up with attributes they don’t expect.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -34,15 +38,18 @@ open class RouteLeg: NSObject, NSSecureCoding {
             if let speed = jsonAttributes["speed"] {
                 segmentSpeeds = speed as? [CLLocationSpeed] ?? []
             }
-            if let nodes = jsonAttributes["nodes"] {
-                openStreetMapNodeIdentifiers = nodes as? [Int64] ?? []
+            if let congestion = jsonAttributes["congestion"] as? [String] {
+                congestionLevels = congestion.map {
+                    CongestionLevel(description: $0)!
```
frederoni: Is this always safe to unwrap?

1ec5: This will crash if the Directions API returns an unrecognized congestion annotation (like “parking-lot”; “unknown” is fine). Alternatively, we could fall back to `.unknown` or throw an error. It would be inadvisable to omit the value entirely, since the developer expects the array’s length to match the number of segments in the route geometry.

The Directions API reserves the right to introduce new values without notice, which is why we have to fail gracefully when encountering an unknown maneuver type. But if I’m not mistaken, congestion values are tied to the [Mapbox Traffic v1](https://www.mapbox.com/vector-tiles/mapbox-traffic-v1/#traffic) tile set in some fashion. Renaming or introducing new values would probably require a version bump of that tile set at least, right?

/cc @freenerd

1ec5: If we need to accept arbitrary congestion values, let’s handle that as tail work.

freenerd: > congestion values are tied to the Mapbox Traffic v1 tile set in some fashion

In theory yes, but practically they don't share code. Right now they share values, but in the future it is possible that either introduces a value that the other one doesn't copy (e.g. a `closed` tag makes sense for Traffic tiles but in Mapbox Directions we'll never route you through a closed street, thus a `closed` congestion annotation will never occur.

PR #137, opened 2017-06-14
Fix for missing maneuver direction

frederoni:
`maneuverDirection` is missing from a few `RouteStep`s and in those cases, we returned nil instead of decoding the remaining data.

~~We have to set a value when decoding but I don't know if it's correct to assume that a missing maneuverDirection means straightAhead or not.~~

@1ec5 @bsudekum @freenerd 👀

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -557,10 +557,11 @@ open class RouteStep: NSObject, NSSecureCoding {
             return nil
         }
         maneuverType = ManeuverType(description: maneuverTypeDescription)
-        guard let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection") as String? else {
-            return nil
+        if let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection") as String? {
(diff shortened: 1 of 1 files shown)


MapboxDirections/MBRouteStep.swift
```diff
@@ -557,10 +557,11 @@ open class RouteStep: NSObject, NSSecureCoding {
             return nil
         }
         maneuverType = ManeuverType(description: maneuverTypeDescription)
-        guard let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection") as String? else {
-            return nil
+        if let maneuverDirectionDescription = decoder.decodeObject(of: NSString.self, forKey: "maneuverDirection") as String? {
+            maneuverDirection = ManeuverDirection(description: maneuverDirectionDescription)
+        } else {
+            maneuverDirection = ManeuverDirection.straightAhead
```
bsudekum: I wonder if this is worse than not returning one at all. /cc @1ec5

1ec5: We should set `maneuverDirection` to `nil`; likely that’s what the value was before encoding anyways.

frederoni: Fixed

PR #138, opened 2017-06-14
Label first argument of more methods

1ec5:
Renamed some methods to better conform to Swift API design guidelines. The method’s base name shouldn’t contain a preposition. Additionally, the first argument’s label can only be omitted if the base name is a verb or verb phrase.

Also silenced a warning in the Objective-C example target about an unused return value.

/cc @frederoni @bsudekum

### DIFF
--- a/MapboxDirections/MBDirections.swift
+++ b/MapboxDirections/MBDirections.swift
@@ -165,9 +165,9 @@ open class Directions: NSObject {
      */
     @objc(calculateDirectionsWithOptions:completionHandler:)
     open func calculate(_ options: RouteOptions, completionHandler: @escaping CompletionHandler) -> URLSessionDataTask {
-        let url = urlForCalculating(options)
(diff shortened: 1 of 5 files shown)


MapboxDirections/MBDirections.swift
```diff
@@ -165,9 +165,9 @@ open class Directions: NSObject {
      */
     @objc(calculateDirectionsWithOptions:completionHandler:)
     open func calculate(_ options: RouteOptions, completionHandler: @escaping CompletionHandler) -> URLSessionDataTask {
-        let url = urlForCalculating(options)
-        let task = dataTaskWithURL(url, completionHandler: { (json) in
-            let response = options.response(json)
+        let url = self.url(forCalculating: options)
```
1ec5: The `self.` is needed here to avoid a warning about using `url` in its own definition.

PR #140, opened 2017-06-19
Add jazzy documentation

ericrwolfe:
Parallel work to https://github.com/mapbox/mapbox-navigation-ios/pull/285: uses Jazzy to generate a documentation site for Mapbox Directions on iOS. 

Until we can find a home for the docs on Mapbox.com, the docs will live together with the Mapbox Navigation SDK under https://mapbox.github.io/mapbox-navigation-ios/directions


scripts/document.sh
```diff
@@ -0,0 +1,43 @@
...
+    --sdk iphonesimulator \
+    --module-version ${SHORT_VERSION} \
+    --github-file-prefix "https://github.com/mapbox/MapboxDirections.swift/tree/${BRANCH}" \
+    --documentation=docs/guides/*.md \
+    --root-url "https://mapbox.github.io/mapbox-navigation-ios/directions/${RELEASE_VERSION}/" \
+    --theme ${THEME} \
+    --output ${OUTPUT}
+
+find ${OUTPUT} -name *.html -exec \
+    perl -pi -e 's/BRANDLESS_DOCSET_TITLE/Mapbox Directions for iOS $1/, s/MapboxDirections.swift\s+(Docs|Reference)/Mapbox Directions for iOS $1/' {} \;
```
1ec5: The name of this framework is currently MapboxDirections.swift. It isn’t ideal, but it would create more confusion if we start giving the library different names depending whether the user looks. If we aren’t prepared to rename the pod and update all the documentation at the moment, I suggest that we stick to MapboxDirections.swift and ticket out a proper rename.

docs/jazzy.yml
```diff
@@ -0,0 +1,16 @@
+module: MapboxDirections.swift
```
1ec5: The module name is MapboxDirections. It isn’t possible to have a `.` in the module name (one reason I’d like to rename the library).

docs/jazzy.yml
```diff
@@ -0,0 +1,16 @@
...
+author: Mapbox
+author_url: https://www.mapbox.com/directions/
+github_url: https://github.com/mapbox/MapboxDirections.swift
+dash_url: https://mapbox.github.io/mapbox-directions-ios/docsets/Mapbox.xml
```
1ec5: `mapbox-directions-ios`?

docs/README.md
```diff
@@ -0,0 +1,9 @@
+## Generating documentation locally
+
+Run `scripts/document.sh`
+
+## Update the documentation site:
+1. Clone [mapbox-navigation-ios](https://github.com/mapbox/mapbox-navigation-ios) to a mapbox-navigation-ios-docs folder alongside your MapboxDirections.swift clone, and check out the `gh-pages` branch.
```
1ec5: This rigamarole is only necessary if we use the `gh-pages` branch. But we can [point GitHub Pages to the /docs folder](https://github.com/blog/2228-simpler-github-pages-publishing) in the master branch to streamline things further.

ericrwolfe: I'd rather keep the generated documentation out of the main github history, particularly if we move over to a separate mapbox.com subsite.

scripts/document.sh
ericrwolfe: Yes, I'd love to rename the project to something like:
github: `mapbox-directions-ios` 
module: `MapboxDirections`

However this particular PR isn't the time and place for this 😛. Will create a separate ticket.

ericrwolfe:
@1ec5 ready for final review

PR #143, opened 2017-06-22
Avoid traffic by default

1ec5:
In areas without sufficient traffic congestion data, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic` falls back to `MBDirectionsProfileIdentifierAutomobile` anyways, according to [the API documentation](https://www.mapbox.com/api-documentation/#directions).

Developers would naturally expect traffic avoidance to be turned on by default, especially when using this library in conjunction with the navigation SDK. The downsides are that far fewer waypoints and no alternative routes are supported with the traffic-avoiding profile. Therefore, it may make sense to hold this PR until either of these limitations is eased on the server side.

/cc @ericrwolfe @bsudekum @frederoni


ericrwolfe:
Closing as stale. We've also added a subclass of RouteOptions in the Navigation SDK to make these defaults easier to use for navigation https://github.com/mapbox/mapbox-navigation-ios/pull/531.

PR #147, opened 2017-07-06
Expose exit numbers on route steps

### LINKED ISSUE #146: Expose exit numbers on route steps
MapboxDirections.swift should expose the RouteStep `exits` property that was implemented in Project-OSRM/osrm-backend#4215. This property contains the [exit number](https://en.wikipedia.org/wiki/Exit_number) of the exit at the route step’s maneuver.

/cc @bsudekum

1ec5:
Added an `exitCodes` property to RouteStep. It contains the maneuver’s exit number.

This property has this odd name for consistency with `codes` and `destinationCodes`. Even though `exitNumber` might be a more natural name for the property, I wanted to avoid the possibility of a developer thinking it would be a number (e.g., NSNumber) and experiencing some friction as a result.

Fixes #146. Hold until the `exits` property goes live in the Directions API.

/cc @bsudekum @willwhite @frederoni

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -407,11 +407,12 @@ extension String {
 struct Road {
     let names: [String]?
     let codes: [String]?
+    let exitCodes: [String]?
     let destinations: [String]?
     let destinationCodes: [String]?
     let rotaryNames: [String]?
     
-    init(name: String, ref: String?, destination: String?, rotaryName: String?) {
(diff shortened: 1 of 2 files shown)


1ec5:
It occurs to me that, even though the production Directions API doesn’t yet return exit numbers as part of the route response, RouteOptions handles that situation gracefully. So there’s no need to hold this PR until the Directions API upgrades to OSRM v5.9.0.

PR #149, opened 2017-07-09
Split exit codes by semicolon

1ec5:
As seen in [this OSRM response](http://map.project-osrm.org/?z=18&center=39.102957%2C-84.524158&loc=39.103694%2C-84.526727&loc=39.101800%2C-84.523154&hl=en&alt=0), exit numbers are delimited by semicolons ([as in OpenStreetMap](http://www.openstreetmap.org/way/19051476)) rather than commas (as with route numbers and destinations).

/cc @freenerd

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -447,7 +447,7 @@ struct Road {
             self.destinations = destination?.tagValues(separatedBy: ",")
         }
         
-        self.exitCodes = exits?.tagValues(separatedBy: ",")
+        self.exitCodes = exits?.tagValues(separatedBy: ";")
         self.codes = codes
         self.rotaryNames = rotaryName?.tagValues(separatedBy: ";")
     }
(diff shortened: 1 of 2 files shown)


freenerd:
I think this is actually a bug and should be consistently commas. Moved upstream since we can still fix before next OSRM release https://github.com/Project-OSRM/osrm-backend/issues/4270

1ec5:
Per https://github.com/Project-OSRM/osrm-backend/issues/4270#issuecomment-314109170, OSRM is going to stick to semicolons until we’re able to put this data in arrays in a future release. This ticket would be the Swift analogue to Project-OSRM/osrm-text-instructions#126.

freenerd:
For reference since OP has a link to map.project-osrm.org.

before https://github.com/Project-OSRM/osrm-text-instructions/pull/126

2017-07-11 at 15 35

after https://github.com/Project-OSRM/osrm-text-instructions/pull/126

2017-07-11 at 15 39

freenerd:
This is good to go from my side :+1:

PR #152, opened 2017-07-13
Customize variables Quick Look for various types

### LINKED ISSUE #150: Customize Quick Look for debugging routes
Route, RouteLeg, and RouteStep should each implement [Quick Look](https://developer.apple.com/library/content/documentation/IDEs/Conceptual/CustomClassDisplay_in_QuickLook/Introduction/Introduction.html) so a developer can easily inspect these objects in the debugger, similar to how you can Quick Look preview a CLLocationCoordinate2D on a MapKit map view. The Quick Look preview would consist of a map displaying the route and, in the case of a route step, an arrow depicting the maneuver. To minimize dependencies, these implementations could form Static API requests directly instead of using MapboxStatic.swift.

/cc @frederoni

1ec5:
Customized Quick Look for Route, RouteLeg, and RouteStep to display a static map with the geometry overlaid, powered by the Mapbox Static API. Customized Quick Look for Waypoint to display the coordinate using Xcode’s built-in coordinate viewer.

The static map uses the Mapbox Streets style for the `.automobile` profile, Traffic Day for `.automobileAvoidingTraffic`, and Outdoors for `.cycling` and `.walking`.

automobile
automobileavoidingtraffic

Fixes #150.

/ref mapbox/mapbox-gl-native#9484
/cc @frederoni @bsudekum

### DIFF
--- a/MapboxDirections/MBRoute.swift
+++ b/MapboxDirections/MBRoute.swift
@@ -161,6 +161,13 @@ open class Route: NSObject, NSSecureCoding {
      The route options object’s profileIdentifier property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.
      */
     open let routeOptions: RouteOptions
+    
+    func debugQuickLookObject() -> Any? {
+        if let coordinates = coordinates {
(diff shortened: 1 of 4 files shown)


1ec5:
Not sure why the webhook is refusing to update, but both iOS builds did pass on Bitrise. Merging.

PR #154, opened 2017-07-18
Expose road classes in route steps

bsudekum:
The directions API will soon return a class type: `toll`, `restricted`,`highway` and `ferry`.

This array is found on the intersections key and represents the class type in between each intersection.

The test fixtures are so large right now that working through this task is becoming cumbersome. Going to hold off on this until the API returns the actual keys I'm looking for.

/cc @1ec5 @frederoni


MapboxDirections/MBRoadClass.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+
+@objc(MBRoadClass)
+public enum RoadClass: Int, CustomStringConvertible {
+    case toll
```
1ec5: If I understand correctly, the specific `classes` values coming out of the Directions API are dependent on the profile, correct? If there’s a possibility that a different profile might produce exotic values, we should turn this enum into an extensible string enumeration. (`MBDirectionsProfileIdentifier` is an example of an extensible string enumeration.) Otherwise, we’d have to document the fact that unrecognized classes could be dropped entirely.

MapboxDirections/MBRoadClass.swift
```diff
@@ -0,0 +1,42 @@
...
+    case ferry
+    
+    public init?(description: String) {
+        let level: RoadClass
+        switch description {
+        case "toll":
+            level = .toll
+        case "restricted":
+            level = .restricted
+        case "highway":
```
1ec5: Apparently the Directions API will use a `motorway` class, not a `highway` class.

bsudekum: @1ec5 I've been told by @TheMarex it will be `toll`, `restricted`,`highway` and `ferry`. 

@TheMarex is this the case?

MapboxDirections/MBIntersection.swift
```diff
@@ -50,6 +50,8 @@ public class Intersection: NSObject, NSSecureCoding {
      */
     public let usableApproachLanes: IndexSet?
     
+    public let roadClasses: [RoadClass]
```
1ec5: An array of enumeration values won’t bridge very cleanly to Objective-C. Additionally, the order of classes doesn’t matter. So we should turn the RoadClass enumeration into a RoadClasses option set (implemented using `NS_OPTION` in Objective-C), then change this property’s type to simply RoadClasses.

MapboxDirections/MBRoadClass.swift
TheMarex: Urg sorry this was my fault. `motorway` is correct.

bsudekum:
@1ec5 this is ready for review (still need to document).

MapboxDirections/MBRoadClasses.h
```diff
@@ -0,0 +1,10 @@
+typedef NS_OPTIONS(NSUInteger, MBRoadClasses) {
+    
+    MBRoadClassesToll = (1 << 1),
+    
+    MBRoadClassesRestricted = (1 << 2),
+    
+    MBRoadClassesMotorway = (1 << 3),
+    
+    MBRoadClassesFerry = (1 << 4),
```
1ec5: The documentation comment for this value should discuss its relationship to the [`ferry` transport type](https://github.com/mapbox/MapboxDirections.swift/blob/93dabdebb1375e745b4a09912afa2d1d4c669cc1/MapboxDirections/MBRouteStep.swift#L17-L22), and the documentation for that transport type should likewise refer to this road class.

MapboxDirections/MBRoadClasses.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+
+public typealias RoadClasses = MBRoadClasses
+
+extension RoadClasses: CustomStringConvertible {
+    
+    public init?(descriptions: [String]) {
```
1ec5: This initializer needs a documentation comment as well.

MapboxDirections/MBRoadClasses.swift
```diff
@@ -0,0 +1,48 @@
...
+            default:
+                return nil
+            }
+        }
+        self.init(rawValue: roadClasses.rawValue)
+    }
+    
+    public var description: String {
+        if isEmpty {
+            return "none"
```
1ec5: It should be possible to round-trip a road class from an array of strings to a RoadClasses and back to a string. Is `none` a valid string that would appear in a route response? If not, maybe an empty string would be better. Unfortunately, our hands our kind of tied because CustomStringConvertible doesn’t allow for an optional description.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -69,6 +69,8 @@ class V5Tests: XCTestCase {
         let leg = route!.legs.first!
         XCTAssertEqual(leg.name, "I 80, I 80;US 30")
         XCTAssertEqual(leg.steps.count, 59)
+        let roadClasses = leg.steps.first!.intersections!.first!.roadClasses!
```
1ec5: Before `!` force-unwrapping `steps.first` and `intersections` and `intersections.first` and `roadClasses`, we should assert that each of these things are non-nil or non-empty. Otherwise we’ll have an unhelpful assertion message when the following assertion fails.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -69,6 +69,8 @@ class V5Tests: XCTestCase {
...
+        XCTAssertTrue(roadClasses.contains([.toll, .restricted]))
```
1ec5: To prevent the test harness from crashing when any of these things is nil, use optionals:

```swift
let firstStep = leg.steps.first // as opposed to step below
XCTAssertNotNil(firstStep)
let intersections = firstStep?.intersections
XCTAssertNotNil(intersections)
let intersection = intersections?.first
XCTAssertNotNil(intersection)
let roadClasses = intersection?.roadClasses
XCTAssertNotNil(roadClasses)
XCTAssertTrue(roadClasses?.contains([.toll, .restricted]) ?? false)
```

PR #155, opened 2017-07-19
Add accessToken used to make request to Route

bsudekum:
It's helpful to know what access token was used to make the directions request. This adds `acccessToken` to the `Route` class.

/cc @1ec5

### DIFF
--- a/MapboxDirections/MBDirections.swift
+++ b/MapboxDirections/MBDirections.swift
@@ -171,7 +171,7 @@ open class Directions: NSObject {
             if let routes = response.1 {
                 for route in routes {
                     route.accessToken = self.accessToken
-                    route.host = self.apiEndpoint
+                    route.apiEndpoint = self.apiEndpoint
                 }
             }
             completionHandler(response.0, response.1, nil)
--- a/MapboxDirections/MBDirections.swift
+++ b/MapboxDirections/MBDirections.swift
@@ -168,6 +168,11 @@ open class Directions: NSObject {
         let url = self.url(forCalculating: options)
         let task = dataTask(with: url, completionHandler: { (json) in
             let response = options.response(from: json)
+            if let routes = response.1 {
+                for route in routes {
+                    route.accessToken = self.accessToken
+                }
+            }
             completionHandler(response.0, response.1, nil)
         }) { (error) in
             completionHandler(nil, nil, error)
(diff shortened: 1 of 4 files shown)


MapboxDirections/MBDirections.swift
```diff
@@ -168,6 +168,11 @@ open class Directions: NSObject {
         let url = self.url(forCalculating: options)
         let task = dataTask(with: url, completionHandler: { (json) in
             let response = options.response(from: json)
+            if let routes = response.1 {
+                for route in routes {
+                    route.accessToken = self.accessToken
```
1ec5: We need to pair this with `apiEndpoint`, since an access token only makes sense for a particular endpoint (think api.mapbox.com versus api.mapbox.cn).

MapboxDirections/MBRoute.swift
```diff
@@ -162,6 +162,20 @@ open class Route: NSObject, NSSecureCoding {
      */
     open let routeOptions: RouteOptions
     
+    /**
+     The access token used to make the directions request.
+     
+     Only set if a request is made via `Directions.calculate(_:completionHandler:)`.
```
1ec5: The current wording can be misread as, “You should only set this property if…”. How about:

> This property is set automatically if a request is made via `Directions.calculate(_:completionHandler:)`.

Same for `apiEndpoint` below.

MapboxDirections/MBRoute.swift
```diff
@@ -162,6 +162,20 @@ open class Route: NSObject, NSSecureCoding {
      */
     open let routeOptions: RouteOptions
     
+    /**
+     The access token used to make the directions request.
```
1ec5: Link to https://www.mapbox.com/help/define-access-token/ using Markdown syntax.

MapboxDirections/MBRoute.swift
```diff
@@ -162,6 +162,20 @@ open class Route: NSObject, NSSecureCoding {
      */
     open let routeOptions: RouteOptions
     
+    /**
+     The [access token](https://www.mapbox.com/help/define-access-token/) used to make the directions request.
+     
+     This property is set automatically if a request is made via Directions.calculate(_:completionHandler:).
```
1ec5: 0b9d18cc8e906ac81eaf7f3d784240e9673067e1 fixes a syntax error in this line.

PR #159, opened 2017-08-01
[Hold] Update usages of map() to compile in Xcode 9

boundsj:
Addresses errors like:

* 'map' produces '[T]', not the expected contextual result type 'Waypoint'
* Closure tuple parameter '(offset: Int, element: Bool)' does not support destructuring with implicit parameters

### DIFF
--- a/MapboxDirections.xcodeproj/project.pbxproj
+++ b/MapboxDirections.xcodeproj/project.pbxproj
@@ -752,11 +752,11 @@
 					};
 					DA6C9D871CAE442B00094FBC = {
 						CreatedOnToolsVersion = 7.3;
-						LastSwiftMigration = 0800;
+						LastSwiftMigration = 0900;
 					};
 					DA6C9D901CAE442B00094FBC = {
 						CreatedOnToolsVersion = 7.3;
-						LastSwiftMigration = 0800;
+						LastSwiftMigration = 0900;
 					};
 					DADD27B41E5AAAD800D31FAD = {
 						CreatedOnToolsVersion = 8.2.1;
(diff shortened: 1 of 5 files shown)


MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -1360,7 +1360,8 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.mapbox.MapboxDirections;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 3.0;
+				SWIFT_SWIFT3_OBJC_INFERENCE = Off;
+				SWIFT_VERSION = 4.0;
```
frederoni: Should we wait with this migration until the release this fall? And even then, Swift 3.2 should be sufficient. Also, I think CI bails because of this. Is the code changes compatible with Swift 3 and 4?

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -1360,7 +1360,8 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.mapbox.MapboxDirections;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 3.0;
+				SWIFT_SWIFT3_OBJC_INFERENCE = Off;
```
frederoni: Cool, I wonder if this would've fixed issues like https://github.com/mapbox/MapboxStatic.swift/pull/66

boundsj:
I forgot to mention that this is intended for Xcode 9 and Swift 4 later this year. No need to merge now.

boundsj:
Noting that https://bugs.swift.org/browse/SR-4738 provides some insight into the tuple destructuring issue. You can also check https://bugs.swift.org/browse/SR-1620 for more background. From SR-4738

> Mark Lacey added a comment - 1 May 2017 7:24 PM
> As mentioned in SR-2135, this isn't really intended to work because we don't support general destructuring in parameter lists. There was special-case behavior that only worked for a single tuple argument but that is going away in -swift-version 4.
> This on the other hand is a close alternative that explicitly uses destructuring on the single tuple argument:
> [t].map { (tuple) -> Int in
>   let (x, (y, z)) = tuple
>   return x + y + z
> }

cc @bsudekum

boundsj:
stale

PR #165, opened 2017-09-01
Add routeIdentifier

bsudekum:
Now that the directions API returns a unique identifier for each request, we should make this available to the developer.

/cc @1ec5 @ericrwolfe


bsudekum:
Ugh carthage errors.

MapboxDirections/MBDirections.swift
```diff
@@ -121,6 +121,9 @@ open class Directions: NSObject {
     /// The Mapbox access token to associate the request with.
     internal let accessToken: String
     
+    /// The routeIdentifier for the request.
+    internal var routeIdentifier: String?
```
1ec5: A single Directions object can produce multiple requests that may come back in a different order. Storing this information on Directions makes the class vulnerable to race conditions.

MapboxDirections/MBDirections.swift
```diff
@@ -172,6 +175,7 @@ open class Directions: NSObject {
                 for route in routes {
                     route.accessToken = self.accessToken
                     route.apiEndpoint = self.apiEndpoint
+                    route.routeIdentifier = self.routeIdentifier
```
1ec5: Get the route identifier from `json["uuid"]` instead of `Directions.routeIdentifier`.

MapboxDirections/MBRoute.swift
```diff
@@ -182,6 +182,13 @@ open class Route: NSObject, NSSecureCoding {
         }
         return nil
     }
+    
+    /**
+     A unique identifier for a directions request.
+     
+     Note, each `Route` in a request will have the same `routeIdentifier`.
```
1ec5: > Each route produced by a single call to `calculate(_:completionHandler:)` has the same route identifier.

1ec5: Can we say anything about how the developer uses this identifier? Or is it only for the navigation SDK’s use?

bsudekum:
@1ec5 updated.

@frederoni any ideas about the carthage test failures here?

MapboxDirections/MBRoute.swift
```diff
@@ -186,7 +186,7 @@ open class Route: NSObject, NSSecureCoding {
     /**
      A unique identifier for a directions request.
      
-     Note, each `Route` in a request will have the same `routeIdentifier`.
+     Each route produced by a single call to `calculate(_:completionHandler:)` has the same route identifier.
```
1ec5: Oops, I meant to say `Directions.calculate(_:completionHandler:)`. Otherwise, jazzy won’t be able to find the symbol to link to it.

MapboxDirections/MBRoute.swift
```diff
@@ -182,6 +182,13 @@ open class Route: NSObject, NSSecureCoding {
         }
         return nil
     }
+    
+    /**
+     A unique identifier for a directions request.
+     
+     Each route produced by a single call to `calculate(_:completionHandler:)` has the same route identifier.
+     */
+    open var routeIdentifier: String?
```
1ec5: If these UUIDs are persistent across application sessions, we should encode and decode `routeIdentifier` like the other properties.

1ec5:
> any ideas about the carthage test failures here?

The workflows had some really outdated steps. I upgraded them and also removed the “Upgrade Carthage” step that’s no longer necessary, now that the official Carthage step is using a recent version of Carthage.

bsudekum:
@1ec5 updated

MapboxDirections/MBRoute.swift
```diff
@@ -68,6 +68,11 @@ open class Route: NSObject, NSSecureCoding {
             return nil
         }
         routeOptions = options
+        
+        guard let decodedRouteIdentifier = decoder.decodeObject(of: NSString.self, forKey: "routeIdentifier") as String? else {
+            return nil
+        }
+        routeIdentifier = decodedRouteIdentifier
```
1ec5: Nit: Call the local variable `routeIdentifier` and change this line to:

```swift
self.routeIdentifier = routeIdentifier
```

MapboxDirections/MBRoute.swift
```diff
@@ -68,6 +68,11 @@ open class Route: NSObject, NSSecureCoding {
             return nil
         }
         routeOptions = options
+        
+        guard let decodedRouteIdentifier = decoder.decodeObject(of: NSString.self, forKey: "routeIdentifier") as String? else {
+            return nil
```
1ec5: Oh, also, since `routeIdentifier` is optional, we should probably just swallow the lack of an identifier. No need to guard in this case.

PR #174, opened 2017-09-14
Serialize pronunciation fields

### LINKED ISSUE #78: Serialize pronunciation field
The `Road` class or struct we’re planning to add for mapbox/MapboxDirections.swift#67 needs a `pronunciation` property, of type `String?`, that corresponds to the field of the same name in the Directions API response.

/cc @bsudekum

1ec5:
Added `phoneticNames` and `phoneticExitNames` properties to RouteStep.

Fixes #78.

/cc @bsudekum @frederoni


frederoni:
Do we have phonetic destinations? It would solidify https://github.com/mapbox/mapbox-navigation-ios/pull/603

1ec5:
> Do we have phonetic destinations? It would solidify https://github.com/mapbox/mapbox-navigation-ios/pull/603

Unfortunately, OSRM doesn’t support phonetic destinations yet: Project-OSRM/osrm-backend#4505. #603 should be fine regardless, since the Spanish orthography is especially regular.

Issue #133, opened 2017-06-01
Server-side localized instructions


1ec5:
The Directions API has a new `language` query parameter (documentation forthcoming) that localizes the `instruction` property using OSRM Text Instructions. Developers who aren’t using real-time, turn-by-turn guidance instructions would be able to use this `instruction` property directly without having to link against OSRM Text Instructions for Swift.

The `language` query parameter doesn’t currently fall back to a similar language when an unsupported language code is provided. For now, we should omit the parameter by default but consider defaulting it to `Locale.current` in the future.

mapbox/mapbox-java#474 is the equivalent issue for Java.

/cc @freenerd @frederoni

1ec5:
Here’s [the corresponding `locale` option](https://github.com/mapbox/MapboxGeocoder.swift/blob/26f5fbab24ba5563248bc17700b8752be5401119/MapboxGeocoder/MBGeocodeOptions.swift#L56-L65) in MapboxGeocoder.swift.

PR #175, opened 2017-09-14
Add voice instructions

bsudekum:
Adds support for handling the `voice_instructions` query parameter and the response. 

/cc @mapbox/guidance @1ec5 @frederoni @ericrwolfe


MapboxDirections/MBRouteOptions.swift
```diff
@@ -303,6 +339,10 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var includesExitRoundaboutManeuver = false
     
+    open var includeVoiceInstructions = false
+    
+    open var unitsForVoiceInstructions: VoiceUnit = .imperial
```
1ec5: Since this property takes only one value, the plural “units” is misleading. Assuming the type is renamed to `MeasurementSystem`, this property should be called `lengthMeasurementSystem`.

MapboxDirections/MBRouteStep.swift
```diff
@@ -623,6 +626,8 @@ open class RouteStep: NSObject, NSSecureCoding {
         destinations = decoder.decodeObject(of: [NSArray.self, NSString.self], forKey: "destinations") as? [String]
         
         intersections = decoder.decodeObject(of: [NSArray.self, Intersection.self], forKey: "intersections") as? [Intersection]
+        
+        voiceInstructions = decoder.decodeObject(of: [NSArray.self, Intersection.self], forKey: "voiceInstructions") as? [VoiceInstruction]
```
1ec5: This will fail because the array contains `VoiceInstruction`s, not `Intersection`s.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+@objc(MBVoiceInstruction)
+public class VoiceInstruction: NSObject, NSSecureCoding {
+    
+    public let distanceAlongGeometry: CLLocationDirection
```
1ec5: This library has generally avoided the term “geometry” as a countable noun. Anyhow, I think it would be less ambiguous to call it `distanceAlongStep`, since routes also have geometry data attached to them.

1ec5: `CLLocationDistance` would be a more suitable type for this property than `CLLocationDirection`.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    public let textAnnouncement: String
+    
+    public let SSMLAnnouncement: String
```
1ec5: Instruction and announcement are sort of synonymous, so having them both in this class is a bit redundant. Let’s call these methods `text` and `ssmlText`. `ssmlText` should have an Objective-C name of `SSMLText`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -100,6 +100,39 @@ public enum RouteShapeResolution: UInt, CustomStringConvertible {
     }
 }
 
+@objc(MBVoiceUnit)
+public enum VoiceUnit: UInt, CustomStringConvertible {
```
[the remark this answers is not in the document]
1ec5: Alternatively, if you want to reuse an existing type, `NSLengthFormatterUnit` would be one option, though it’d be a bit awkward that specifying `.kilometer` could sometimes produce distances in meters.

MapboxDirections/MBRouteStep.swift
```diff
@@ -668,6 +673,7 @@ open class RouteStep: NSObject, NSSecureCoding {
         coder.encode(codes, forKey: "codes")
         coder.encode(destinationCodes, forKey: "destinationCodes")
         coder.encode(destinations, forKey: "destinations")
+        coder.encode(voiceInstructions, forKey: "voiceInstructions")
```
1ec5: Remember to add something about `voiceInstructions` to `RouteStepTests.testCoding()`.

MapboxDirections/MBRouteStep.swift
```diff
@@ -876,6 +882,8 @@ open class RouteStep: NSObject, NSSecureCoding {
         }
         return nil
     }
+    
+    open let voiceInstructions: [VoiceInstruction]?
```
1ec5: Move this property closer to the `instructions` property. Also call it `instructionsSpokenAlongStep` or `instructionsSpokenInAdvance` to clarify that these instructions pertain to the following step, unlike the `instructions` property.

MapboxDirections/MBRouteStep.swift
```diff
@@ -623,6 +626,8 @@ open class RouteStep: NSObject, NSSecureCoding {
         destinations = decoder.decodeObject(of: [NSArray.self, NSString.self], forKey: "destinations") as? [String]
         
         intersections = decoder.decodeObject(of: [NSArray.self, Intersection.self], forKey: "intersections") as? [Intersection]
+        
+        instructionsSpokenAlongStep = decoder.decodeObject(of: [NSArray.self, VoiceInstruction.self], forKey: "voiceInstructions") as? [VoiceInstruction]
```
bsudekum: @1ec5 how do you feel about moving away from the word `spoken` in favor of `voice`? I find myself naming variables voice a lot in https://github.com/mapbox/mapbox-navigation-ios/pull/614/files

1ec5: “Voice” is a noun, not an adjective. In both AVFoundation and Polly, a “voice” refers is a style of speech output that you can choose, e.g. “Joanna” versus “Giorgio”; it doesn’t refer to the act of speaking. Along these lines, “SpokenInstructions” would be preferable to “VoiceInstructions”, and if we name a method that does the speaking, it should be called “speak”, not “voice”.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -100,18 +100,24 @@ public enum RouteShapeResolution: UInt, CustomStringConvertible {
...
-@objc(MBVoiceUnit)
-public enum VoiceUnit: UInt, CustomStringConvertible {
+/**
+ A `MeasurementSystem` indicates the type of units used for formatted a voice instruction.
+ */
+@objc(MBMeasurementSystem)
+public enum MeasurementSystem: UInt, CustomStringConvertible {
     
+    /**
+     Imperial units.
```
1ec5: > U.S. customary and British imperial units.
> 
> Distances are measured in miles and feet.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -100,18 +100,24 @@ public enum RouteShapeResolution: UInt, CustomStringConvertible {
     }
 }
 
-@objc(MBVoiceUnit)
-public enum VoiceUnit: UInt, CustomStringConvertible {
+/**
+ A `MeasurementSystem` indicates the type of units used for formatted a voice instruction.
```
1ec5: Some of this detail is better left to the `RouteOptions.distanceMeasurementSystem` documentation. How about:

> A system of units of measuring distances and other quantities.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -100,18 +100,24 @@ public enum RouteShapeResolution: UInt, CustomStringConvertible {
...
+@objc(MBMeasurementSystem)
+public enum MeasurementSystem: UInt, CustomStringConvertible {
     
+    /**
+     Imperial units.
+     */
     case imperial
     
+    /**
+     Metric system.
```
1ec5: > The metric system.
> 
> Distances are measured in kilometers and meters.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -100,18 +100,24 @@ public enum RouteShapeResolution: UInt, CustomStringConvertible {
...
+     */
     case metric
     
-    // Add once available
-    // case imperialGB
-    
     public init?(description: String) {
-        let format: VoiceUnit
+        let format: MeasurementSystem
```
1ec5: Nit: Rename `format` to `measurementSystem` for consistency.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -339,9 +345,19 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
+     */
+    open var includesVoiceInstructions = false
     
-    open var unitsForVoiceInstructions: VoiceUnit = .imperial
+    /**
+     The distance units used for the `includesVoiceInstructions`.
+     
+     `includesVoiceInstructions` must be true for this property to have any effect.
+     */
+    open var distanceMeasurementSystem: MeasurementSystem = .imperial
```
1ec5: The default should not be `imperial`. If we need to choose a default, it should be `metric`, the international standard. But I think it would be better to default to a dynamic value based on the `locale`, which would require #133 to be implemented first.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -339,9 +345,19 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var includesExitRoundaboutManeuver = false
     
-    open var includeVoiceInstructions = false
+    /**
+     A Boolean value indicating whether the route includes an array of `VoiceInstructions`.
```
1ec5: The Route doesn’t include an array of `VoiceInstructions`:

> A Boolean value indicating whether each route step includes an array of `SpokenInstruction`s.
> 
> If this option is set to `true`, the `RouteStep.instructionsSpokenAlongStep` property is set to an array of `SpokenInstruction`s.

MapboxDirections/MBRouteStep.swift
```diff
@@ -883,7 +883,7 @@ open class RouteStep: NSObject, NSSecureCoding {
         return nil
     }
     
-    open let voiceInstructions: [VoiceInstruction]?
+    open let instructionsSpokenAlongStep: [VoiceInstruction]?
```
1ec5: > Instructions about the next step’s maneuver, optimized for speech synthesis.
> 
> As the user traverses this step, you can give them advance notice of the upcoming maneuver by reading aloud each item in this array in order as the user reaches the specified distances along this step. The text of the spoken instructions refers to the details in the next step, but the distances are measured from the beginning of this step.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -339,9 +345,19 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
+     
+     The order of each `VoiceInstructions` in the array indicates the order of which each instruction should be given in.
+     */
+    open var includesVoiceInstructions = false
     
-    open var unitsForVoiceInstructions: VoiceUnit = .imperial
+    /**
+     The distance units used for the `includesVoiceInstructions`.
```
1ec5: > The measurement system used in spoken instructions included in route steps.
> 
> If the `includesSpokenInstructions` property is set to `true`, this property determines the units used for measuring the distance remaining until an upcoming maneuver. If the `includesSpokenInstructions` property is set to `false`, this property has no effect.
> 
> You should choose a measurement system appropriate for the current region. You can also allow the user to indicate their preferred measurement system via a setting.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -3,30 +3,43 @@ import Foundation
 @objc(MBVoiceInstruction)
 public class VoiceInstruction: NSObject, NSSecureCoding {
     
-    public let distanceAlongGeometry: CLLocationDirection
+    /**
+     The distance in meters along the current `RouteStep` to give the voice announcement.
```
1ec5: “Along the current step” is ambiguous. How about:

> A distance along the associated `RouteStep` at which to read the instruction aloud.
> 
> The distance is measured in meters from the beginning of the associated step.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -3,30 +3,43 @@ import Foundation
 @objc(MBVoiceInstruction)
 public class VoiceInstruction: NSObject, NSSecureCoding {
```
1ec5: > An instruction about an upcoming `RouteStep`’s maneuver, optimized for speech synthesis.
> 
> The instruction is provided in two formats: plain text and text marked up according to the [Speech Synthesis Markup Language](https://en.wikipedia.org/wiki/Speech_Synthesis_Markup_Language) (SSML). Use a speech synthesizer such as `AVSpeechSynthesizer` or Amazon Polly to read aloud the instruction.
> 
> The `distanceAlongStep` property is measured from the beginning of the step associated with this object. By contrast, the `text` and `ssmlText` properties refer to the details in the following step. It is also possible for the instruction to refer to two following steps simultaneously when needed for safe navigation.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -3,30 +3,43 @@ import Foundation
...
+    /**
+     The distance in meters along the current `RouteStep` to give the voice announcement.
+     */
+    public let distanceAlongStep: CLLocationDistance
 
-    public let textAnnouncement: String
     
-    public let SSMLAnnouncement: String
+    /**
+     A plain text representation of the voice instruction to give to the user.
```
1ec5: > A plain-text representation of the speech-optimized instruction.
> 
> This representation is appropriate for speech synthesizers that lack support for the [Speech Synthesis Markup Language](https://en.wikipedia.org/wiki/Speech_Synthesis_Markup_Language) (SSML), such as `AVSpeechSynthesizer`. For speech synthesizers that support SSML, use the `ssmlText` property instead.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -3,30 +3,43 @@ import Foundation
...
+     */
+    public let text: String
+    
+    
+    /**
+     The voice instruction formatted with [SSML](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/speech-synthesis-markup-language-ssml-reference) (Speech Synthesis Markup Language).
+     
+     Note, when speaking an instruction with this property, the speed synthesis engine must support SSML.
+     */
+    public let ssml: String
```
1ec5: Name this property `ssmlText` to make its relationship to `text` more clear. Also give it an `@objc(SSMLText)` attribute.

MapboxDirections/MBVoiceInstruction.swift
```diff
@@ -3,30 +3,43 @@ import Foundation
...
+    /**
+     A plain text representation of the voice instruction to give to the user.
+     */
+    public let text: String
+    
+    
+    /**
+     The voice instruction formatted with [SSML](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/speech-synthesis-markup-language-ssml-reference) (Speech Synthesis Markup Language).
+     
+     Note, when speaking an instruction with this property, the speed synthesis engine must support SSML.
```
1ec5: This library should be synthesizer-agnostic, so let’s link to [Wikipedia](https://en.wikipedia.org/wiki/Speech_Synthesis_Markup_Language) instead of the Alexa documentation.

> A formatted representation of the speech-optimized instruction.
> 
> This representation is appropriate for speech synthesizers that support the [Speech Synthesis Markup Language](https://en.wikipedia.org/wiki/Speech_Synthesis_Markup_Language) (SSML), such as [Amazon Polly](https://aws.amazon.com/polly/). Numbers and names are marked up to ensure correct pronunciation. For speech synthesizers that lack SSML support, use the `text` property instead.

bsudekum:
@1ec5 this is ready

bsudekum:
This will not be merged until this API is publicly documented.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -101,6 +101,49 @@ public enum RouteShapeResolution: UInt, CustomStringConvertible {
 }
 
 /**
+ A `MeasurementSystem` indicates the type of units used for formatted a voice instruction.
```
1ec5: https://github.com/mapbox/MapboxDirections.swift/pull/175#discussion_r139205854

MapboxDirections/MBRouteOptions.swift
```diff
@@ -304,6 +350,22 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
+    open var includesVoiceInstructions = false
+    
+    /**
+     The measurement system used in spoken instructions included in route steps.
+     
+     If the `includesSpokenInstructions` property is set to `true`, this property determines the units used for measuring the distance remaining until an upcoming maneuver. If the `includesSpokenInstructions` property is set to `false`, this property has no effect.
+     
+     You should choose a measurement system appropriate for the current region. You can also allow the user to indicate their preferred measurement system via a setting.
+     */
+    open var distanceMeasurementSystem: MeasurementSystem = .metric
```
1ec5: Flagging here that I think https://github.com/mapbox/MapboxDirections.swift/pull/175#discussion_r139206621 still applies, that we should do something based on #133 instead of making the developer figure out the right measurement system.

MapboxDirections/SpokenInstruction.swift
```diff
@@ -0,0 +1,56 @@
...
+    internal init(json: JSONDictionary) {
+        distanceAlongStep = json["distanceAlongGeometry"] as! CLLocationDistance
+        text = json["announcement"] as! String
+        ssmlText = json["ssmlAnnouncement"] as! String
+    }
+    
+    public required init?(coder decoder: NSCoder) {
+        distanceAlongStep = decoder.decodeDouble(forKey: "distanceAlongStep")
+        text = decoder.decodeObject(of: [NSArray.self, NSString.self], forKey: "text") as! String
+        ssmlText = decoder.decodeObject(of: [NSArray.self, NSString.self], forKey: "ssmlText") as! String
```
1ec5: These two keys hold strings, not arrays of strings.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -201,126 +246,151 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
+     The measurement system used in spoken instructions included in route steps.
+
+     If the `includesSpokenInstructions` property is set to `true`, this property determines the units used for measuring the distance remaining until an upcoming maneuver. If the `includesSpokenInstructions` property is set to `false`, this property has no effect.
+
+     You should choose a measurement system appropriate for the current region. You can also allow the user to indicate their preferred measurement system via a setting.
+     */
+    open var distanceMeasurementSystem: MeasurementSystem = usesMetric ? .metric : .imperial
+    
+    internal static var usesMetric: Bool {
+        let locale = Locale.current as NSLocale
```
1ec5: If the `RouteOptions.locale` property is set, we should probably use that locale instead of the current one for the default value.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -356,15 +356,23 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
      By default, this property is set to `nil`, causing instructions to be written in the default language, English.
      */
-    open var locale: Locale?
+    open var locale: Locale? {
+        didSet {
+            guard let locale = locale else {
+                self.distanceMeasurementSystem = Locale.current.usesMetric ? .metric : .imperial
+                return
+            }
+            self.distanceMeasurementSystem = locale.usesMetric ? .metric : .imperial
```
1ec5: This is pretty cool, although are we sure it works in Objective-C? Maybe add some tests to that effect.

A simpler alternative would be to eschew these `didSet`s and move this logic to the `params` getter.

bsudekum: Tested in the example: 

```
options.includesVoiceInstructions = YES;
options.locale = [NSLocale currentLocale];
```

And put a breakpoint on the next line to see what the value:

```
(lldb) po options.distanceMeasurementSystem
<nil>
```

frederoni: Hm, exactly the same changes in the Obj-C example:
```
(lldb) po options.distanceMeasurementSystem
MBMeasurementSystemImperial
```

bsudekum: Cleaned my workspace and tried again, worked.

PR #176, opened 2017-09-15
Add language/locale

bsudekum:
Closes: https://github.com/mapbox/MapboxDirections.swift/issues/133

/cc @1ec5

### DIFF
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -316,8 +316,6 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var locale: Locale?
     
-    private let supportedLocales: [String] = ["de", "en", "es", "fr", "id", "nl", "ru", "sv", "vi", "zh"]
-    
     /**
      An array of URL parameters to include in the request URL.
      */
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -182,7 +182,10 @@ open class RouteOptions: NSObject, NSSecureCoding {
         
         includesExitRoundaboutManeuver = decoder.decodeBool(forKey: "includesExitRoundaboutManeuver")
         
-        locale = Locale(identifier: decoder.decodeObject(of: NSString.self, forKey: "locale") as String? ?? "")
+        guard let locale = decoder.decodeObject(of: NSLocale.self, forKey: "locale") as Locale? else {
+            return nil
+        }
+        self.locale = locale
     }
     
     open static var supportsSecureCoding = true
@@ -197,9 +200,7 @@ open class RouteOptions: NSObject, NSSecureCoding {
         coder.encode(routeShapeResolution.description, forKey: "routeShapeResolution")
         coder.encode(attributeOptions.description, forKey: "attributeOptions")
         coder.encode(includesExitRoundaboutManeuver, forKey: "includesExitRoundaboutManeuver")
-        if let locale = locale, let languageCode = (locale as NSLocale?)?.object(forKey: .languageCode) as? String {
-            coder.encode(languageCode, forKey: "locale")
-        }
+        coder.encode(locale, forKey: "locale")
     }
     
     // MARK: Specifying the Path of the Route
@@ -315,6 +316,8 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var locale: Locale?
     
+    private let supportedLocales: [String] = ["de", "en", "es", "fr", "id", "nl", "ru", "sv", "vi", "zh-hans"]
+    
     /**
      An array of URL parameters to include in the request URL.
      */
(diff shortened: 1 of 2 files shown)


MapboxDirections/MBRouteOptions.swift
```diff
@@ -319,6 +331,10 @@ open class RouteOptions: NSObject, NSSecureCoding {
             params.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
         }
         
+        if let locale = locale, let languageCode = (locale as NSLocale?)?.object(forKey: .languageCode) as? String {
```
1ec5: The language code is only the two-letter ISO 369 code. It doesn’t ever include a region or script code; for that, you need the full locale identifier. Unfortunately, the Directions API currently lacks any sort of fallback mechanism. So this code will fail for Simplified Chinese (`zh-Hant`), yet if we were to pass in the full locale identifier, the code would break for a Spanish speaker in Mexico (`es-MX`) instead.

Until the Directions API gains basic language fallback capability, we’ll need to hard-code a whitelist of locales that the Directions API supports.

bsudekum:
@1ec5 updated.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -315,6 +316,8 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var locale: Locale?
     
+    private let supportedLocales: [String] = ["de", "en", "es", "fr", "id", "nl", "ru", "sv", "vi", "zh-hans"]
```
1ec5: If the locale is Simplified Chinese, its locale identifier will be `zh-Hans`, not `zh-hans`. The Directions API accepts `zh-Hans` as a `language`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -332,7 +335,11 @@ open class RouteOptions: NSObject, NSSecureCoding {
         }
         
         if let locale = locale, let languageCode = (locale as NSLocale?)?.object(forKey: .languageCode) as? String {
-            params.append(URLQueryItem(name: "language", value: languageCode))
+            if supportedLocales.contains(languageCode) {
+                params.append(URLQueryItem(name: "language", value: languageCode))
+            } else {
+                params.append(URLQueryItem(name: "language", value: "en"))
```
1ec5: Is it necessary to append any `language` parameter in this case?

MapboxDirections/MBRouteOptions.swift
```diff
@@ -332,7 +335,11 @@ open class RouteOptions: NSObject, NSSecureCoding {
         }
         
         if let locale = locale, let languageCode = (locale as NSLocale?)?.object(forKey: .languageCode) as? String {
-            params.append(URLQueryItem(name: "language", value: languageCode))
+            if supportedLocales.contains(languageCode) {
```
1ec5: This will fail for a Locale with the identifier `zh-Hans`, since its `languageCode` will be just `zh`.

1ec5:
Added the <kbd>DO NOT MERGE</kbd> label because we’re looking into whether this language fallback logic can be implemented in the Directions API itself.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -319,6 +334,13 @@ open class RouteOptions: NSObject, NSSecureCoding {
             params.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
         }
         
+        if let locale = locale, var languageCode = (locale as NSLocale?)?.object(forKey: .languageCode) as? String, supportedLocales.contains(languageCode) {
+            if languageCode == "zh" {
+                languageCode = "zh-Hans"
+            }
+            params.append(URLQueryItem(name: "language", value: languageCode))
```
1ec5: Once we implement fallback logic on the server side, we’ll want to pass in `locale.localeIdentifier` here.

bsudekum: Updated.

bsudekum:
@1ec5 once https://github.com/Project-OSRM/osrm-text-instructions/pull/168 is deployed, we can start to look at merging this.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -304,6 +310,13 @@ open class RouteOptions: NSObject, NSSecureCoding {
     open var includesExitRoundaboutManeuver = false
     
     /**
+     If set, this `Locale` will be used to translate instructions.
+     
+     See supported languages [here](https://www.mapbox.com/api-documentation/#instructions-languages).
+     */
+    open var locale: Locale?
```
1ec5: It’d be nice to default to `Locale.autoupdatingCurrent` or `Bundle.main.preferredLocalizations.first` so the behavior is always what the user would expect, but that can happen as tail work, possibly at the same time that we set a default locale [in MapboxGeocoder.swift](https://github.com/mapbox/MapboxGeocoder.swift/blob/631481f2e01cb9151b7211895cb6a5842d6081d2/MapboxGeocoder/MBGeocodeOptions.swift#L56-L65).

bsudekum: Let's keep this optional for now.

PR #180, opened 2017-09-19
Add exclude option

bsudekum:
Adds the `exclude` option.

/cc @miccolis @1ec5

### DIFF
--- a/MapboxDirections/MBRoadClasses.h
+++ b/MapboxDirections/MBRoadClasses.h
@@ -34,5 +34,5 @@ typedef NS_OPTIONS(NSUInteger, MBRoadClasses) {
     MBRoadClassesFerry = (1 << 4),
     
     
-    MBRoadClassesNone = (1 << 5),
+    MBRoadClassesNone = (1 << 0),
 };
--- a/MapboxDirections/MBRoadClasses.h
+++ b/MapboxDirections/MBRoadClasses.h
@@ -32,7 +32,4 @@ typedef NS_OPTIONS(NSUInteger, MBRoadClasses) {
      In general, the transport type of the step containing the road segment is also `TransportType.ferry`.
      */
     MBRoadClassesFerry = (1 << 4),
-    
-    
-    MBRoadClassesNone = (1 << 0),
 };
--- a/MapboxDirections/MBRoadClasses.h
+++ b/MapboxDirections/MBRoadClasses.h
@@ -32,4 +32,7 @@ typedef NS_OPTIONS(NSUInteger, MBRoadClasses) {
      In general, the transport type of the step containing the road segment is also `TransportType.ferry`.
      */
     MBRoadClassesFerry = (1 << 4),
+    
+    
+    MBRoadClassesNone = (1 << 5),
 };
--- a/MapboxDirections/MBRoadClasses.swift
+++ b/MapboxDirections/MBRoadClasses.swift
@@ -46,9 +46,6 @@ extension RoadClasses: CustomStringConvertible {
         if contains(.ferry) {
             descriptions.append("ferry")
         }
-        if contains(.none) {
-            descriptions.append("none")
-        }
         return descriptions.joined(separator: ",")
     }
 }
--- a/MapboxDirections/MBRoadClasses.swift
+++ b/MapboxDirections/MBRoadClasses.swift
@@ -46,6 +46,9 @@ extension RoadClasses: CustomStringConvertible {
         if contains(.ferry) {
             descriptions.append("ferry")
         }
+        if contains(.none) {
+            descriptions.append("none")
+        }
         return descriptions.joined(separator: ",")
     }
 }
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -395,9 +395,9 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
     @objc open var distanceMeasurementSystem: MeasurementSystem = Locale.autoupdatingCurrent.usesMetric ? .metric : .imperial
 
     /**
-     An Array of `RoadClass` that the routing engine will attempt to avoid.
+     The route classes that the calculated routes will avoid.
      
-     Only a single `RoadClasses` can be provided.
+     Currently, you can only specify a single road class to avoid.
      */
     open var roadClassesToAvoid: RoadClasses = []
     
@@ -427,7 +427,7 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
             let allRoadClasses = roadClassesToAvoid.description.components(separatedBy: ",")
             let firstRoadClass = String(describing: allRoadClasses.first)
             if allRoadClasses.count > 1 {
-                assert(false, "`roadClassesToAvoid` only accepts one `RoadClasses`. Although \(roadClassesToAvoid.description) was specified, only \(firstRoadClass) will be used.")
+                assert(false, "`roadClassesToAvoid` only accepts one `RoadClasses`.")
             }
             params.append(URLQueryItem(name: "exclude", value: firstRoadClass))
         }
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -234,6 +234,12 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
         if let distanceMeasurementSystem = MeasurementSystem(description: decoder.decodeObject(of: NSString.self, forKey: "distanceMeasurementSystem") as String? ?? "") {
             self.distanceMeasurementSystem = distanceMeasurementSystem
         }
+        
+        guard let excludeClassesDescriptions = decoder.decodeObject(of: NSString.self, forKey: "excludeRoadClasses") as String?,
+            let excludeRoadClasses = RoadClasses(descriptions: excludeClassesDescriptions.components(separatedBy: ",")) else {
+                return nil
+        }
+        self.excludeRoadClasses = excludeRoadClasses
     }
 
     open static var supportsSecureCoding = true
@@ -251,6 +257,7 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
         coder.encode(locale, forKey: "locale")
         coder.encode(includesSpokenInstructions, forKey: "includesSpokenInstructions")
         coder.encode(distanceMeasurementSystem.description, forKey: "distanceMeasurementSystem")
+        coder.encode(excludeRoadClasses, forKey: "excludeRoadClasses")
     }
 
     // MARK: Specifying the Path of the Route
@@ -391,6 +398,13 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
     @objc open var distanceMeasurementSystem: MeasurementSystem = Locale.autoupdatingCurrent.usesMetric ? .metric : .imperial
 
     /**
+     An Array of `RoadClass` that the routing engine will attempt to avoid.
+     
+     The order of this Array does not impact the results.
+     */
+    open var excludeRoadClasses: RoadClasses? = []
+    
+    /**
      An array of URL parameters to include in the request URL.
      */
     internal var params: [URLQueryItem] {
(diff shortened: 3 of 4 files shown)


MapboxDirections/MBRouteOptions.swift
```diff
@@ -304,6 +312,13 @@ open class RouteOptions: NSObject, NSSecureCoding {
     open var includesExitRoundaboutManeuver = false
     
     /**
+     An Array of `RoadClass` that the routing engine will attempt to avoid.
```
1ec5: RoadClasses is an OptionSet, not an Array. OptionSets just happen to use the same `[]` syntax, but it’s perfectly valid to set `intersection.outletRoadClasses = .toll`, whereas that would be invalid if `roadClasses` were of type `[RoadClasses]`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -304,6 +312,13 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
+     
+     The order of this Array does not impact the results.
+     */
+    open var excludeRoadClasses: RoadClasses? = []
```
1ec5: “Exclude road classes” sounds like a verb, which makes this Objective-C syntax look like it’s performing an action rather than getting a property value:

```objc
[routeOptions excludeRoadClasses];
```

`excludesRoadClasses` is also problematic, because it sounds like a Boolean (as in, “does this property exclude road classes?”). So `excludedRoadClasses` would be the best name for this property.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -195,6 +202,7 @@ open class RouteOptions: NSObject, NSSecureCoding {
         coder.encode(routeShapeResolution.description, forKey: "routeShapeResolution")
         coder.encode(attributeOptions.description, forKey: "attributeOptions")
         coder.encode(includesExitRoundaboutManeuver, forKey: "includesExitRoundaboutManeuver")
+        coder.encode(excludeRoadClasses, forKey: "excludeRoadClasses")
```
1ec5: Encode `excludedRoadClasses.description`, which is a string, not `excludeRoadClasses`, whose underlying type is `Int` and thus more fragile.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -181,6 +181,13 @@ open class RouteOptions: NSObject, NSSecureCoding {
         self.attributeOptions = attributeOptions
         
         includesExitRoundaboutManeuver = decoder.decodeBool(forKey: "includesExitRoundaboutManeuver")
+        
+        
+        guard let excludeClassesDescriptions = decoder.decodeObject(of: NSString.self, forKey: "excludeRoadClasses") as String?,
+            let excludeRoadClasses = RoadClasses(descriptions: excludeClassesDescriptions.components(separatedBy: ",")) else {
+                return nil
```
1ec5: By returning `nil` here, we’re rejecting any encoded RouteOptions object that lacks an `excludedRoadClasses`, even one that was created before this PR lands. Let’s quietly set `excludedRoadClasses` to `nil` instead.

MapboxDirections/MBRouteOptions.swift
miccolis: Presently we do not support more than one road class. You may avoid one or another, but not more than one.

1ec5: If there’s the potential that the API might support more than one excluded road class in the future, let’s leave the code as is but change the documentation to avoid using the term “array” and assert at runtime that only one is set. Otherwise, if there’s a fundamental reason why the API wouldn’t support more than one, then I guess we could consider a separate enumeration that doesn’t bridge as an OptionSet, but that would be much less elegant.

MapboxDirections/MBRouteOptions.swift
1ec5: An optional `NS_OPTION` doesn’t bridge to Objective-C. To make this property bridge, we’d have to introduce a `none` RoadClass and make both this property and `Intersection.outletRoadClasses` non-optional.

MapboxDirections/MBRoadClasses.h
```diff
@@ -32,4 +32,7 @@ typedef NS_OPTIONS(NSUInteger, MBRoadClasses) {
      In general, the transport type of the step containing the road segment is also `TransportType.ferry`.
      */
     MBRoadClassesFerry = (1 << 4),
+    
+    
+    MBRoadClassesNone = (1 << 5),
```
1ec5: Let’s give `MBRoadClassesNone` the value 0.

MapboxDirections/MBRoadClasses.swift
```diff
@@ -46,6 +46,9 @@ extension RoadClasses: CustomStringConvertible {
         if contains(.ferry) {
             descriptions.append("ferry")
         }
+        if contains(.none) {
+            descriptions.append("none")
```
1ec5: Is `none` a valid value for the `exclude` URL query parameter? If not, omit this line so nothing gets added to the `descriptions` array.

MapboxDirections/MBRouteOptions.swift
1ec5:
```swift
let excludedRoadClassesDescriptions = decoder.decodeObject(of: NSString.self, forKey: "excludedRoadClasses") as String?
excludedRoadClasses = RoadClasses(descriptions: excludedRoadClassesDescriptions?.components(separatedBy: ",") ?? "")
```

MapboxDirections/MBRouteOptions.swift
```diff
@@ -181,6 +181,10 @@ open class RouteOptions: NSObject, NSSecureCoding {
         self.attributeOptions = attributeOptions
         
         includesExitRoundaboutManeuver = decoder.decodeBool(forKey: "includesExitRoundaboutManeuver")
+        
+        
+        let excludedRoadClassesDescriptions = decoder.decodeObject(of: NSString.self, forKey: "excludedRoadClasses") as String?
+        excludedRoadClasses = RoadClasses(descriptions: excludedRoadClassesDescriptions?.components(separatedBy: ",") ?? [""]) ?? []
```
1ec5: Nit: replace `[""]` with `[]`.

MapboxDirections/MBRouteOptions.swift
1ec5: How about:

> The road classes that the resulting routes will avoid.
> 
> For example, you can set this property to `.toll` to avoid toll roads. No road classes are avoided by default. Currently, it is only possible to avoid one road class at a time.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -304,6 +309,13 @@ open class RouteOptions: NSObject, NSSecureCoding {
     open var includesExitRoundaboutManeuver = false
     
     /**
+     An Array of `RoadClass` that the routing engine will attempt to avoid.
+     
+     The order of this Array does not impact the results.
+     */
+    open var excludedRoadClasses: RoadClasses = []
```
1ec5: Sorry for the indecision here, but after writing the documentation suggestion above, it just occurred to me that `roadClassesToAvoid` would be even more readily understood than `excludedRoadClasses`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -319,6 +331,10 @@ open class RouteOptions: NSObject, NSSecureCoding {
             params.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
         }
         
+        if !excludedRoadClasses.isEmpty {
+            params.append(URLQueryItem(name: "exclude", value: excludedRoadClasses.description))
```
1ec5: This would be a good place to assert that only one option is specified in `roadClassesToAvoid`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -331,8 +331,12 @@ open class RouteOptions: NSObject, NSSecureCoding {
...
         }
         
-        if !excludedRoadClasses.isEmpty {
-            params.append(URLQueryItem(name: "exclude", value: excludedRoadClasses.description))
+        if !roadClassesToAvoid.isEmpty {
+            let allRoadClasses = roadClassesToAvoid.description.components(separatedBy: ",")
+            if allRoadClasses.count > 1 {
+                print("`roadClassesToAvoid` only accepts one `RoadClasses`. Although \(roadClassesToAvoid.description) was specified, only \(String(describing: allRoadClasses.first)) will be used.")
+            }
+            params.append(URLQueryItem(name: "exclude", value: roadClassesToAvoid.description))
```
1ec5: Given the assertion above, we should include (er, exclude) only the first roadClass.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -309,11 +309,11 @@ open class RouteOptions: NSObject, NSSecureCoding {
     open var includesExitRoundaboutManeuver = false
     
     /**
-     An Array of `RoadClass` that the routing engine will attempt to avoid.
+     The road classes that the resulting routes will avoid.
      
-     The order of this Array does not impact the results.
+     For example, you can set this property to `.toll` to avoid toll roads. No road classes are avoided by default. Currently, it is only possible to avoid one road class at a time.
```
1ec5: > If you specify more than one road class, which class is avoided is undefined.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -319,6 +331,14 @@ open class RouteOptions: NSObject, NSSecureCoding {
             params.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
         }
         
+        if !roadClassesToAvoid.isEmpty {
+            let allRoadClasses = roadClassesToAvoid.description.components(separatedBy: ",")
+            if allRoadClasses.count > 1 {
+                print("`roadClassesToAvoid` only accepts one `RoadClasses`. Although \(roadClassesToAvoid.description) was specified, only \(String(describing: allRoadClasses.first)) will be used.")
```
1ec5: Given the API’s behavior of erroring out on multiple excluded classes, let’s make this an assertion.

bsudekum:
@1ec5 you think we could resurrect this PR?

MapboxDirections/MBRouteOptions.swift
```diff
@@ -391,6 +395,13 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
     open var distanceMeasurementSystem: MeasurementSystem = Locale.autoupdatingCurrent.usesMetric ? .metric : .imperial
     
 
+    /**
+     An Array of `RoadClass` that the routing engine will attempt to avoid.
```
1ec5: Typo: “array”, not “Array”.

> The route classes that the calculated routes will avoid.
> 
> Currently, you can only specify a single road class to avoid.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -412,6 +423,15 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
             params.append(URLQueryItem(name: "voice_instructions", value: String(includesSpokenInstructions)))
             params.append(URLQueryItem(name: "voice_units", value: String(describing: distanceMeasurementSystem)))
         }
+        
+        if !roadClassesToAvoid.isEmpty {
+            let allRoadClasses = roadClassesToAvoid.description.components(separatedBy: ",")
+            let firstRoadClass = String(describing: allRoadClasses.first)
+            if allRoadClasses.count > 1 {
+                assert(false, "`roadClassesToAvoid` only accepts one `RoadClasses`. Although \(roadClassesToAvoid.description) was specified, only \(firstRoadClass) will be used.")
```
1ec5: No road classes will be used, because this assertion will crash the application.

MapboxDirectionsTests/RouteOptionsTests.swift
```diff
@@ -32,6 +32,7 @@ class RouteOptionsTests: XCTestCase {
         XCTAssertEqual(unarchivedOptions.locale, options.locale)
         XCTAssertEqual(unarchivedOptions.includesSpokenInstructions, options.includesSpokenInstructions)
         XCTAssertEqual(unarchivedOptions.distanceMeasurementSystem, options.distanceMeasurementSystem)
+        XCTAssertEqual(unarchivedOptions.roadClassesToAvoid, options.roadClassesToAvoid)
```
1ec5: Update `RouteOptions.testInstance` to specify `roadClassesToAvoid`. Otherwise, this assertion is meaningless.

bsudekum:
@1ec5 updated.

PR #181, opened 2017-09-26
Pin polyline to v4.1.1

bsudekum:
Carthage has an issue where `~> 4.1.1` resolves to v4.2.0 when running carthage update.

This pins it to the last swift 3 version.

/cc @1ec5

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "raphaelmor/Polyline" ~> 4.1.1
+github "raphaelmor/Polyline" == 4.1.1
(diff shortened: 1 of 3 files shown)


1ec5:
The incompatibility was introduced in raphaelmor/Polyline#45.

According to [this documentation](https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#version-requirement), downloading v4.2.0 automatically is intended behavior:

> This means that any version greater than or equal to 1.5.1, but less than 2.0, will be considered “compatible” with 1.5.1.

We can pin the dependency to the last Swift 3 release, but that might introduce dependency incompatibilities for projects that directly use Polyline.

bsudekum:
Yes, I think so. @1ec5 what will be the behavior when using 4.1.1 in a project that uses swift 4?

bsudekum:
@1ec5 update podspec

1ec5:
> what will be the behavior when using 4.1.1 in a project that uses swift 4?

It’ll build fine as long as the `SWIFT_VERSION` build setting is set correctly for that target. This is guaranteed to be the case when installing via Carthage; I’m less certain about CocoaPods.

PR #187, opened 2017-10-11
Properly decode locale optional

bsudekum:
Followup to https://github.com/mapbox/MapboxDirections.swift/pull/176

Locale is optional so we should only be setting this optional value if it's set in the first place. I'm currently running into issues where route created without a `locale` are not conforming to `Route`.

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -225,9 +225,7 @@ open class RouteOptions: NSObject, NSSecureCoding {
 
         includesExitRoundaboutManeuver = decoder.decodeBool(forKey: "includesExitRoundaboutManeuver")
         
-        if let locale = decoder.decodeObject(of: NSLocale.self, forKey: "locale") as Locale? {
-            self.locale = locale
-        }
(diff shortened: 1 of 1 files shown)


bsudekum:
Example: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxNavigationTests/MapboxNavigationTests.swift#L128

MapboxDirections/MBRouteOptions.swift
```diff
@@ -225,10 +225,9 @@ open class RouteOptions: NSObject, NSSecureCoding {
 
         includesExitRoundaboutManeuver = decoder.decodeBool(forKey: "includesExitRoundaboutManeuver")
         
-        guard let locale = decoder.decodeObject(of: NSLocale.self, forKey: "locale") as Locale? else {
-            return nil
+        if let locale = decoder.decodeObject(of: NSLocale.self, forKey: "locale") as Locale? {
```
1ec5: Since the locale property is optional, you can set it directly without the if statement.

PR #196, opened 2017-10-23
Upgrade to swift 4

bsudekum:
/cc @1ec5 @frederoni


bsudekum:
@1ec5 should this also be changed? https://github.com/mapbox/MapboxDirections.swift/blob/master/.swift-version

1ec5:
Yes, I think we need to change that .swift-version file, which is what CocoaPods uses to determine the target's Swift version.

bsudekum:
@1ec5 updated this since your last 👀 . Can you take a look again?

README.md
```diff
@@ -29,7 +29,7 @@ pod 'MapboxDirections.swift', '~> 0.12'
 
 Then `import MapboxDirections` or `@import MapboxDirections;`.
 
-v0.7.0 is the last release of MapboxDirections.swift written in Swift 2.3. The `swift2.3` branch corresponds to this release, plus any critical bug fixes that have been applied since. All subsequent releases will be based on the `master` branch, which is written in Swift 3. The Swift examples below are written in Swift 3; see the `swift2.3` branch’s readme for Swift 2.3 examples.
+v0.12.1 is the last release of MapboxDirections.swift written in Swift 3. All subsequent releases will be based on the `master` branch, which is written in Swift 4. The Swift examples below are written in Swift 4.
```
1ec5: Before merging, cut a branch called swift3.2 based on master. That’ll be our answer for folks who need Swift 3 in the near term. Like the swift2.3 branch, which we can probably delete now, we won’t be adding new features or nonessential fixes to that branch.

bsudekum: Done, https://github.com/mapbox/MapboxDirections.swift/tree/swift3.2

PR #197, opened 2017-10-26
Visual instructions

bsudekum:
Adds optional `inclueBannerInstructions` property to RouteOptions for including banner instructions and the ability to decode them on `RouteStep`.

Example of a banner object:

```
bannerInstructions: [
    {
        distanceAlongGeometry: 1405.3,
        primaryText: {
            text: "Bryant Street",
            components: [
                {
                    text: "Bryant Street"
                }
            ]
        },
        secondaryText: null
    }
]
```

/cc @1ec5 @ericrwolfe @frederoni @mcwhittemore


MapboxDirections/MBRouteOptions.swift
```diff
@@ -387,6 +387,8 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var distanceMeasurementSystem: MeasurementSystem = Locale.current.usesMetric ? .metric : .imperial
     
+    open var inclueBannerInstructions = false
```
ericrwolfe: Typo `inclue`

MapboxDirections/MBRouteOptions.swift
```diff
@@ -387,6 +387,8 @@ open class RouteOptions: NSObject, NSSecureCoding {
      */
     open var distanceMeasurementSystem: MeasurementSystem = Locale.current.usesMetric ? .metric : .imperial
     
+    open var includeBannerInstructions = false
```
1ec5: Rename this property to `includesBannerInstructions` so it doesn’t look like a method in Objective-C.

1ec5: This property needs documentation.

MapboxDirections/MBBannerInstructionComponent.swift
```diff
@@ -0,0 +1,34 @@
+import Foundation
+
+
+@objc(MBBannerInstructionComponent)
```
1ec5: This class needs documentation.

MapboxDirections/MBBannerInstructionComponent.swift
```diff
@@ -0,0 +1,34 @@
...
+public class BannerInstructionComponent: NSObject, NSSecureCoding {
+    
+    public let text: String
+    
+    public let components: [BannerInstructionDetailComponent]
```
1ec5: I don’t follow: a banner instruction component contains components?

MapboxDirections/MBBannerInstructions.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+
+@objc(MBBannerInstruction)
+public class BannerInstruction: NSObject, NSSecureCoding {
```
1ec5: MapboxDirections.swift shouldn’t assume a particular user interface paradigm. Are we sure this information will only ever be used for a turn banner, as opposed to a step table, local notification, or HUD? Let’s call this class VisualInstruction, unless there’s an overriding reason to refer to banners specifically.

MapboxDirections/MBBannerInstructionComponent.swift
bsudekum: @1ec5 what I'm thinking here:

```
bannerInstructions: [                  // [BannerInstructionComponent]
    {
        distanceAlongGeometry: 1405.3, // Double
        primaryText: {                 // BannerInstructionComponent
            text: "Bryant Street",     // String
            components: [            // [BannerInstructionDetailComponent]
                {
                    text: "Bryant Street"  // String
                }
            ]
        },
        secondaryText: null
    }
]
```

1ec5: Let’s flatten things down a little:

* `bannerInstructions` → `RouteStep.visualInstructions: [VisualInstruction]`
* `bannerInstructions[0].primaryText.text` → `RouteStep.visualInstructions[0].primaryText: String`
* `bannerInstructions[0].primaryText.components` → `RouteStep.visualInstructions[0].primaryTextComponents: [VisualInstructionComponent]`
* `bannerInstructions[0].secondaryText.text` → `RouteStep.visualInstructions[0].secondaryText: String`
* `bannerInstructions[0].secondaryText.components` → `RouteStep.visualInstructions[0].secondaryTextComponents: [VisualInstructionComponent]`

1ec5:
As I understand it, this feature is designed for internal use by the navigation SDK and not by arbitrary clients of MapboxDirections.swift. If that’s the case, we should mark each of the documentation comments with `/// :nodoc:` (but still document them internally). Additionally, since the navigation SDKs are written in Swift and Objective-C compatibility becomes a non-issue, we can simplify these classes into Swift structs that don’t require custom encoding/decoding or even initializers.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,93 @@
...
+     @1x PNG to inline in visual instruction.
+     */
+    public var imageURL1x: URL? = nil
+    
+    
+    /**
+     :nodoc:
+     @2x PNG to inline in visual instruction.
+     */
+    public var imageURL2x: URL? = nil
```
bsudekum: @frederoni @1ec5 @JThramer what do you think of this api design?

JThramer: Well if this was swift-only, I think this should be encapsulated in a named tuple or a struct, but OBJ-C continues to bind our hands in this regard (until we start [playing with](http://nshipster.com/new-years-2016/) `__ObjCConvertable`, anyway...)

How does the rest of the class feel about using objects to encapsulate closely-related properties? It will be way less of a breaking change for an API user to go from Object -> Struct down the road then from 4 or 5 top-level properties into a struct.

bsudekum:
Let's not make this swift only.

MapboxDirections/MBVisualInstructionDetailComponent.swift
1ec5: Combine the various `imageURL*x` properties into a single dictionary mapping UITraitCollections (created using `UITraitCollection(displayScale:)`) to image URLs. This will make it easier for the client to build a UIImageAsset from the various URLs if desired, and it gives us the flexibility to vend different images based on orientation or other traits in the future. A `[UITraitCollection: URL]` also bridges to Objective-C as is.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,93 @@
...
+     @3x PNG to inline in visual instruction.
+     */
+    public var imageURL3x: URL? = nil
+    
+    
+    /**
+     :nodoc:
+     SVG to inline in visual instruction.
+     */
+    public var imageURLSVG: URL? = nil
```
1ec5: Given that SVG isn’t natively supported on Apple platforms, do we need this property? If there’s a good case for SVG in the future, then we can add the property in later.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -0,0 +1,48 @@
...
+     Basical text representation of visual component of a `RouteStep`.
+     */
+    public let text: String
+    
+    
+    /**
+     :nodoc:
+     A more detailed representation of the a `RouteStep`. With components, it's possible to show an image representation of highway shields.
+     */
+    public let components: [VisualInstructionDetailComponent]
```
1ec5: Per https://github.com/mapbox/MapboxDirections.swift/pull/197#discussion_r148165905, I still don’t see a point to a component class that contains instances of another component class. Why this extra level of indirection, as opposed to putting `primaryText` and `primaryComponents` directly on VisualInstruction?

MapboxDirections/MBVisualInstructionDetailComponent.swift
bsudekum: Noting, this will require importing UIKit.

1ec5: Conditionally import UIKit for iOS and tvOS.

For macOS, it’s possible to create an NSImage that has representations at multiple scales, and it’s also possible to create an NSImage lazily loads a URL using `NSImage(representingUrl:)`, but I don’t think it’s possible to combine the two features. So `[NSNumber: URL]` appears to be our only option on that platform.

bsudekum:
@1ec5 this is ready for review.

MapboxDirections/MBRouteStep.swift
```diff
@@ -743,6 +749,12 @@ open class RouteStep: NSObject, NSSecureCoding {
      */
     open let instructionsSpokenAlongStep: [SpokenInstruction]?
     
+    /**
+     :nodoc:
+     Contains an array of `VisualInstruction` used for visually conveying information about a given `RouteStep`.
+     */
+    open let visualInstructionsAlongStep: [VisualInstruction]?
```
1ec5: Rename this property `instructionsDisplayedAlongStep`. This name is more consistent with `instructionsSpokenAlongStep` and also emphasizes that the visual instructions are displayed along the containing step but don’t pertain to the containing step.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -0,0 +1,82 @@
+import Foundation
+
+/**
+ :nodoc:
+ Encompasses all information necessary for creating a visual queue about a given `RouteStep`.
```
1ec5: A “queue” is a waiting line; a “cue” is a prompt.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -0,0 +1,82 @@
...
+ */
+@objc(MBVisualInstruction)
+public class VisualInstruction: NSObject, NSSecureCoding {
+    
+    /**
+     :nodoc:
+     Distance in meters from the beginning of the step at which the visual instruction should be visible.
+     */
+    public let distanceAlongStep: CLLocationDistance
```
1ec5: Consider factoring out `distanceAlongStep` from both VisualInstruction and SpokenInstruction into a new PositionedInstruction protocol. (Use `MBPositionedInstruction` as the Objective-C name if it’s necessary to expose this protocol to Objective-C.)

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -0,0 +1,82 @@
...
+    
+    
+    public let primaryText: String
```
1ec5: This property and `secondaryText` need documentation comments.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     :nodoc:
+     Most important visual content to convey to the user about the `RouteStep`.
```
1ec5: Indicate that this is the structured representation of the value that’s in the `primaryText` property; likewise for `secondaryTextComponents` and `secondaryText`.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -0,0 +1,82 @@
...
+        self.primaryText = primaryText
+        
+        primaryTextComponents = decoder.decodeObject(of: [NSArray.self, VisualInstructionComponent.self], forKey: "primaryTextComponents") as? [VisualInstructionComponent] ?? []
+        
+        guard let secondaryText = decoder.decodeObject(of: NSString.self, forKey: "primarysecondaryTextText") as String? else {
+            return nil
+        }
+        self.secondaryText = secondaryText
+        
+        secondaryTextComponents = decoder.decodeObject(of: [NSArray.self, VisualInstructionComponent.self], forKey: "primaryTextComponents") as? [VisualInstructionComponent]
```
1ec5: Add unit tests to ensure that this class can round-trip through encoding and decoding.

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -17,6 +17,14 @@
...
 		C51538CC1E807FF00093FF3E /* MBAttribute.swift in Sources */ = {isa = PBXBuildFile; fileRef = C51538CB1E807FF00093FF3E /* MBAttribute.swift */; };
 		C5247D711E818A24004B6154 /* AnnotationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = C5247D701E818A24004B6154 /* AnnotationTests.swift */; };
+		C52552B91FA15D5900B1545C /* MBVisualInstruction.swift in Sources */ = {isa = PBXBuildFile; fileRef = C52552B81FA15D5900B1545C /* MBVisualInstruction.swift */; };
+		C52552BA1FA15D5E00B1545C /* MBVisualInstruction.swift in Sources */ = {isa = PBXBuildFile; fileRef = C52552B81FA15D5900B1545C /* MBVisualInstruction.swift */; };
+		C52552BB1FA15D5F00B1545C /* MBVisualInstruction.swift in Sources */ = {isa = PBXBuildFile; fileRef = C52552B81FA15D5900B1545C /* MBVisualInstruction.swift */; };
+		C52552BC1FA15D6000B1545C /* MBVisualInstruction.swift in Sources */ = {isa = PBXBuildFile; fileRef = C52552B81FA15D5900B1545C /* MBVisualInstruction.swift */; };
+		C52552C01FA1628A00B1545C /* MBVisualInstructionDetailComponent.swift in Sources */ = {isa = PBXBuildFile; fileRef = C52552BF1FA1628A00B1545C /* MBVisualInstructionDetailComponent.swift */; };
```
1ec5: Rename this file to match its primary class, MBVisualInstructionComponent.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    /**
+     :nodoc:
+     Single part of a visual instruction.
+     */
+    public let text: String?
+    
+    /**
+     :nodoc:
+     Dictionary containing `UITraitCollection` for scales 1.0, 2.0 and 3.0. Each key's value is a `URL`.
```
1ec5: There should be two documentation comments, one within the `#if` block:

> A dictionary containing the URLs to image representations of this component.
> 
> The keys of this dictionary are trait collections specifying different display scales. The values of this dictionary are URLs to remote images that you should prefer over the `text` property once they are available. Use the URL that is best suited to the system’s current traits.

and the other within the `#else` block

> A dictionary containing the URLs to image representations of this component.
> 
> The keys of this dictionary are screen scale factors. The values of this dictionary are URLs to remote images that you should prefer over the `text` property once they are available. Use the URL that is best suited to the current screen’s scale factor.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,76 @@
...
+     */
+    #if os(iOS) || os(tvOS)
+        public var imageURLS: [UITraitCollection: URL] = [:]
```
1ec5: Typo: rename this property to `imageURLs`.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+
+#if os(iOS) || os(tvOS)
+    import UIKit
+#endif
+
+
+/**
+ :nodoc:
+ Contains necessary features to build a more rich visual experience for a `RouteStep`.
```
1ec5: > A component of a `VisualInstruction` that represents a single run of similarly formatted text or an image with a textual fallback representation.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,76 @@
...
+ */
+@objc(MBVisualInstructionComponent)
+public class VisualInstructionComponent: NSObject, NSSecureCoding {
+    
+    /**
+     :nodoc:
+     Single part of a visual instruction.
```
1ec5: > The plain text representation of this component.
> 
> Use this property if `imageURLs` is an empty dictionary or if the URLs contained in that property are not yet available.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,76 @@
...
+        
+        if let baseURL = json["imageBaseURL"] as? String {
+            let oneXURL = URL(string: "\(baseURL)@1x.png")
+            let twoXURL = URL(string: "\(baseURL)@2x.png")
+            let threeXURL = URL(string: "\(baseURL)@3x.png")
+            
+            #if os(iOS) || os(tvOS)
+                imageURLS[UITraitCollection(displayScale: 1.0)] = oneXURL
+                imageURLS[UITraitCollection(displayScale: 2.0)] = twoXURL
+                imageURLS[UITraitCollection(displayScale: 3.0)] = threeXURL
```
1ec5: Upon seeing this code, I have to wonder: do we need to store the URLs for all three scale factors if the current screen can only have one scale factor at a time? It would be simpler to store just one image URL computed using the same logic seen here:

https://github.com/mapbox/MapboxStatic.swift/blob/279ce127c788f0bcaee15e214a1bee433e22eeca/MapboxStatic/SnapshotOptions.swift#L220-L247

macOS applications would be unable to account for changes to the scale factor, such as when the window containing the component gets dragged from a LoDPI screen to a HiDPI screen mid-route. But that’s multiple levels of edge case that we hopefully aren’t worrying about right now.

The only other reason I can think of for keeping track of multiple scale factors is that the navigation SDK might choose to display a lower-resolution image while on a poor connection and a best-resolution image while on a good connection. But these images are so small that I can’t imagine the scale factor mattering much even then.

MapboxDirections/MBVisualInstructionDetailComponent.swift
```diff
@@ -0,0 +1,76 @@
...
+            guard let imageURLS = decoder.decodeObject(of: [NSDictionary.self, UITraitCollection.self, NSURL.self], forKey: "imageURLS") as? [UITraitCollection: URL] else {
+                return nil
+            }
+        #else
+            guard let imageURLS = decoder.decodeObject(of: [NSDictionary.self, NSNumber.self, NSURL.self], forKey: "imageURLS") as? [NSNumber: URL] else {
+                return nil
+            }
+        #endif
+        
+        self.imageURLS = imageURLS
```
1ec5: Please add a unit test for round-tripping this class through encoding and decoding.

MapboxDirections/MBRouteStep.swift
```diff
@@ -674,6 +679,7 @@ open class RouteStep: NSObject, NSSecureCoding {
         coder.encode(destinationCodes, forKey: "destinationCodes")
         coder.encode(destinations, forKey: "destinations")
         coder.encode(instructionsSpokenAlongStep, forKey: "instructionsSpokenAlongStep")
+        coder.encode(visualInstructionsAlongStep, forKey: "visualInstructionsAlongStep")
```
1ec5: Please exercise this code in RouteStepTests and V5Tests.

MapboxDirections/MBRouteStep.swift
```diff
@@ -743,6 +749,12 @@ open class RouteStep: NSObject, NSSecureCoding {
      */
     open let instructionsSpokenAlongStep: [SpokenInstruction]?
     
+    /**
+     :nodoc:
+     Contains an array of `VisualInstruction` used for visually conveying information about a given `RouteStep`.
```
1ec5: > Instructions about the next step’s maneuver, optimized for display.
> 
> As the user traverses this step, you can give them advance notice of the upcoming maneuver by displaying each item in this array in order as the user reaches the specified distances along this step. The text and images of the visual instructions refer to the details in the next step, but the distances are measured from the beginning of this step.
> 
> This property is non-`nil` if the `RouteOptions.includesVisualInstructions` option is set to `true`. For instructions designed for speech synthesis, use the `instructionsSpokenAlongStep` property. For instructions designed for display in a static list, use the `instructions` property.

MapboxDirectionsTests/RouteStepTests.swift
```diff
@@ -157,5 +157,6 @@ class RouteStepTests: XCTestCase {
         XCTAssertEqual(unarchivedStep.transportType, step.transportType)
         XCTAssertEqual(unarchivedStep.destinations ?? [], step.destinations ?? [])
         XCTAssertEqual(unarchivedStep.instructionsSpokenAlongStep ?? [], step.instructionsSpokenAlongStep ?? [])
+        XCTAssertEqual(unarchivedStep.instructionsDisplayedAlongStep ?? [], step.instructionsDisplayedAlongStep ?? [])
```
1ec5: Besides coding tests, is there anything else we can test about this new feature? I would’ve expected a test fixture to be updated as part of adding tests.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -20,43 +23,47 @@ public class VisualInstructionComponent: NSObject, NSSecureCoding {
...
+     The scale factor of the image.
      
-        The keys of this dictionary are screen scale factors. The values of this dictionary are URLs to remote images that you should prefer over the `text` property once they are available. Use the URL that is best suited to the current screen’s scale factor.
-        */
-        public var imageURLs: [NSNumber: URL] = [:]
+     If you multiply the logical size of the image (stored in the `size` property) by the value in this property, you get the dimensions of the image in pixels.
+     
+     The default value of this property matches the natural scale factor associated with the main screen. However, only images with a scale factor of 1.0, 2.0, or 3.0 are ever returned by the API, so a scale factor of 1.0 of less results in a 1× (standard-resolution) image, while a scale factor greater than 1.0 results in a 2× (high-resolution or Retina) image.
+     */
+    open var scale: CGFloat = UIScreen.main.scale
```
1ec5: Oh, I didn’t mean to imply that the entire set of properties needed to be copied from MapboxStatic.swift, just that we would choose a scale automatically using `UIScreen.main.scale` or whatever the macOS or watchOS equivalent would be.

I think you can safely inline this code, but if you think it’s important to make the scale configurable, then a better way would be to replace `imageURL` with a method that takes the scale as an argument. Because this class doesn’t need to remember the scale.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -20,43 +23,47 @@ public class VisualInstructionComponent: NSObject, NSSecureCoding {
...
     #endif
     
+    /**
+    URL to image representations of this component.
+ 
+    By default, an image based on the devices scale will be used. If you'd like a different scale, mutate the `scale` property.
```
1ec5: Typo: “device’s”.

MapboxDirectionsTests/IntructionsTests.swift
```diff
@@ -68,15 +66,22 @@ class SpokenInstructionsTests: XCTestCase {
...
+        XCTAssertEqual(spokenInstructions[1].ssmlText, "<speak><amazon:effect name=\"drc\"><prosody rate=\"1.08\">In a quarter mile, turn left onto John F Kennedy Drive</prosody></amazon:effect></speak>")
+        XCTAssertEqual(spokenInstructions[2].ssmlText, "<speak><amazon:effect name=\"drc\"><prosody rate=\"1.08\">Turn left onto John F Kennedy Drive</prosody></amazon:effect></speak>")
         
-        XCTAssertEqual(spokenInstructions[0].text, "Head south on 8th Avenue")
-        XCTAssertEqual(spokenInstructions[1].text, "In 1000 feet, turn left onto John F Kennedy Drive")
+        XCTAssertEqual(spokenInstructions[0].text, "Head south on 8th Avenue for a half mile")
+        XCTAssertEqual(spokenInstructions[1].text, "In a quarter mile, turn left onto John F Kennedy Drive")
         XCTAssertEqual(spokenInstructions[2].text, "Turn left onto John F Kennedy Drive")
+        
+        let visualInstructions = step.instructionsDisplayedAlongStep!
```
1ec5: If `instructionsDisplayedAlongStep` happens to be `nil`, the test will crash and you’ll get a pretty unintelligible error on CI. At least add a non-nil assertion before force-unwrapping, and consider chaining optionals to avoid having to force-unwrap.

1ec5:
Also, a reminder to make the small change in https://github.com/mapbox/MapboxDirections.swift/pull/197#discussion_r151308060 too.

MapboxDirections/MBVisualInstructionComponent.swift
1ec5: #213 is what I meant by inlining the code.

Issue #121, opened 2017-04-05
Built-in support for OSRM


1ec5:
This library should offer built-in support for talking to an OSRM server directly. Similar to the existing support for Directions API v4, an OSRMRouteOptions class would inherit from RouteOptions, and we'd have subclasses of Route, RouteLeg, and RouteStep that perform any necessary translation of the JSON data into the Directions API format.

/ref mapbox/MapboxNavigation.swift#46
/cc @danpat

1ec5:
The scope of this feature would be a minimal compatibility layer; I don't think it would be worth porting all of the Directions API's logic directly to Swift. 😉

@freenerd points out that the Directions object would need to avoid setting the access_token parameter when used in conjunction with an OSRMRouteOptions object.

1ec5:
Copied from https://github.com/mapbox/mapbox-navigation-ios/issues/46#issuecomment-296352234:

> Since MapboxDirections.swift currently only knows how to issue requests to the Mapbox Directions API, you’ll need to write your own code to construct the request URL to your OSRM server and send a request using URLRequest (in Swift) or NSURLRequest (in Objective-C). Then, use JSONSerialization/NSJSONSerialization to turn the response data into a Dictionary/NSDictionary before passing it to the Route initializer.

> You’ll want to compare the [OSRM response format](https://github.com/Project-OSRM/osrm-backend/blob/master/docs/http.md#route-object) with the [Mapbox Directions API response format](https://www.mapbox.com/api-documentation/#directions-response-object). Modify the dictionary representation of the response data to match the Directions API format. They already look very similar; I couldn’t spot any differences at a glance.

PR #198, opened 2017-10-27
Osrm support

RomainQuidet:
I just updated the directions init to really support any custom host. This allows us to use any OSRM server.


RomainQuidet:
this is for issue #121

RomainQuidet:
any clue why bitrise fails ? It compiles on Xcode 9.1 / swift 3.2

bsudekum:
It looks like requests are still being made with a Mapbox access token (or it's expecting one to be there):

```
Selected tests
Test Suite MapboxDirectionsTests.xctest started
V4Tests
    ✗ testPolyline, XCTAssertNil failed: "Error Domain=MBDirectionsErrorDomain Code=-1 "Not Authorized - Invalid Token" UserInfo={NSLocalizedFailureReason=Not Authorized - Invalid Token}" - Error: Error Domain=MBDirectionsErrorDomain Code=-1 "Not Authorized - Invalid Token" UserInfo={NSLocalizedFailureReason=Not Authorized - Invalid Token}
    ✗ testPolyline, XCTAssertNotNil failed -
❌  fatal error: unexpectedly found nil while unwrapping an Optional value
```

RomainQuidet:
Ha yes sorry I need to check the unit tests ! Thanks.

MapboxDirections/MBDirections.swift
```diff
@@ -128,14 +131,27 @@ open class Directions: NSObject {
...
+        if host == nil {
+            let accessToken = accessToken ?? defaultAccessToken
+            assert(accessToken != nil && !accessToken!.isEmpty, "A Mapbox access token is required. Go to <https://www.mapbox.com/studio/account/tokens/>. In Info.plist, set the MGLMapboxAccessToken key to your access token, or use the Directions(accessToken:host:) initializer.")
+
+            self.accessToken = accessToken
+			baseURLComponents = URLComponents()
+			baseURLComponents.scheme = "https"
+			baseURLComponents.host = mapboxHost
+        }
+		else
```
bsudekum: Could you fix the indentation here and also put curly braces on the same line as the else? `} else {`

MapboxDirections/MBDirections.swift
```diff
@@ -229,11 +245,26 @@ open class Directions: NSObject {
...
+            path = options.path
+        }
+        else
+        {
+            assert(!options.queries.isEmpty, "No query")
+            let queryComponent = options.queries.joined(separator: ";")
+            path = "route/v1/\(options.profileIdentifier.rawValue)/\(queryComponent)"
+        }
+
+        let unparameterizedURL = URL(string: path, relativeTo: apiEndpoint)!
```
bsudekum: This should be properly unwrapped.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -335,7 +335,7 @@ open class RouteOptions: NSObject, NSSecureCoding {
         assert(!queries.isEmpty, "No query")
 
         let queryComponent = queries.joined(separator: ";")
-        return "directions/v5/\(profileIdentifier.rawValue)/\(queryComponent).json"
+        return "directions/v5/mapbox/\(profileIdentifier.rawValue)/\(queryComponent).json"
```
bsudekum: Could we actually pull this out into a public variable `profilePrefix`?

MapboxDirections/MBDirections.swift
```diff
@@ -229,11 +245,26 @@ open class Directions: NSObject {
      */
     @objc(URLForCalculatingDirectionsWithOptions:)
     open func url(forCalculating options: RouteOptions) -> URL {
-        let params = options.params + [
-            URLQueryItem(name: "access_token", value: accessToken),
-        ]
-        
-        let unparameterizedURL = URL(string: options.path, relativeTo: apiEndpoint)!
+        let isMapboxHost = apiEndpoint.host == mapboxHost
```
bsudekum: This is tricky. We use access tokens on other hosts that are not equal to `mapboxHost`. Perhaps we can make `mapboxHost` public so we can override this when using other mapbox hostnames.

MapboxDirections/MBDirections.swift
RomainQuidet: I use tabs for indentation in my Xcode... is that a problem ?
I moved the brackets correctly.

MapboxDirections/MBDirections.swift
RomainQuidet: As you wish. The goal here is to know when we are using mapbox host.

MapboxDirections/MBDirections.swift
RomainQuidet: I did not changed this unparameterizedURL unwrap... It is the original code.

MapboxDirections/MBRouteOptions.swift
RomainQuidet: Do I have to do it ?

MapboxDirections/MBDirections.swift
bsudekum: 👍 , tabs should be fine. Just trying to keep the library consistent.

MapboxDirections/MBRouteOptions.swift
bsudekum: @RomainQuidet yes, sometimes this is not equal to `mapbox`.

RomainQuidet:
@bsudekum Hi, how can we move forward with this pull request ? Do I have to change something ?

MapboxDirections/MBDirections.swift
```diff
@@ -128,14 +131,26 @@ open class Directions: NSObject {
...
+        if host == nil {
+            let accessToken = accessToken ?? defaultAccessToken
+            assert(accessToken != nil && !accessToken!.isEmpty, "A Mapbox access token is required. Go to <https://www.mapbox.com/studio/account/tokens/>. In Info.plist, set the MGLMapboxAccessToken key to your access token, or use the Directions(accessToken:host:) initializer.")
+
+            self.accessToken = accessToken
+			baseURLComponents = URLComponents()
+			baseURLComponents.scheme = "https"
+			baseURLComponents.host = mapboxHost
+        }
+		else {
```
bsudekum: Can you put the previous curly brace and the `else` on the same line please?

MapboxDirections/MBRouteOptions.swift
bsudekum: `mapbox` still needs to be pulled out into a public property.

MapboxDirections/MBDirections.swift
```diff
@@ -128,14 +131,26 @@ open class Directions: NSObject {
...
-        let accessToken = accessToken ?? defaultAccessToken
-        assert(accessToken != nil && !accessToken!.isEmpty, "A Mapbox access token is required. Go to <https://www.mapbox.com/studio/account/tokens/>. In Info.plist, set the MGLMapboxAccessToken key to your access token, or use the Directions(accessToken:host:) initializer.")
-        
-        self.accessToken = accessToken!
-        
-        var baseURLComponents = URLComponents()
-        baseURLComponents.scheme = "https"
-        baseURLComponents.host = host ?? "api.mapbox.com"
+		var baseURLComponents: URLComponents
+        if host == nil {
```
bsudekum: It'd be cleaner to do a guard let here.

bsudekum:
Also cc @1ec5

MapboxDirections/MBDirections.swift
RomainQuidet: I quite disagree. It's a pure if else. There's nothing to guard against.

bsudekum: You're guarding against assuming [`host`](https://github.com/mapbox/MapboxDirections.swift/pull/198/files/d2cce68548e431d48838dedc3e326fda5feec0a8#diff-e76be193f7c380a503afe0e47c1cd829R145) is not nil and [also this assert](https://github.com/mapbox/MapboxDirections.swift/pull/198/files/d2cce68548e431d48838dedc3e326fda5feec0a8#diff-e76be193f7c380a503afe0e47c1cd829R150). IMO, asserting is great time to throw in a guard.

vincethecoder: @RomainQuidet upon reviewing the code, I understand your intent. Probably it would be more clear if you safely unwrapped the `host` variable like so: 
```
if let host = host {
   var urlString = host
   if host.hasPrefix("http") == false {
      urlString = "http://\(urlString)"
   }
   let url = URL(string: urlString)
   assert(url != nil, "A valid custom host must be provided (ex: my_host.com or full http://my_host.com:8080")
   baseURLComponents = URLComponents(url: url!, resolvingAgainstBaseURL: false)!
} else {
   let accessToken = accessToken ?? defaultAccessToken
   assert(accessToken != nil && !accessToken!.isEmpty, "A Mapbox access token is required. Go to <https://www.mapbox.com/studio/account/tokens/>. In Info.plist, set the MGLMapboxAccessToken key to your access token, or use the Directions(accessToken:host:) initializer.") 
   self.accessToken = accessToken
   baseURLComponents = URLComponents()
   baseURLComponents.scheme = "https"
   baseURLComponents.host = mapboxHost
}
```

MapboxDirections/MBDirections.swift
```diff
@@ -229,11 +245,26 @@ open class Directions: NSObject {
...
-        let unparameterizedURL = URL(string: options.path, relativeTo: apiEndpoint)!
+        let isMapboxHost = apiEndpoint.host == mapboxHost
+        var params = options.params
+        if isMapboxHost {
+            params += [
+                URLQueryItem(name: "access_token", value: accessToken),
+            ]
+        }
+
+        let path: String
```
vincethecoder: shouldn't this be a `var`?

MapboxDirections/MBRouteOptions.swift
vincethecoder: @bsudekum I agree

datwelk:
What's the status of this?

MapboxDirections/MBDirections.swift
1ec5: The `access_token` parameter is unnecessary when talking to an OSRM endpoint, but it should already be easy enough to set the access token to something bogus, as the unit tests do. If the OSRM endpoint is erroring out because of that parameter, I guess we could’ve made `Credentials(accessToken:)` take an empty string as a signal to not add the `access_token` parameter to the URL. 🤷‍♂️

MapboxDirections/MBRouteOptions.swift
1ec5: This change should be unnecessary, because you can initialize a `ProfileIdentifier` to hold any arbitrary identifier using `ProfileIdentifier(rawValue:)`. If this is inconvenient, an application can always extend `ProfileIdentifier` with additional constants as desired.

1ec5 (commented):
I always felt bad about never getting around to reviewing this PR after having entertained the notion in #121. That was about when navigation went from being a side hobby project to something real inside Mapbox. Just to set the record straight about that issue, the original idea was that this library could add some facilities to process raw OSRM _responses_ as easily as Mapbox Directions API responses. As far as I knew, there wasn’t a particular need to support the OSRM _request_ format differently than Mapbox Directions API requests.

This PR seems to be focused on the request format, but it isn’t clear to me that requesting routes was ever blocked in the first place. Rather, over the years, Mapbox request and response formats have both accumulated new bits, some of which the navigation SDK has come to depend on. The most prominent part is banner and voice instructions, although the navigation SDK can kind of degrade gracefully without them.

Unfortunately, I’m not sure there’s much to be gained by leaving this PR open, since the library has been refactored so many times since.

Issue #151, opened 2017-07-12
Implement RouteStep equality


1ec5:
RouteStep currently inherits the `==` operator from NSObject, so it only considers whether two RouteStep objects are the same object. RouteStep should override the `==` operator to indicate whether the objects represent the same logical step. It should at least consider the step’s coordinates, maneuver type, and maneuver direction. Should it consider anything else about the step?

/cc @frederoni @bsudekum

1ec5:
RouteStep conforms to Equatable as of #382, though I think there was an equality method before that.

PR #200, opened 2017-11-08
Adding support for `NSCopying` protocol.

JThramer:
What it says on the box. Making `RouteObject` conform to `NSCopying`. Needed for mapbox/mapbox-navigation-ios#806.

/cc @1ec5 @bsudekum

### DIFF
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -149,7 +149,7 @@ public enum MeasurementSystem: UInt, CustomStringConvertible {
  Pass an instance of this class into the `Directions.calculate(_:completionHandler:)` method.
  */
 @objc(MBRouteOptions)
-open class RouteOptions: NSObject, NSSecureCoding {
+open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
     // MARK: Creating a Route Options Object
 
     /**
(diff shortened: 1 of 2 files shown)


MapboxDirections/MBRouteOptions.swift
```diff
@@ -458,6 +458,26 @@ open class RouteOptions: NSObject, NSSecureCoding {
         }
         return (waypoints, routes)
     }
+    
+    // MARK: NSCopying
+    /**
+     Returns a copy of the object.
+     - parameter zone: an NSZone to use in the copy operation.
+     - returns: A copy of the object.
```
1ec5: I _think_ this method already inherits documentation from the protocol’s method declaration, at least as far as Quick Help and other tools are concerned.

frederoni:
Would it be possible to avoid all this boilerplate code by adopting Codable now that we're on an Xcode 9 stack already?

JThramer:
I agree @frederoni, but wouldn't it be better to do that refactor in one tech-debt chore, rather than piecemeal? There are 8 classes that use `NSSecureCoding`, and two that use `NSCopying`, so it seems like there's enough of a payload to justify a chore here.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -476,8 +477,32 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
...
     }
+    
+    //MARK: - OBJ-C Equality
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let opts = object as? RouteOptions else { return false }
+        return isEqual(to: opts)
+    }
+    
+    @objc(isEqualToRouteOptions:)
+    open func isEqual(to routeOptions: RouteOptions?) -> Bool {
```
1ec5: Cool, this gets us a little bit closer to implementing #151. 👍

PR #209, opened 2017-11-28
Allow Directions.calculate(_:completionHandler:) result to be implictly discarded

friedbunny:
This PR makes the practice of discarding the result of `Directions.calculate(_:completionHandler:)` a bit friendlier by annotating it as `@discardableResult`. This means that the implementer does not need to explicitly discard via `_ =` or `(void)` (or face a warning when they do not).

/cc @1ec5 @frederoni @bsudekum @captainbarbosa

### DIFF
--- a/Directions Example/ViewController.m
+++ b/Directions Example/ViewController.m
@@ -36,7 +36,7 @@ NSString * const MapboxAccessToken = @"<# your Mapbox access token #>";
     MBRouteOptions *options = [[MBRouteOptions alloc] initWithWaypoints:waypoints profileIdentifier:nil];
     options.includesSteps = YES;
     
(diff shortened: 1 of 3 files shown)


1ec5:
I’m actually not entirely convinced this is a good idea. With this change, it’s much less discoverable that the method returns a task. Developers may not realize they should hang onto the task in order to cancel it before a subsequent request – this is a common need for use cases such as rerouting based on a live user location or a drag gesture.

friedbunny:
> it’s much less discoverable that the method returns a task

Hrm, I see what you mean, but I also hope that this warning wasn’t the primary clue that they should consider canceling unneeded tasks. Presumably a developer will still read the docs and/or look at the method signature, which seems like ideal places to teach/learn such things.

I think it’s also not unreasonable to argue that, for those who do read the method signature, the `@discardableResult` annotation makes it more obvious that the result is non-essential.

In any case, libraries like [Alamofire](https://github.com/Alamofire/Alamofire/search?utf8=✓&q=discardableResult&type=) and [many others](https://github.com/search?l=Swift&q=discardableResult+URLSessionDataTask&type=Code&utf8=✓) use `@discardableResult` in this exact scenario.

PR #216, opened 2017-12-01
Make Voice and Visual Instructions open

bsudekum:
This PR aims to make the classes `VisualInstruction` et al and `SpokenInstruction` open and provide public initializers to make testing easier.

Also added a few missing `@objc`.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxDirections/MBSpokenInstruction.swift
+++ b/MapboxDirections/MBSpokenInstruction.swift
@@ -8,7 +8,7 @@ import Foundation
  The `distanceAlongStep` property is measured from the beginning of the step associated with this object. By contrast, the `text` and `ssmlText` properties refer to the details in the following step. It is also possible for the instruction to refer to two following steps simultaneously when needed for safe navigation.
  */
 @objc(MBSpokenInstruction)
-public class SpokenInstruction: NSObject, NSSecureCoding {
+open class SpokenInstruction: NSObject, NSSecureCoding {
     
     /**
      A distance along the associated `RouteStep` at which to read the instruction aloud.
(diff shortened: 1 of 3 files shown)


MapboxDirections/MBSpokenInstruction.swift
```diff
@@ -33,10 +33,28 @@ public class SpokenInstruction: NSObject, NSSecureCoding {
      */
     @objc public let ssmlText: String
     
-    internal init(json: JSONDictionary) {
-        distanceAlongStep = json["distanceAlongGeometry"] as! CLLocationDistance
-        text = json["announcement"] as! String
-        ssmlText = json["ssmlAnnouncement"] as! String
+    /**
+     Initialize a `SpokenInStruction` from a dictionary.
```
1ec5: Typo: `SpokenInstruction` (×2).

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -43,11 +49,22 @@ public class VisualInstructionComponent: NSObject, NSSecureCoding {
...
+        
+        self.init(text: text, imageURL: imageURL)
     }
     
-    public required init?(coder decoder: NSCoder) {
+    /**
+     :nodoc:
+     Initialize A `VisualInstructionComponent`.
+     */
+    @objc public init(text: String?, imageURL: URL?) {
```
1ec5: In the future, as we add additional fields to the component object, if we find that the number of arguments gets unwieldy, we can make the properties `var` instead and leave them out of the initializer.

Issue #217, opened 2017-12-02
RouteOptions.response(from:) not adding waypoint names from JSON response.


JThramer:
There is a bug in https://github.com/mapbox/MapboxDirections.swift/blob/master/MapboxDirections/MBRouteOptions.swift#L480 . Although L486 _is correct logic_, the result ends up being basically thrown away, because `Route.initWithJSON:waypoints:routeOptions` doesn’t insert those `namedWaypoints` into it’s model, it just uses them to build `RouteLeg`s. I think the simple fix here is to insert `self.waypoints = namedWaypoints` between Line 490 and 491 in `MBRouteOptions.swift`.

PR #218, opened 2017-12-02
217 OSRM waypoint names

JThramer:
Implements #217. This was a fun little Friday-afternoon mystery.

/cc @1ec5 @bsudekum

### DIFF
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -485,13 +485,15 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
                 let coordinate = CLLocationCoordinate2D(geoJSON: location)
                 return Waypoint(coordinate: coordinate, name: waypoint.name ?? json["name"] as? String)
             }
+            self.waypoints = namedWaypoints
         } else {
             namedWaypoints = self.waypoints
         }
+        
         let routes = (json["routes"] as? [JSONDictionary])?.map {
             Route(json: $0, waypoints: namedWaypoints, routeOptions: self)
         }
-        return (waypoints, routes)
+        return (namedWaypoints, routes)
     }
     
     // MARK: NSCopying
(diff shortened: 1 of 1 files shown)


MapboxDirections/MBRouteOptions.swift
```diff
@@ -485,13 +485,15 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
                 let coordinate = CLLocationCoordinate2D(geoJSON: location)
                 return Waypoint(coordinate: coordinate, name: waypoint.name ?? json["name"] as? String)
             }
+            self.waypoints = namedWaypoints
```
1ec5: I’m not sure that receiving a response should have the side effect of changing the original RouteOptions request. Consider that the developer might intentionally avoid naming one waypoint, but then it gets filled in with OSRM’s name for the waypoint. The next time the RouteOptions object gets reused, the name becomes explicit.

It seems like the key to this fix is the other change, which ensures that the returned array of waypoints is named.

JThramer: - [x] Don't update request model with response values.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -478,20 +478,21 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
...
+                let location = api["location"] as! [Double]
                 let coordinate = CLLocationCoordinate2D(geoJSON: location)
-                return Waypoint(coordinate: coordinate, name: waypoint.name ?? json["name"] as? String)
+                return Waypoint(coordinate: coordinate, name: local.name ?? api["name"] as? String)
             }
-        } else {
-            namedWaypoints = self.waypoints
         }
+        
+        let wpts = namedWaypoints ?? waypoints
```
1ec5: FYI, we try not to overly abbreviate like this in the codebase. In this case, we could instead inline the variable in the two lines below, or we could’ve overridden namedWaypoints and used that variable below.

PR #219, opened 2017-12-04
Add roadSide to RouteStep

### LINKED ISSUE #212: Add RouteStep property indicating a side of the road
`RouteStep` should have a `roadSide` property indicating the side of the road to use while traversing the step, otherwise known as the “[rule of the road](https://en.wikipedia.org/wiki/Right-_and_left-hand_traffic)”. The Directions API response includes a `driving_side` property with this information. Despite the name, it’s present in the cycling and walking profiles as well.

/cc @bsudekum @danpat

frederoni:
Fixes #212 - Add roadSide to RouteStep

Unblocks https://github.com/mapbox/mapbox-navigation-ios/issues/909

@bsudekum @1ec5 👀


MapboxDirections/MBRouteStep.swift
```diff
@@ -421,6 +421,35 @@ public enum ManeuverDirection: Int, CustomStringConvertible {
     }
 }
 
+@objc(MBRoadSide)
+public enum RoadSide: Int, CustomStringConvertible {
+    case left
+    case right
```
bsudekum: Although this is pretty self explanatory, can you document this class it's properties?

MapboxDirections/MBRouteStep.swift
```diff
@@ -785,6 +824,11 @@ open class RouteStep: NSObject, NSSecureCoding {
      */
     open let maneuverDirection: ManeuverDirection?
     
+    /**
+     Indicates what side of a bidirectional road the driver must be driving on. Also referred to as the rule of the road.
+     */
+    open let roadSide: RoadSide?
```
bsudekum: This property is always returned, I think we can make this non-optional. Noting, it is in fact returned for biking/walking.

1ec5: Making it optional will also allow it to bridge to Objective-C.

MapboxDirections/MBRouteStep.swift
```diff
@@ -557,6 +597,7 @@ open class RouteStep: NSObject, NSSecureCoding {
         let maneuverType = ManeuverType(description: maneuver["type"] as! String)
         let maneuverDirection = ManeuverDirection(description: maneuver["modifier"] as? String ?? "")
         let maneuverLocation = CLLocationCoordinate2D(geoJSON: maneuver["location"] as! [Double])
+        let roadSide = RoadSide(description: json["driving_side"] as! String) ?? .right
```
bsudekum: Unsure how to handle a `nil` here.

1ec5:
Regarding the property’s name, `roadSide` is potentially a little misleading for the walking and cycling profiles, because it doesn’t indicate the side of the road that the sidewalk or bike lane is on. On the other hand, `drivingSide` would also be misleading because it does indicate the side of a bike trail to use. I don’t think either choice is a huge problem, but here are some more options:

* `ruleOfRoad`
* `handedness`
* `sideToKeep` (as in “Keep left/right”, as opposed to “overtaking side”)

MapboxDirections/MBRouteStep.swift
```diff
@@ -602,6 +643,11 @@ open class RouteStep: NSObject, NSSecureCoding {
             maneuverDirection = nil
         }
         
+        guard let roadSideDescription = decoder.decodeObject(of: NSString.self, forKey: "roadSide") as String?, let roadSide = RoadSide(description: roadSideDescription) else {
+            return nil
+        }
+        self.roadSide = roadSide
```
bsudekum: Will this be an issue for old routes that are encoded without a `roadSide`?

MapboxDirections/MBRouteStep.swift
1ec5: Is it possible for any production profile to not provide a `driving_side` property? If so, we would add an `unknown` enumeration value. If not, we would assert or implicitly unwrap the optional.

friedbunny:
> Regarding the property’s name, `roadSide` is potentially a little misleading for the walking and cycling profiles, because it doesn’t indicate the side of the road that the sidewalk or bike lane is on.

`sideOfTravel`/`travelSide`?

bsudekum:
@1ec5 I agree. To riff on this a little more, it's important to note that it might make sense to incorporate the word `car` or `driving` into the property name. Your suggestions might be confusing for pedestrians though as `sideToKeep` sounds like pedestrians must keep to this side.

1ec5:
I think that suffers from the same problem, in that it implies something about the current cycleway’s geometry relative to a road. `trafficSide` would be less problematic in this regard.

The more I think about this, the more comfortable I am with calling it `drivingSide` as the API does: the rule of the road for bike paths and bike lanes is generally based on the one for vehicular traffic. When it differs, as with reversed bike lanes, we would need some other field to make that clear anyways.

In the end, I think developers will have to just read the documentation to understand this property, unless we call it something silly like `steeringWheelSide` (which isn’t necessarily accurate either).

bsudekum:
👍  to `drivingSide`.

MapboxDirections/MBRouteStep.swift
1ec5: Sorry, I meant non-optional. It should be non-optional.

bsudekum:

frederoni:
✅  LGTM but I can't approve.

MapboxDirections/MBRouteStep.swift
1ec5: Yes, let’s assume it’s `right` if unspecified. That would match OSRM’s behavior prior to Project-OSRM/osrm-backend#4415.

/cc @danpat

MapboxDirections/MBRouteStep.swift
```diff
@@ -643,10 +643,10 @@ open class RouteStep: NSObject, NSSecureCoding {
             maneuverDirection = nil
         }
         
-        guard let roadSideDescription = decoder.decodeObject(of: NSString.self, forKey: "roadSide") as String?, let roadSide = RoadSide(description: roadSideDescription) else {
+        guard let drivingSideDescription = decoder.decodeObject(of: NSString.self, forKey: "drivingSide") as String?, let drivingSide = DrivingSide(description: drivingSideDescription) else {
```
1ec5: Let’s default this value to `right` if it’s unspecified, so that preexisting objects don’t suddenly fail to decode.

PR #221, opened 2017-12-07
Adopt Codable and JSONDecoder

### LINKED ISSUE #134: Adopt JSONDecoder
Swift 4 introduces JSONDecoder and built-in coding protocols, which should allow us to remove the majority of code from this library.

/cc @bsudekum @frederoni

frederoni:
Fixes #134 

- [x] Codable
- [x] JSONDecoder
- [x] Fix outletIndexes
- [x] Fix visual instruction imageBaseURL
- [x] Re-enable route options tests

@bsudekum @1ec5 👀


frederoni:
> Is there anything we can do to make this extension fit in more naturally?

This got replaced by a `UncertainCodable<Geometry, String>`.

frederoni:
V4 compatibility was pretty isolated before. I wonder if I made that part worse. 🤔

bsudekum:
Before merging, let's double check this integrates properly with another library that uses this as a dependency.

MapboxDirections/MBDirections.swift
```diff
@@ -192,32 +162,18 @@ open class Directions: NSObject {
...
-            
-            let apiStatusCode = json["code"] as? String
-            let apiMessage = json["message"] as? String
-            guard data != nil && error == nil && ((apiStatusCode == nil && apiMessage == nil) || apiStatusCode == "Ok") else {
-                let apiError = Directions.informativeError(describing: json, response: response, underlyingError: error as NSError?)
-                DispatchQueue.main.async {
-                    errorHandler(apiError)
-                }
+            guard let data = data, response?.mimeType == "application/json" else {
+                errorHandler(Directions.informativeError(describing: [:], response: response, underlyingError: error as NSError?))
```
1ec5: This call used to take a JSON object whose `code` and `message` keys determined the message. Does that no longer matter?

frederoni: Good catch. Errors are handled more gracefully in https://github.com/mapbox/MapboxDirections.swift/pull/221/commits/1a161d6fc7694f2e2dc7597baaae6e3cdc9da34e

frederoni:
I haven't noticed any changes or regressions in mapbox-navigation-ios after these changes but let's hold off a bit more. V5Tests.testGeoJSON frequently times out on CI with a timeout set to 2 seconds.

Before Codable, testGeoJSON consistently finishes at 750ms compared to 1.7s after adopting Codable. It's a 3.2MB JSON but still.

frederoni:
Managed to get it down to 1.3s but we may want to wait for https://bugs.swift.org/browse/SR-6252

frederoni:
The bottleneck persists in Swift 4.2

1ec5:
Superseded by #382.

PR #222, opened 2017-12-07
Mapbox Directions API OSRM Name Fix

JThramer:
This mostly fixes issue https://github.com/mapbox/mapbox-navigation-ios/issues/921, where an empty string is bypassing the `CLGeocoder` and causing the coordinate logic to trigger (which does test for empty string -- a bug, but moot because it's being removed anyway). This change fixes this issue and makes sure that `waypoint.name == nil` when the Mapbox Directions API does not return an OSRM name for the waypoint.

*  Fixing issue where querying a nonexistent `waypoints[n].name` keypath in the result `json` will produce `""` instead of the expected `nil`. Adding logic to compensate for this.

ToDo: 
- [x] Add Test that covers `MBRouteOptions.response(from:)` and tests this edge-case
- [x] `isNotEmpty` / `nonEmptyString` extension and usage

/cc @bsudekum @1ec5 @frederoni


bsudekum:
Would it be possible to add a test case?

MapboxDirections/MBRouteOptions.swift
```diff
@@ -480,19 +480,21 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
     internal func response(from json: JSONDictionary) -> ([Waypoint]?, [Route]?) {
         var namedWaypoints: [Waypoint]?
         if let jsonWaypoints = (json["waypoints"] as? [JSONDictionary]) {
-            namedWaypoints = zip(jsonWaypoints, waypoints).map { (api, local) -> Waypoint in
+            namedWaypoints = zip(jsonWaypoints, self.waypoints).map { (api, local) -> Waypoint in
                 let location = api["location"] as! [Double]
                 let coordinate = CLLocationCoordinate2D(geoJSON: location)
-                return Waypoint(coordinate: coordinate, name: local.name ?? api["name"] as? String)
+                let possibleAPIName = api["name"] as? String
+                let apiName = (possibleAPIName?.isEmpty ?? true) ? nil : possibleAPIName!
```
1ec5: Core Navigation [extends String with `nonEmptyString`](https://github.com/mapbox/mapbox-navigation-ios/blob/9047a41dca57f99a08e3452a9a114dcec52f743b/MapboxCoreNavigation/String.swift#L11-L13), which is pretty convenient for situations like this. Feel free to copy it here.

JThramer:
I think #217 should be closed as well after this is merged and verified.

bsudekum:
@JThramer mind checking out the test failure:

```
❌  /Users/vagrant/git/MapboxDirections/MBRouteOptions.swift:487:46: value of type 'String' has no member 'nonEmptyString'
                let apiName = possibleAPIName?.nonEmptyString
                              ~~~~~~~~~~~~~~~^ ~~~~~~~~~~~~~~
```

MapboxDirections/Extensions/String.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+extension String {
+    var isNotEmpty: Bool { return !isEmpty }
```
1ec5: This is probably a bit extreme. For one thing, it’s more verbose than using the `!` operator, which is more versatile than hard-coding negative versions of Boolean properties.

PR #224, opened 2017-12-11
Add isDelimiter to VisualInstructionComponent

bsudekum:
Adds the bool `isDelimiter` to `VisualInstructionComponent` indicating that the component should be visualized as a delimiter in the UI.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxDirections/MBVisualInstructionComponent.swift
+++ b/MapboxDirections/MBVisualInstructionComponent.swift
@@ -45,7 +45,7 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
     @objc public convenience init(json: [String: Any]) {
         let text = json["text"] as? String
         var type: VisualInstructionComponentType?
-        if delimiter = json["delimiter"] as? Bool {
+        if let _ = json["delimiter"] as? Bool {
             type = .delimiter
         } else {
             type = .destination
--- a/MapboxDirections/MBVisualInstructionComponent.swift
+++ b/MapboxDirections/MBVisualInstructionComponent.swift
@@ -34,9 +34,9 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
     
     /**
      :nodoc:
-     `Bool` whether the `VisualInstructionComponent` should be visualized as a delimiter between two text components.
+     The type of the `VisualInstructionComponent`. Used for influencing how the compomenent is rendered.
      */
-    @objc public var type: VisualCompomentType
+    @objc public var type: VisualInstructionComponentType
     
     /**
      :nodoc:
@@ -44,8 +44,8 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      */
     @objc public convenience init(json: [String: Any]) {
         let text = json["text"] as? String
-        var type: VisualCompomentType?
-        if let delimiter = json["delimiter"] as? String, VisualCompomentType(description: delimiter) == .delimiter {
+        var type: VisualInstructionComponentType?
+        if delimiter = json["delimiter"] as? Bool {
             type = .delimiter
         } else {
             type = .destination
@@ -71,7 +71,7 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      :nodoc:
      Initialize A `VisualInstructionComponent`.
      */
-    @objc public init(text: String?, imageURL: URL?, type: VisualCompomentType) {
+    @objc public init(text: String?, imageURL: URL?, type: VisualInstructionComponentType) {
         self.text = text
         self.imageURL = imageURL
         self.type = type
@@ -88,7 +88,7 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
         }
         self.imageURL = imageURL
         
-        guard let typeString = decoder.decodeObject(of: NSString.self, forKey: "type") as String?, let type = VisualCompomentType(description: typeString) else {
+        guard let typeString = decoder.decodeObject(of: NSString.self, forKey: "type") as String?, let type = VisualInstructionComponentType(description: typeString) else {
                 return nil
         }
         self.type = type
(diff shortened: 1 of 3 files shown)


bsudekum:
Alternatively, we could start to name component types as enums: `.cardinalDirection`, `.streetSuffix`, `.delimiter`, etc.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -52,16 +60,17 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
-        self.init(text: text, imageURL: imageURL)
+        self.init(text: text, imageURL: imageURL, isDelimiter: isDelimiter)
     }
     
     /**
      :nodoc:
      Initialize A `VisualInstructionComponent`.
      */
-    @objc public init(text: String?, imageURL: URL?) {
+    @objc public init(text: String?, imageURL: URL?, isDelimiter: Bool) {
```
1ec5: To avoid argument proliferation, let’s leave this argument out and set the `isDelimiter` property to `false` by default. It’s variable, so the caller can set it to `true` right after creating the object.

bsudekum:
@1ec5 updated this to use `type` instead.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -84,14 +88,57 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
+        coder.encode(type, forKey: "type")
+    }
+}
+
+
+/**
+ `VisualCompomentType` describes the `text` type.
+ */
+@objc(MBVisualComponentType)
+public enum VisualCompomentType: Int, CustomStringConvertible {
```
1ec5: Typo: `Component`.

1ec5: This enumeration is a little difficult to find here. Consider moving it to the top of the file, where enumerations and other non-class types are typically declared, or to a separate file.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -36,18 +36,22 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
     
     /**
      :nodoc:
      Initialize A `VisualInstructionComponent`.
      */
     @objc public convenience init(json: [String: Any]) {
         let text = json["text"] as? String
-        let isDelimiter = json["delimiter"] as? Bool ?? false
+        var type: VisualCompomentType?
+        if let delimiter = json["delimiter"] as? String, VisualCompomentType(description: delimiter) == .delimiter {
```
1ec5: In the response JSON, `delimiter` is either set to `true` or left undefined. It isn’t ever set to the string `delimiter`, as this code would suggest.

bsudekum:

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -34,20 +34,24 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
     
     /**
      :nodoc:
-     `Bool` whether the `VisualInstructionComponent` should be visualized as a delimiter between two text components.
+     The type of the `VisualInstructionComponent`. Used for influencing how the compomenent is rendered.
```
1ec5: Typo: component. Also, it isn’t necessary to link to `VisualInstructionComponent`, since it’s the containing class:

> The type of visual instruction component. You can display the component differently depending on its type.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -60,17 +64,17 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
-        self.init(text: text, imageURL: imageURL, isDelimiter: isDelimiter)
+        self.init(text: text, imageURL: imageURL, type: type!)
     }
     
     /**
      :nodoc:
      Initialize A `VisualInstructionComponent`.
      */
-    @objc public init(text: String?, imageURL: URL?, isDelimiter: Bool) {
+    @objc public init(text: String?, imageURL: URL?, type: VisualInstructionComponentType) {
```
1ec5: It feels more natural to put `type`, a non-optional argument, ahead of `text` and `imageURL`, which may be `nil`.

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+/**
+ `VisualInstructionComponentType` describes the type of `VisualInstructionComponent`.
+ */
+@objc(MBVisualInstructionComponentType)
+public enum VisualInstructionComponentType: Int, CustomStringConvertible {
+    
+    /**
+     Text is a delimiter.
```
1ec5: > The component separates two other destination components.
> 
> If the two adjacent components are both displayed as images, you can hide this delimiter component.

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -0,0 +1,40 @@
...
+     */
+    case delimiter
+    
+    /**
+     Text is a way name.
```
1ec5: A destination isn’t necessarily the name of a street:

> The component bears an exit number or the name of a place or street.

Come to think of it, it would be nice to be able to distinguish exit numbers from place or street names. 🤔

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -0,0 +1,40 @@
...
+     */
+    case destination
+    
+    public init?(description: String) {
+        let level: VisualInstructionComponentType
```
1ec5: This isn’t a level.

MapboxDirections/MBVisualInstructionType.swift
bsudekum: I agree, but more changes will need to be made on the server.

1ec5: In the meantime, I think this description is accurate:

> The component bears an exit number or the name of a place or street.

PR #226, opened 2017-12-18
[Collab] Swift 4 Coder fix for fred

JThramer:
Hey @frederoni, I Didn't want to commit right into #221 without running this by you first.

The trick here seems to be overriding the `Decodable` initializer and manually setting up the model after decoding the JSON.

*  Adding custom init that sets leg destination names properly now. All tests now pass.


frederoni:
Thanks, @JThramer. I've cherry-picked your commits into #221

PR #227, opened 2018-01-02
Expose more RouteStep properties to Obj-C

### LINKED ISSUE #117: RouteStep’s maneuverType, maneuverDirection, transportType properties don’t bridge to Objective-C
RouteStep’s `maneuverType`, `maneuverDirection`, and `transportType` properties don’t bridge to Objective-C, because it isn’t possible for an enumeration-typed value to be nullable in Objective-C. Instead of making these properties optional, we’d need to add `.none` cases to the ManeuverType, ManeuverDirection, and TransportType enumerations. The tradeoff is that Swift developers would be unable to use optional chaining syntax with these properties; instead, they’d have to explicitly check for `.none`. (Yuck!)

/cc @bsudekum @cammace

frederoni:
Fixes #117 

`maneuverType`, `maneuverDirection`, and `transportType` now bridges to Obj-C.

Another way to expose these properties to Obj-C could be to add an additional computed non-optional `bridgableManeuverType` if optional chaining syntax is preferred.

<details>

```swift
@objc open var bridgableManeuverType: ManeuverType {
    get {
        return maneuverType ?? .none
    }
}
```

</details>
<br>
@1ec5 @bsudekum 👀

### DIFF
--- a/MapboxDirections/MBRouteStep.swift
+++ b/MapboxDirections/MBRouteStep.swift
@@ -342,6 +342,13 @@ public enum ManeuverType: Int, CustomStringConvertible {
 @objc(MBManeuverDirection)
 public enum ManeuverDirection: Int, CustomStringConvertible {
     /**
+     The step does not have a particular maneuver direction associated with it.
+     
+     This maneuver direction is used as a workaround for bridging to Objective-C which does not support nullable enumeration-typed values.
+     */
+    case none
+    
+    /**
      The maneuver requires a sharp turn to the right.
      */
     case sharpRight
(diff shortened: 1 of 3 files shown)


MapboxDirections/MBRouteStep.swift
```diff
@@ -669,7 +696,7 @@ open class RouteStep: NSObject, NSSecureCoding {
         guard let transportTypeDescription = decoder.decodeObject(of: NSString.self, forKey: "transportType") as String? else {
             return nil
         }
-        transportType = TransportType(description: transportTypeDescription)
+        transportType = TransportType(description: transportTypeDescription)!
```
bsudekum: This should probably be safely unwrapped.

bsudekum:
Looks good. It looks like in a few places we're asserting that values are non-optional where as code around either unwraps the value or provides a default fallback.

Niduank:
Hi,

When will you release a new MapboxDirections.swift pod which include bridge to ObjC for RouteStep.maneuverX variables ?

Cheers

PR #235, opened 2018-01-22
Add spoken locale

bsudekum:
The Mapbox Directions API will soon tell the developer in which locale to speak the instructions in. This is helpful for in situations where there is a mismatch between the written instructions the API supports and the spoken voice locales as well.

Looks like this is failing with the error

```
XCTAssertEqual failed: ("en-US (fixed)") is not equal to ("en-US")
```

Any ideas @1ec5 about this? The test fixture addition adds `voiceLocale: "en-us"`. Is this a case issue?

/cc @mapbox/navigation-ios


MapboxDirections/MBRoute.swift
```diff
@@ -194,6 +197,11 @@ open class Route: NSObject, NSSecureCoding {
      Each route produced by a single call to `Directions.calculate(_:completionHandler:)` has the same route identifier.
      */
     @objc open var routeIdentifier: String?
+    
+    /**
+     The locale to use for spoken instructions.
+     */
+    @objc open var spokenLocale: Locale?
```
1ec5: Rename this property to `speechLocale` or `voiceLocale` to better emphasize that this locale gets plugged into MapboxSpeech or the Voice API.

MapboxDirections/MBDirections.swift
```diff
@@ -173,6 +173,9 @@ open class Directions: NSObject {
                     route.accessToken = self.accessToken
                     route.apiEndpoint = self.apiEndpoint
                     route.routeIdentifier = json["uuid"] as? String
+                    if let vLocale = json["voiceLocale"] as? String {
+                        route.spokenLocale = Locale(identifier: vLocale)
```
1ec5: What do you think about also copying this locale into each SpokenInstruction, so that the navigation SDK’s voice controller classes only need to work with individual SpokenInstruction objects?

MapboxDirections/MBRoute.swift
```diff
@@ -194,6 +197,11 @@ open class Route: NSObject, NSSecureCoding {
      Each route produced by a single call to `Directions.calculate(_:completionHandler:)` has the same route identifier.
      */
     @objc open var routeIdentifier: String?
+    
+    /**
+     The locale to use for spoken instructions.
```
1ec5: Explain that this locale is specifically meant for use with the Mapbox Voice API. Add a note about what `nil` means.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -58,6 +58,7 @@ class V5Tests: XCTestCase {
         XCTAssertEqual(route!.accessToken, BogusToken)
         XCTAssertEqual(route!.apiEndpoint, URL(string: "https://api.mapbox.com"))
         XCTAssertEqual(route!.routeIdentifier, "cj725hpi30yp2ztm2ehbcipmh")
+        XCTAssertEqual(route!.spokenLocale!.description, "en-US")
```
1ec5: `description` is a free-form, human-readable string that may contain extra details beyond the identifier. Use `identifier` instead.

MapboxDirections/MBDirections.swift
bsudekum: Yeah that works

bsudekum:
This is ready for review @mapbox/navigation-ios.

Issue #158, opened 2017-07-25
Map Matching API


1ec5:
Given the [Mapbox Map Matching API](https://www.mapbox.com/api-documentation/#retrieve-a-match)’s similarity to the Directions API, we could pretty cleanly implement support for the former in this library. Here’s one possible design:

* The Directions class would gain a new `match(_:completionHandler:)` method that takes a MatchOptions object based on [these options](https://www.mapbox.com/api-documentation/#retrieve-a-match).
* The completion handler would accept an array of [Match](https://www.mapbox.com/api-documentation/#match-object) objects, an array of [Tracepoint](https://www.mapbox.com/api-documentation/#tracepoint-object) objects, and an optional Error.
* Match would be a subclass of Route.
* Tracepoint would be a subclass of Waypoint.

/cc @bsudekum

ericrwolfe:
This will be unblocked by https://github.com/Project-OSRM/osrm-backend/issues/4669.

1ec5:
It sounds like Project-OSRM/osrm-backend#4669 blocks a particular use case for this feature, but there’s nothing stopping us from implementing Map Matching API support for general use cases, as Mapbox Java Services has.

ericrwolfe:
> but there’s nothing stopping us from implementing Map Matching API support for general use cases, as Mapbox Java Services has.

Indeed. We can go ahead and implement this whenever we have the bandwidth, but once https://github.com/Project-OSRM/osrm-backend/issues/4669 lands it'll make testing this use case a lot easier.

ericrwolfe:
A new `waypoints` parameter (https://github.com/Project-OSRM/osrm-backend/pull/4732) was added to the map matching API which makes this usable for turn-by-turn navigation.

https://github.com/mapbox/mapbox-navigation-android/issues/659 describes similar changes that need to be added to support map matching and custom routes in Mapbox Java and Mapbox Navigation on Android.

PR #236, opened 2018-01-25
Add map matching

bsudekum:
closes: https://github.com/mapbox/MapboxDirections.swift/issues/158

This PR adds support for the Mapbox [Map Matching API](https://www.mapbox.com/api-documentation/#map-matching). 

todo:
- [ ] document
- [x] test
- [x] make demo app work with both matching and directions.

/cc @mapbox/navigation-ios


MapboxDirections/MBDirections.swift
```diff
@@ -183,6 +183,26 @@ open class Directions: NSObject {
         return task
     }
     
+    @objc(calculateMatchingWithOptions:completionHandler:)
+    @discardableResult open func calculate(_ options: MatchOptions, completionHandler: @escaping CompletionHandler) -> URLSessionDataTask {
```
1ec5: “Calculate matching” sounds a bit contrived. The existing `calculate(_:completionHandler:)` is so named because “route” can easily be confused with a noun. I think we can call this method `match(_:completionHandler:)` in Swift and `-matchWithOptions:completionHandler:` in Objective-C, but I don’t have a strong opinion on this.

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,82 @@
+import Foundation
+
+@objc(MBMatchOptions)
+open class MatchOptions: RouteOptions {
```
1ec5: MatchOptions sounds even more like a method name than RouteOptions. Should we call it MatchingOptions, or is it more important to keep the two class names parallel?

MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+@objc(MBTracePoint)
+public class TracePoint: Waypoint {
```
1ec5: The “point” in Waypoint is lowercase, so this class should be named Tracepoint.

MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -0,0 +1,23 @@
...
+    
+    @objc open var alternateCount: Int
+    
+    @objc open var waypointIndex: Int
```
1ec5: Would it be more convenient for the Tracepoint object to maintain an index to a waypoint or a weak pointer to the Waypoint object itself?

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,82 @@
...
+    public init(tracePoints: [TracePoint], profileIdentifier: MBDirectionsProfileIdentifier?, resample: Bool?, timestamps: [Date]?) {
+        super.init(waypoints: tracePoints, profileIdentifier: profileIdentifier)
+    }
+    
+    @objc open var resample: Bool = false
```
1ec5: “Resample” sounds like a method name. Rename this to `resamples<some noun>` or `tidies<some noun>`.

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,82 @@
...
+    
+    @objc open var timestamps: [Date]?
```
1ec5: Maintaining parallel arrays is prone to mismatches. Add a `Date`-typed `timestamp` property to Tracepoint or have the initializer take an array of `CLLocation`s (which has a `timestamp` property). Also consider adding a convenience initializer that takes an array of `CLLocationCoordinate2D`s, for when the developer doesn’t have timestamps to provide.

An open question is what to do if some tracepoints have timestamps and others do not. That can be a runtime check when computing a request URL.

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,82 @@
+import Foundation
+
+@objc(MBMatchOptions)
+open class MatchOptions: RouteOptions {
+    public init(tracePoints: [TracePoint], profileIdentifier: MBDirectionsProfileIdentifier?, resample: Bool?, timestamps: [Date]?) {
```
bsudekum: @1ec5 do you think it makes sense to init this with tracePoints? TracePoints seem more like the result from a matching request.

1ec5: I think CLLocation would be a better choice, because a few of the properties on Tracepoint don’t make sense as part of a request.

MapboxDirections/MBDirections.swift
```diff
@@ -105,6 +105,8 @@ open class Directions: NSObject {
      */
     public typealias CompletionHandler = (_ waypoints: [Waypoint]?, _ routes: [Route]?, _ error: NSError?) -> Void
```
1ec5: Let’s rename this type alias to `DirectionsCompletionHandler`.

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,86 @@
...
+        let waypoints = locations.map {
+            Waypoint(location: $0)
+        }
+        self.timestamps = locations.map { $0.timestamp }
+        super.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
+    }
+    
+    @objc open var resample: Bool = false
+    
+    @objc open var timestamps: [Date]?
```
1ec5: Now that the timestamps are specified on the CLLocations in the input, make this property private.

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,86 @@
...
+    public required init?(coder decoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override internal var params: [URLQueryItem] {
+        var params: [URLQueryItem] = [
+            URLQueryItem(name: "geometries", value: String(describing: shapeFormat)),
+            URLQueryItem(name: "overview", value: String(describing: routeShapeResolution)),
+            URLQueryItem(name: "steps", value: String(includesSteps)),
+            URLQueryItem(name: "language", value: locale.identifier),
```
1ec5: Most of these parameters are copied from the superclass. Instead of repeating the code here, get the superclass’s `params` and add to them in this implementation.

MapboxDirections/Match/MBMatchOptions.swift
bsudekum: MatchingOptions workings

MapboxDirections/Match/MBTracepoint.swift
bsudekum: I think for 100% consistence sake, we should define this property with what the API is giving us.

1ec5: Strict consistency with the Directions API hasn’t ever been a high priority for this library. At least, consistency with the API has always been subordinate to intuitiveness. The API suffers from a lot of technical debt for backwards compatibility reasons that don’t apply to this library. Think about how the developer would need to use this library and whether there’s any boilerplate we can take off the developer’s hands.

In this particular case, the API uses indices because JSON doesn’t have a concept of [object references](https://en.wikipedia.org/wiki/JSON#Object_references). Just as this library already pairs steps with their origin and destination waypoints, I’m pretty sure a developer would expect us to pair tracepoints with waypoints, rather than make them `zip` the objects themselves.

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,68 @@
...
+                let waypointIndex = api["waypoint_index"] as! Int
+                let matchingIndex = api["matchings_index"] as! Int
+                let name = api["name"] as? String
+                // TODO: add name
+                // let possibleAPIName = api["name"] as? String
+                return Tracepoint(coordinate: coordinate, alternateCount: alternateCount, waypointIndex: waypointIndex, matchingIndex: matchingIndex, name: name)
+            }
+        }
+        
+        let tracePoints = namedTracepoints!
```
bsudekum: @1ec5 On the `Waypoints` side of things, we fall back to [`self.waypoints`](https://github.com/mapbox/MapboxDirections.swift/blob/8ea5ad5f16f270c852985789156e302691cb5a35/MapboxDirections/MBRouteOptions.swift#L492) which doesn't make much sense here. 

Should we have a `tracePoints` property on `MatchingOptions` as well? This would introduce some complexity/refactoring I think.

1ec5: namedTracepoints shouldn’t be nil coming out the response, right? So we shouldn’t have to worry about falling back to self.waypoint, nor is a tracepoints property needed.

bsudekum:
@1ec5 this is ready for another round of review.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -160,7 +160,6 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
      */
     @objc public init(waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = nil) {
         assert(waypoints.count >= 2, "A route requires at least a source and destination.")
-        assert(waypoints.count <= 25, "A route may not have more than 25 waypoints.")
```
bsudekum: This assertion throws for map matching now. Can we let the server handle this error?

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,113 @@
...
+        coder.encode(resampleTraces, forKey: "resampleTraces")
+        coder.encode(timestamps, forKey: "timestamps")
+        coder.encode(waypointIndices, forKey: "waypointIndices")
+    }
+    
+    @objc private var paramsThatOverlap: [String] = ["geometries", "radiuses", "steps", "overview", "timestamps", "annotations"]
+    
+    override internal var params: [URLQueryItem] {
+        var params = super.params.filter {
+            paramsThatOverlap.contains($0.name)
```
bsudekum: Slightly annoying but if we do not remove these query params from `super.params`, the server errors.

bsudekum: Noting, the removed params are still exposed on MatchOptions. If a developer sets one of the unused options, it will just a no-op.

bsudekum:
new3

It's starting to work!

MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -0,0 +1,41 @@
...
+    
+    /**
+     Number of probable alternative matchings for this trace point. A value of zero indicates that this point was matched unambiguously.
+     */
+    @objc open var alternateCount: Int
+    
+    /**
+     Index of the waypoint inside the matched route.
+     */
+    open var waypointIndex: Int?
```
bsudekum: Sadly, this can be nil. @1ec5 how can we make this bridge to objc?

MapboxDirections/MBRouteOptions.swift
1ec5: This is a strong sign that RouteOptions and MatchOptions need to be siblings under either an abstract superclass or a protocol. For now, though, let’s factor the assertions out into a private preflight method that gets called from this initializer. The MatchOptions initializer can override that method with the assertions that make sense for that API.

MapboxDirections/Match/MBMatchOptions.swift
1ec5: This is another good case for having RouteOptions and MatchOptions be siblings. Conceptually, it only makes sense for MatchOptions to inherit from RouteOptions if RouteOptions is a strict subset of MatchOptions.

Making RouteOptions and MatchOptions into siblings isn’t as scary as it sounds:

1. Rename RouteOptions to DirectionsOptions.
1. Make MatchOptions a subclass of DirectionsOptions.
1. Create a new RouteOptions that looks a lot like MatchOptions, except with routing-specific options (essentially anything that doesn’t overlap with MatchOptions).
1. Consider turning DirectionsOptions into a protocol with a default implementation.

MapboxDirections/Match/MBTracepoint.swift
1ec5: How will developers tend to encounter this property? When a developer finds themselves needing to associate a tracepoint with a route, do they already have a route on hand? If so, consider making the route hold a reference to a tracepoint somehow.

(`NSNotFound` represents a nonexistent index.)

Directions Example/DrawingView.swift
```diff
@@ -0,0 +1,101 @@
...
+        
+        context.beginPath()
+        context.move(to: firstPoint)
+    }
+    
+    override public func touchesMoved(_ touches: Set<UITouch>, with event: UIEvent?) {
+        guard let lastPoint = points.last else { return }
+        context.move(to: lastPoint)
+        
+        let point = touches.first!.location(in: self)
```
vincethecoder: @bsudekum should we also guard the touch here like so:
```
guard let touch = touches.first else { return }
```

Directions Example/DrawingView.swift
```diff
@@ -0,0 +1,101 @@
...
+        UIGraphicsBeginImageContextWithOptions(bounds.size, false, UIScreen.main.scale)
+        context = UIGraphicsGetCurrentContext()!
+        setStrokeColor(strokeColor: self.strokeColor)
+        setLineWidth(lineWidth: self.lineWidth)
+    }
+    
+    override public func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        points.removeAll(keepingCapacity: false)
+        
+        let firstPoint = touches.first!.location(in: self)
```
vincethecoder: @bsudekum could we safely unwrap the touches here? 
```
guard let touch = touches.first else { return }
```

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,454 @@
...
+    }
+    
+    /**
+     An array of directions query strings to include in the request URL.
+     */
+    internal var queries: [String] {
+        return waypoints.map { "\($0.coordinate.longitude),\($0.coordinate.latitude)" }
+    }
+    
+    internal var path: String { return "" }
```
bsudekum: Unsure what to do here. We need this prop on DirectionOptions but _always_ overiden by the class which subclasses it.

MapboxDirections/MBRoute.swift
```diff
@@ -174,7 +174,7 @@ open class Route: NSObject, NSSecureCoding {
      
      The route options object’s profileIdentifier property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.
      */
-    @objc open let routeOptions: RouteOptions
+    @objc open let routeOptions: DirectionOptions
```
bsudekum: This is an issue. What will happen in the navigation SDK when we use this param to reroute? Not all properties will be present.

MapboxDirections/MBDirectionOptions.swift
1ec5: Yeah, there’s no way in Objective-C to make an abstract class’s instance method purely virtual, as in C++, since “abstract class” is really just a convention rather than a compiler feature. In this situation, you could assert that this abstract method should never be called directly, then return an empty string.

On the other hand, this is a good case for using a protocol instead of an abstract class. The only difficulty with using a protocol is that we can’t provide a default implementation in an Objective-C-compatible protocol, so there would be a lot of code duplication. We’d have to find alternative ways to share code between RouteOptions and MatchOptions.

MapboxDirections/MBRoute.swift
1ec5: If we have RouteOptions and MatchOptions be siblings under DirectionsOptions, then Route and Match must be siblings under DirectionsResult (or whatever we’d call it). Then Route and Match would have a covariant `options` property that has the appropriate type.

MapboxDirectionsTests/MatchTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        // confirming actual decoded values is important because the Directions API
+        // uses an atypical precision level for polyline encoding
+        XCTAssertEqual(round(match!.coordinates!.first!.latitude), 33)
+        XCTAssertEqual(round(match!.coordinates!.first!.longitude), -117)
+        XCTAssertEqual(match!.legs.count, 6)
+        XCTAssertEqual(match!.confidence, 0.95)
+
+        let opts = match!.directionOptions
+
+        XCTAssertEqual(opts, matchOptions)
```
bsudekum: This assertion is failing and might point out a bigger issue with this approach. Is is possible to override a superclasses type? If so, in `Match`, we should override `directionOptions` to be of type `MatchOptions`. Perhaps the issue here is that we're casting the type.

Directions Example/MBDrawingView.swift
```diff
@@ -0,0 +1,101 @@
...
+}
+
+public class MBDrawingView: UIView {
+    
+    private var points: [CGPoint]!
+    private var context: CGContext!
+    private var strokeColor = UIColor.blue.withAlphaComponent(0.75)
+    private var lineWidth: CGFloat = 3
+    
+    public var delegate: MBDrawingViewDelegate?
```
vincethecoder: you probably want to declare this delegate as a `weak` reference.

Directions Example/MBDrawingView.swift
```diff
@@ -0,0 +1,101 @@
...
+    private func createContext() {
+        UIGraphicsBeginImageContextWithOptions(bounds.size, false, UIScreen.main.scale)
+        context = UIGraphicsGetCurrentContext()!
+        setStrokeColor(strokeColor: self.strokeColor)
+        setLineWidth(lineWidth: self.lineWidth)
+    }
+    
+    override public func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        points.removeAll(keepingCapacity: false)
+
```
vincethecoder: we should probably remove the points after we retrieve the `firstPoint` from the `touches`.

Directions Example/ViewController.swift
```diff
@@ -18,16 +20,58 @@ class ViewController: UIViewController {
         
         mapView = MGLMapView(frame: view.bounds)
         mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.setCenter(CLLocationCoordinate2D(latitude: 37.3300, longitude: -122.0312), animated: false)
```
vincethecoder: could we specify what location we are setting our map to, preferably setting it to a variable?

MapboxDirectionsTests/MatchTests.swift
1ec5: > If so, in Match, we should override directionOptions to be of type MatchOptions.

Yes, that’s called a covariant property.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -497,20 +161,14 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
         return (waypoints, routes)
     }
     
+    
     // MARK: NSCopying
-    open func copy(with zone: NSZone? = nil) -> Any {
-        let copy = RouteOptions(waypoints: waypoints, profileIdentifier: profileIdentifier)
+    override open func copy(with zone: NSZone? = nil) -> Any {
+//        let copy = RouteOptions(waypoints: waypoints, profileIdentifier: profileIdentifier)
+        let copy = super.copy() as! RouteOptions
```
bsudekum: Calling super here throws the process into an infinite loop. 

image

@JThramer any ideas?

1ec5: What’s the stack look like further down, where the recursion begins? What causes `DirectionsOptions.copy(with:)` to reenter `RouteOptions.copy(with:)`?

bsudekum: Unsure. Here is where super.copy is called: https://github.com/mapbox/MapboxDirections.swift/pull/236/files#diff-c1adc517aeca20c01aad5859dff47b21R213

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,454 @@
...
+    @objc public init(waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = nil) {
+        self.waypoints = waypoints
+        self.profileIdentifier = profileIdentifier ?? .automobile
+    }
+    
+    open static var supportsSecureCoding = true
+    
+    // MARK: NSCopying
+    open func copy(with zone: NSZone? = nil) -> Any {
+        let copy = DirectionOptions(waypoints: waypoints, profileIdentifier: profileIdentifier)
```
1ec5: This seems suspect. DirectionsOptions is an abstract class, so why would we create an instance of DirectionsOptions that isn’t a RouteOptions or MatchOptions? `self.init(waypoints:profileIdentifier:)` would seem more appropriate in this case.

MapboxDirections/MBRouteOptions.swift
1ec5: When you hit this infinite recursion, scroll down to where it stops saying `RouteOptions.copy(with:)`. What are the first few rows that are different?

Directions Example/ViewController.swift
```diff
@@ -18,16 +20,58 @@ class ViewController: UIViewController {
...
+        
+        setupSegmentedControll()
     }
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-
+    }
+    
+    func setupSegmentedControll() {
```
akitchen: nit: `setUpSegmentedControl()` (capital U and one l)

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,457 @@
+//
+//  MBDirectionOptions.swift
+//  MapboxDirections
+//
+//  Created by Bobby Sudekum on 2/19/18.
+//  Copyright © 2018 Mapbox. All rights reserved.
+//
```
akitchen: Shouldn't comment headers be removed?

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,449 @@
...
+@objc(MBDirecitonOptions)
+open class DirectionOptions: NSObject, NSSecureCoding, NSCopying {
+    
+    /**
+     Initializes a route options object for routes between the given waypoints and an optional profile identifier.
+     
+     - parameter waypoints: An array of `Waypoint` objects representing locations that the route should visit in chronological order. The array should contain at least two waypoints (the source and destination) and at most 25 waypoints. (Some profiles, such as `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, [may have lower limits](https://www.mapbox.com/api-documentation/#directions).)
+     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. This parameter, if set, should be set to `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`. `MBDirectionsProfileIdentifierAutomobile` is used by default.
+     */
+    @objc required public init(waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = nil) {
```
bsudekum: @1ec5 ideally, we don't have to make this change. This would make it so there are no changes to the nav sdk. Noting this is required because of https://github.com/mapbox/MapboxDirections.swift/pull/236/files#diff-c1adc517aeca20c01aad5859dff47b21R207

Are there any ways around this?

MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -0,0 +1,162 @@
+import Foundation
+
+@objc(MBMatchingOptions)
+open class MatchingOptions: DirectionOptions {
```
frederoni: nit: filename should match (no pun intended) class name.

bsudekum:
Holding on merging this until https://github.com/Project-OSRM/osrm-backend/issues/4928 is resolved.

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,449 @@
...
+     
+     This machine-readable format is considerably more compact than `geoJSON` but less precise than `polyline6`.
+     */
+    case polyline
+    /**
+     The route’s shape is delivered in [encoded polyline algorithm](https://developers.google.com/maps/documentation/utilities/polylinealgorithm) format with 1×10<sup>−6</sup> precision.
+     
+     This format is an order of magnitude more precise than `polyline`.
+     */
+    case polyline6
```
JThramer: One day, we'll have the ability to use associated-value enums.

`case polyline(version: FrameworkVersion)`

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,449 @@
...
+        switch self {
+        case .text:
+            return "text"
+        case .html:
+            return "html"
+        }
+    }
+}
+
+@objc(MBDirecitonOptions)
```
akitchen: oops typo

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,449 @@
...
+open class DirectionOptions: NSObject, NSSecureCoding, NSCopying {
```
JThramer: This is a stylistic nitpick, but the ordering of code for this class reads a little weird.

I generally try to keep things in the following order: 
`Class Constants` -> `Instance Properties` -> `Intializers` -> `Biz Logic` -> `Utility Methods` ->  `Delegate Conformance extensions` -> `Private Extensions`

Using a pattern like this tends to keep things consistent and enhances readability. It's just kind of weird when you're looking at an initializer and you encounter references to properties that you haven't seen yet.

MapboxDirections/MBDirectionOptions.swift
1ec5: No, anyways, it’s a good practice for a class with multiple initializers to mark one as `required` (“designated” in Objective-C parlance) and the others as `optional` (convenience).

MapboxDirections/MBDirectionOptions.swift
1ec5: Someday perhaps. Note that the “6” here refers to a precision (number of decimal places), not a version. Six is actually the usual value for the Polyline encoding, but OSRM v5 went with five places before adding a sixth as an option more recently for better interoperability.

MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -0,0 +1,41 @@
...
+      Index to the match object in matchings the sub-trace was matched to.
+     */
+    @objc open var matchingIndex: Int
+    
+    init(coordinate: CLLocationCoordinate2D, alternateCount: Int, waypointIndex: Int?, matchingIndex: Int, name: String?) {
+        self.alternateCount = alternateCount
+        self.waypointIndex = waypointIndex
+        self.matchingIndex = matchingIndex
+        super.init(coordinate: coordinate)
+        self.name = name
```
1ec5: Call the initializer that takes a CLLocation instead; that’ll set `name` automatically. Along the way, this initializer could take a CLLocation too.

MapboxDirectionsTests/MatchTests.swift
```diff
@@ -0,0 +1,115 @@
+import XCTest
+import OHHTTPStubs
+@testable import MapboxDirections
+
+class MatchTest: XCTestCase {
```
1ec5: Rename this class MatchTests to match the file name MatchTests.swift.

MapboxDirectionsTests/RoutableMatch.swift
```diff
@@ -0,0 +1,111 @@
...
+import OHHTTPStubs
+@testable import MapboxDirections
+
+class RoutableMatchTest: XCTestCase {
+    override func tearDown() {
+        OHHTTPStubs.removeAllStubs()
+        super.tearDown()
+    }
+    
+    func test() {
```
1ec5: Rename this method to something more descriptive. Consider moving this method into either MatchTests or V5Tests.

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -1134,9 +1263,11 @@
 			files = (
 				DAE9E0F41EB7DE2E001E8E8B /* RouteOptionsTests.swift in Sources */,
 				C5247D711E818A24004B6154 /* AnnotationTests.swift in Sources */,
+				C59666392048A20E00C45CE5 /* RoutableMatch.swift in Sources */,
```
1ec5: Rename this file to RoutableMatchTests.swift to match the class inside, RoutableMatchTests.

MapboxDirections/MBDirectionRoute.swift
```diff
@@ -0,0 +1,213 @@
...
+     `RouteOptions` used to create the directions request.
+     
+     The route options object’s profileIdentifier property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.
+     */
+    @objc open let directionOptions: DirectionOptions
+    
+    /**
+     The [access token](https://www.mapbox.com/help/define-access-token/) used to make the directions request.
+     
+     This property is set automatically if a request is made via `Directions.calculate(_:completionHandler:)`.
```
1ec5: That’s no longer the only method that sets this property automatically.

MapboxDirections/MBRoute.swift
```diff
@@ -52,170 +47,22 @@ open class Route: NSObject, NSSecureCoding {
...
-     Do not assume that the user would travel along the route at a fixed speed. For more granular travel times, use the `RouteLeg.expectedTravelTime` or `RouteStep.expectedTravelTime`. For even more granularity, specify the `AttributeOptions.expectedTravelTime` option and use the `RouteLeg.expectedSegmentTravelTimes` property.
-     */
-    @objc open let expectedTravelTime: TimeInterval
-    
-    /**
-     `RouteOptions` used to create the directions request.
-     
-     The route options object’s profileIdentifier property reflects the primary mode of transportation used for the route. Individual steps along the route might use different modes of transportation as necessary.
-     */
-    @objc open let routeOptions: RouteOptions
```
1ec5: Restore this property as a computed property:

```swift
@objc open let routeOptions: RouteOptions {
    return directionsOptions as! RouteOptions
}
```

and add a similar `matchingOptions` to Match.

MapboxDirectionsTests/V5Tests.swift
```diff
@@ -67,7 +67,7 @@ class V5Tests: XCTestCase {
         XCTAssertEqual(round(route!.coordinates!.first!.longitude), -122)
         XCTAssertEqual(route!.legs.count, 1)
         
-        let opts = route!.routeOptions
+        let opts = route!.directionOptions
```
1ec5: Revert this change once `Route.routeOptions` is restored.

MapboxDirectionsTests/MatchTests.swift
1ec5: Since a covariant property isn’t possible here, add a `matchingOptions` property to Match that returns a casted `directionsOptions`. Then assert here that it’s equal to the local `matchOptions` variable.

MapboxDirections/MBRoute.swift
bsudekum: It's still there, just moved: https://github.com/mapbox/MapboxDirections.swift/pull/236/files#diff-6212178132f11eb7e002f4e0973e9d66R53

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -0,0 +1,449 @@
+import Foundation
+
+/**
+ A `RouteShapeFormat` indicates the format of a route’s shape in the raw HTTP response.
```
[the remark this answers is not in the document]
bsudekum: Oh yeah this should be moved.

MapboxDirections/Match/MBMatchingOptions.swift
```diff
@@ -0,0 +1,162 @@
...
+     */
+    @objc open var resampleTraces: Bool = false
+    
+    
+    /**
+     Timestamps corresponding to each coordinate provided in the request.
+     
+     There must be as many `timestamps` as there are coordinates in the request.
+     */
+    @objc private var timestamps: [Date]?
```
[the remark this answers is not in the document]
bsudekum: By that time though, we only have access to a `Waypoint` which does not contain a timestamp property.

MapboxDirections/Match/MBMatchingOptions.swift
```diff
@@ -0,0 +1,162 @@
...
+        // This is a real bummer and is another place we need to maintain options.
+        let opts = RouteOptions(waypoints: waypoints, profileIdentifier: self.profileIdentifier)
+        opts.includesSteps = self.includesSteps
+        opts.includesVisualInstructions = self.includesVisualInstructions
+        opts.includesSpokenInstructions = self.includesSpokenInstructions
+        opts.attributeOptions = self.attributeOptions
+        opts.routeShapeResolution = self.routeShapeResolution
+        opts.distanceMeasurementSystem = self.distanceMeasurementSystem
+        opts.shapeFormat = self.shapeFormat
+        opts.locale = self.locale
```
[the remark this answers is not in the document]
bsudekum: Adding a not to self to come back to this issue.

MapboxDirections/Match/MBTracepoint.swift
bsudekum: I don't have access to a CLLocation at this point though. You think it's worth it to create one?

bsudekum:
todo: 

- [ ]  https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172684807
- [ ] https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172682010
- [ ] https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172037287

bsudekum:
@1ec5 I moved to use POST internally exclusively for matching. Can you check out https://github.com/mapbox/MapboxDirections.swift/pull/236/commits/0bc08e30465abfa5d24c7b2b76c0ab48e5cf23a7?

bsudekum:
@mapbox/navigation-ios a lot to take in here, but this this is ready for another round of review.

MapboxDirections/MBDirectionOptions.swift
JThramer: 🤔 Hmm, TIL!

Directions Example/MBDrawingView.swift
```diff
@@ -0,0 +1,100 @@
...
+    required public init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setup()
+    }
+    
+    deinit {
+        UIGraphicsEndImageContext()
+    }
+    
+    public func setStrokeColor(strokeColor: UIColor) {
```
JThramer: Why manual setters here and not in `strokeColor`'s `didSet`? Especially considering that these are `public`, I would argue that `setThing(_ thing: ThingType)` "setter" methods are an anti-pattern.

Directions Example/MBDrawingView.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.strokeColor = strokeColor
+        context.setStrokeColor(strokeColor.cgColor)
+    }
+    
+    public func setLineWidth(lineWidth: CGFloat) {
+        self.lineWidth = lineWidth
```
JThramer: and here

MapboxDirectionsTests/MatchTests.swift
```diff
@@ -0,0 +1,103 @@
...
+import OHHTTPStubs
+@testable import MapboxDirections
+
+class MatchTests: XCTestCase {
+    override func tearDown() {
+        OHHTTPStubs.removeAllStubs()
+        super.tearDown()
+    }
+    
+    func test() {
```
JThramer: I know it's punctilious, but could we have a more descriptive name for this test? maybe `testDirectionsMatchingResponseParsing`?

MapboxDirectionsTests/RoutableMatchTests.swift
```diff
@@ -0,0 +1,101 @@
...
+import OHHTTPStubs
+@testable import MapboxDirections
+
+class RoutableMatchTest: XCTestCase {
+    override func tearDown() {
+        OHHTTPStubs.removeAllStubs()
+        super.tearDown()
+    }
+    
+    func test() {
```
JThramer: Here too.

Directions Example/ViewController.swift
```diff
@@ -18,16 +21,58 @@ class ViewController: UIViewController {
         
         mapView = MGLMapView(frame: view.bounds)
         mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        mapView.setCenter(ViewController.initialMapCenter, animated: false)
+        mapView.setZoomLevel(15, animated: false)
```
JThramer: Constant-ize as `ExampleStartingZoom`? Dealer's choice, it's the example.

JThramer: Or we could follow the pattern of the line above... `ViewController.initialZoom`

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -184,8 +184,11 @@ public enum InstructionFormat: UInt, CustomStringConvertible {
     }
 }
 
-@objc(MBDirecitonOptions)
-open class DirectionOptions: NSObject, NSSecureCoding, NSCopying {
+/**
+ A base class that contains shared resources between `MatchOptions` and `RouteOptions`.
```
1ec5: This is a public class, so document what the developer needs to know, not what we need to know internally:

> Options for calculating results from the Mapbox Directions service.
> 
> You do not create instances of this class directly. Instead, create instances of `MatchOptions` or `RouteOptions`.

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -301,7 +304,10 @@ open class DirectionOptions: NSObject, NSSecureCoding, NSCopying {
         return waypoints.map { "\($0.coordinate.longitude),\($0.coordinate.latitude)" }
     }
     
-    internal var path: String { return "" }
+    internal var path: String {
+        assert(self.path.isEmpty, "path should be overriden by subclass")
```
1ec5: This would result in infinite recursion rather than a failing assertion. Assert `false` instead.

MapboxDirections/MBDirectionRoute.swift
```diff
@@ -1,60 +1,20 @@
 import Polyline
 
 /**
- A `Route` object defines a single route that the user can follow to visit a series of waypoints in order. The route object includes information about the route, such as its distance and expected travel time. Depending on the criteria used to calculate the route, the route object may also include detailed turn-by-turn instructions.
- 
- Typically, you do not create instances of this class directly. Instead, you receive route objects when you request directions using the `Directions.calculate(_:completionHandler:)` method. However, if you use the `Directions.url(forCalculating:)` method instead, you can pass the results of the HTTP request into this class’s initializer.
+ A `DirectionRoute` represents a result returned from either the Directions API or the Map Matching API.
```
1ec5: `DirectionsResult`, not `DirectionRoute`.

MapboxDirections/MBDirectionRoute.swift
```diff
@@ -1,60 +1,20 @@
 import Polyline
 
 /**
- A `Route` object defines a single route that the user can follow to visit a series of waypoints in order. The route object includes information about the route, such as its distance and expected travel time. Depending on the criteria used to calculate the route, the route object may also include detailed turn-by-turn instructions.
- 
- Typically, you do not create instances of this class directly. Instead, you receive route objects when you request directions using the `Directions.calculate(_:completionHandler:)` method. However, if you use the `Directions.url(forCalculating:)` method instead, you can pass the results of the HTTP request into this class’s initializer.
```
1ec5: Something along these lines is still necessary, so that developers (especially when coding in Objective-C) don’t mistakenly create one of these objects directly.

/ref https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172036593

MapboxDirections/MBDirections.swift
```diff
@@ -107,8 +107,6 @@ open class Directions: NSObject {
     
     public typealias MatchCompletionHandler = (_ waypoints: [Tracepoint]?, _ routes: [Match]?, _ error: NSError?) -> Void
```
1ec5: These typealiases need documentation comments. Otherwise the jazzy documentation won’t show their underlying signatures.

MapboxDirections/MBDirections.swift
```diff
@@ -187,11 +185,20 @@ open class Directions: NSObject {
         return task
     }
     
-    @objc(calculateMatchingWithOptions:completionHandler:)
+    /**
+     Begins asynchronously calculating a match using the given options and delivers the results to a closuer.
```
1ec5: Typo: “closure”.

MapboxDirections/MBDirections.swift
```diff
@@ -183,6 +187,46 @@ open class Directions: NSObject {
         return task
     }
     
+    @objc(calculateMatchingWithOptions:completionHandler:)
+    @discardableResult open func match(_ options: MatchingOptions, completionHandler: @escaping MatchCompletionHandler) -> URLSessionDataTask {
```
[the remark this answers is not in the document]
1ec5: A reminder to rename this method to `calculate(_:completionHandler:)` in Swift and `-calculateMatchesWithOptions:completionHandler:` in Objective-C.

MapboxDirections/MBDirections.swift
```diff
@@ -207,11 +214,12 @@ open class Directions: NSObject {
         return task
     }
     
-    @objc(calculateRoutableMatchWithOptions:completionHandler:)
-    @discardableResult open func routableMatch(_ options: MatchingOptions, completionHandler: @escaping RouteableMatchCompletionHandler) -> URLSessionDataTask {
+    @objc(calculateRoutesMatchingOptions:completionHandler:)
+    @discardableResult open func calculateRoutes(_ options: MatchingOptions, completionHandler: @escaping RouteCompletionHandler) -> URLSessionDataTask {
```
1ec5: Per https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172036785, this method name should be `calculateRoutes(matching:completionHandler:)`, not `calculateRoutes(_:completionHandler:)`.

MapboxDirections/MBDirections.swift
```diff
@@ -272,10 +287,17 @@ open class Directions: NSObject {
      After requesting the URL returned by this method, you can parse the JSON data in the response and pass it into the `Route.init(json:waypoints:profileIdentifier:)` initializer.
      */
     @objc(URLForCalculatingDirectionsWithOptions:)
-    open func url(forCalculating options: DirectionOptions) -> URL {
-        let params = options.params + [
-            URLQueryItem(name: "access_token", value: accessToken),
-        ]
+    open func url(forCalculating options: DirectionsOptions) -> URL {
+        let params: [URLQueryItem]
+        if options is MatchingOptions {
```
1ec5: Directions shouldn’t have any code that switches on the type of `options` like this. This conditional seems to exist for the purpose of ignoring the `options.params` property. That’s surprising. It suggests that we need to push more functionality into RouteOptions and MatchingOptions rather than working around the functionality here in Directions.

MapboxDirections/MBDirectionOptions.swift
```diff
@@ -218,13 +221,13 @@ open class DirectionOptions: NSObject, NSSecureCoding, NSCopying {
...
-        guard let opts = object as? DirectionOptions else { return false }
+        guard let opts = object as? DirectionsOptions else { return false }
         return isEqual(to: opts)
     }
     
-    @objc(isEqualToDirectionOptions:)
-    open func isEqual(to directionOptions: DirectionOptions?) -> Bool {
-        guard let other = directionOptions else { return false }
+    @objc(isEqualToDirectionsOptions:)
+    open func isEqual(to directionsOptions: DirectionsOptions?) -> Bool {
```
1ec5: https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172036916

```swift
matchOptions.isEqual(to: routeOptions) // true!
```

MapboxDirections/MBRouteOptions.swift
```diff
@@ -39,6 +39,10 @@ open class RouteOptions: DirectionOptions {
         super.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
         self.allowsUTurnAtWaypoint = ![MBDirectionsProfileIdentifier.automobile.rawValue, MBDirectionsProfileIdentifier.automobileAvoidingTraffic.rawValue].contains(self.profileIdentifier.rawValue)
     }
+    
+    @objc internal convenience init(matchOptions: MatchingOptions) {
+        self.init(waypoints: matchOptions.waypoints, profileIdentifier: matchOptions.profileIdentifier)
```
1ec5: Do any of the other properties of MatchingOptions need to be carried over? That can happen here in this initializer, after the call to `init(waypoints:profileIdentifier:)`.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -1,14 +1,26 @@
 import Polyline
 
+/**
+ A `Match` object defines a single route that was created from a series of points that were matched against a road network.
+ 
+ Typically, you do not create instances of this class directly. Instead, you receive route objects when you request directions using the `Directions.match(_:completionHandler:)` method.
```
1ec5: > Instead, you receive match objects when you pass a `MatchingOptions` object into the `Directions.calculate(_:completionHandler:)` or `Directions.calculateRoutes(matching:completionHandler:)` method.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -1,14 +1,26 @@
...
         self.confidence = confidence
-        super.init(directionOptions: matchOptions, legs: legs, distance: distance, expectedTravelTime: expectedTravelTime, coordinates: coordinates, speechLocale: speechLocale)
+        super.init(options: matchOptions, legs: legs, distance: distance, expectedTravelTime: expectedTravelTime, coordinates: coordinates, speechLocale: speechLocale)
     }
     
-    convenience init(json: [String: Any], tracePoints: [Tracepoint], matchOptions: MatchingOptions) {
+    /**
+     Initializes a new match object with the given JSON dictionary representation and tracepoints.
+     
+     - parameter json: A JSON dictionary representation of the route as returned by the Mapbox Directions API.
```
1ec5: This class works with the Map Matching API, not the Directions API.

MapboxDirections/Match/MBTracepoint.swift
1ec5: After the call to `CLLocation(coordinate:)`, this object has an initialized `location` property of type CLLocation.

MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -11,16 +11,16 @@ public class Tracepoint: Waypoint {
     /**
      Index of the waypoint inside the matched route.
      */
-    open var waypointIndex: Int?
+    @objc open var waypointIndex: Int = NSNotFound
     
     /**
-      Index to the match object in matchings the sub-trace was matched to.
+      If more than one `Match` are returned, this index represents which `Match` this tracepoint was used on.
```
1ec5: This is a very poor API. The tracepoint object has an index into an arbitrary array of match objects. What if client code rearranges the array or creates a new array? Then this index is meaningless but is still set.

This suggests very strongly that `MatchCompletionHandler`’s first argument (`tracepoints`) should be removed, and instead Match should have a property containing an array of tracepoints.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -1,14 +1,26 @@
...
+     - parameter tracePoints: An array of `Tracepoint` that the match found in order.
+     - parameter matchOptions: The `MatchingOptions` used to create the request.
+    */
+    @objc public convenience init(json: [String: Any], tracePoints: [Tracepoint], matchOptions: MatchingOptions) {
```
1ec5: Typo: `tracepoints`, not `tracePoints`. (Also in the documentation above.)

MapboxDirections/MBDirections.swift
1ec5: If we keep the first argument, it should be named `tracepoints`, not `waypoints`. The second argument should be named `matches`, not `routes`.

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -78,13 +80,41 @@
 		C58EA7AB1E9D7F73008F98CE /* MBCongestion.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58EA7A91E9D7EAD008F98CE /* MBCongestion.swift */; };
 		C58EA7AC1E9D7F74008F98CE /* MBCongestion.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58EA7A91E9D7EAD008F98CE /* MBCongestion.swift */; };
 		C58EA7AD1E9D7F75008F98CE /* MBCongestion.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58EA7A91E9D7EAD008F98CE /* MBCongestion.swift */; };
+		C59094B5203B5C7F00EB2417 /* MBDrawingView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C59094B4203B5C7F00EB2417 /* MBDrawingView.swift */; };
+		C59094BF203B800300EB2417 /* MBDirectionOptions.swift in Sources */ = {isa = PBXBuildFile; fileRef = C59094BE203B800300EB2417 /* MBDirectionOptions.swift */; };
+		C59094C1203DE6BC00EB2417 /* MBDirectionRoute.swift in Sources */ = {isa = PBXBuildFile; fileRef = C59094C0203DE6BC00EB2417 /* MBDirectionRoute.swift */; };
```
1ec5: MBDirectionRoute.swift should be renamed MBDirectionsResult.swift.

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -78,13 +80,41 @@
 		C58EA7AB1E9D7F73008F98CE /* MBCongestion.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58EA7A91E9D7EAD008F98CE /* MBCongestion.swift */; };
 		C58EA7AC1E9D7F74008F98CE /* MBCongestion.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58EA7A91E9D7EAD008F98CE /* MBCongestion.swift */; };
 		C58EA7AD1E9D7F75008F98CE /* MBCongestion.swift in Sources */ = {isa = PBXBuildFile; fileRef = C58EA7A91E9D7EAD008F98CE /* MBCongestion.swift */; };
+		C59094B5203B5C7F00EB2417 /* MBDrawingView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C59094B4203B5C7F00EB2417 /* MBDrawingView.swift */; };
+		C59094BF203B800300EB2417 /* MBDirectionOptions.swift in Sources */ = {isa = PBXBuildFile; fileRef = C59094BE203B800300EB2417 /* MBDirectionOptions.swift */; };
```
1ec5: MBDirectionOptions.swift should be renamed MBDirectionsOptions.swift.

MapboxDirections/Match/MBMatchingOptions.swift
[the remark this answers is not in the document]
1ec5: What about `Waypoint.location.timestamp`?

MapboxDirections/MBDirections.swift
```diff
@@ -103,7 +103,17 @@ open class Directions: NSObject {
         If the request was canceled or there was an error obtaining the routes, this parameter is `nil`. This is not to be confused with the situation in which no results were found, in which case the array is present but empty.
      - parameter error: The error that occurred, or `nil` if the placemarks were obtained successfully.
      */
-    public typealias CompletionHandler = (_ waypoints: [Waypoint]?, _ routes: [Route]?, _ error: NSError?) -> Void
+    public typealias RouteCompletionHandler = (_ waypoints: [Waypoint]?, _ routes: [Route]?, _ error: NSError?) -> Void
+    
+    /**
+     A closure (block) to be called when a map matching request is complete.
+     
+     - parameter tracepoint: An array of `Tracepoint` objects.
```
1ec5: There’s no longer a `tracepoint` argument in this signature.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,74 @@
...
+        self.init(matchOptions: matchOptions, legs: legs, tracepoints: tracepoints, distance: distance, expectedTravelTime: expectedTravelTime, coordinates: coordinates, confidence: confidence, speechLocale: speechLocale)
+    }
+    
+    /**
+     A number between 0 and 1 that indicates the Map Matching API’s confidence that the match is accurate. A higher confidence means the match is more likely to be accurate.
+     */
+    @objc open var confidence: Float
+    
+    
+    @objc let tracepoints: [Tracepoint]?
```
1ec5: This property needs documentation and should be public.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    public var matchOptions: MatchingOptions {
+        return super.directionsOptions as! MatchingOptions
+    }
+    
+    @objc public required convenience init?(coder decoder: NSCoder) {
+        self.init(coder: decoder)
+        confidence = decoder.decodeFloat(forKey: "confidence")
+    }
```
1ec5: Also encode and decode the tracepoints, and potentially override the equality methods to consider whether the tracepoints are equal.

MapboxDirections/Match/MBMatchingOptions.swift
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     Timestamps corresponding to each coordinate provided in the request.
+     
+     There must be as many `timestamps` as there are coordinates in the request.
+     */
+    @objc private var timestamps: [Date]?
+    
+    
+    /**
+     An IndexSet of unique integers representing which coordinates should be treated as `Wapoints`.
```
1ec5: Typo: `Waypoint`s.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    /**
+     A number between 0 and 1 that indicates the Map Matching API’s confidence that the match is accurate. A higher confidence means the match is more likely to be accurate.
+     */
+    @objc open var confidence: Float
+    
+    
+    /**
+     Represents a the location an input location was matched with.
```
1ec5: > Tracepoints on the road network that match the tracepoints in the matching options.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,85 @@
...
+        confidence = decoder.decodeFloat(forKey: "confidence")
+        
+        guard let tracepoints = decoder.decodeObject(of: [NSArray.self, Tracepoint.self], forKey: "tracepoints") as? [Tracepoint] else {
+            return nil
+        }
+        self.tracepoints = tracepoints
+    }
+    
+    @objc public override func encode(with coder: NSCoder) {
+        coder.encode(confidence, forKey: "confidence")
```
1ec5: Also encode the tracepoints, and add a unit test to ensure that this encoding and decoding work correctly.

bsudekum:
@1ec5 I think I hit all your comments.

MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -0,0 +1,32 @@
...
+@objc(MBTracepoint)
+public class Tracepoint: Waypoint {
+    
+    /**
+     Number of probable alternative matchings for this trace point. A value of zero indicates that this point was matched unambiguously.
+     */
+    @objc open var alternateCount: Int
+    
+    /**
+     Index of the waypoint inside the matched route.
```
1ec5: It’s an antipattern to put the index of an item inside the item itself. As soon as you isolate the tracepoint from the array of tracepoints, this property becomes meaningless or misleading.

Instead, Match should have a `waypoints` property that contains whichever items in `tracepoints` happens to be a waypoint. If for some reason you want to know the index of a waypoint, you could look up the Tracepoint object’s index in `Match.waypoints`.

/ref https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r163961821

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,86 @@
...
+        coder.encode(tracepoints, forKey: "tracepoints")
+    }
+}
```
1ec5: If two matches are identical except for the tracepoints that begot them, are the matches equal? If not, override `isEqual(_:)` and `isEqual(to:)`.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,86 @@
...
+    init(matchOptions: MatchingOptions, legs: [RouteLeg], tracepoints: [Tracepoint], distance: CLLocationDistance, expectedTravelTime: TimeInterval, coordinates: [CLLocationCoordinate2D]?, confidence: Float, speechLocale: Locale?) {
+        self.confidence = confidence
+        self.tracepoints = tracepoints
+        super.init(options: matchOptions, legs: legs, distance: distance, expectedTravelTime: expectedTravelTime, coordinates: coordinates, speechLocale: speechLocale)
+    }
+    
+    /**
+     Initializes a new match object with the given JSON dictionary representation and tracepoints.
+     
+     - parameter json: A JSON dictionary representation of the route as returned by the Mapbox Mapbox Map Matching API.
```
1ec5: Typo: Mapbox Mapbox.

MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -0,0 +1,457 @@
...
+    
+    //MARK: - OBJ-C Equality
+    open override func isEqual(_ object: Any?) -> Bool {
+        guard let opts = object as? DirectionsOptions else { return false }
+        return isEqual(to: opts)
+    }
+    
+    @objc(isEqualToDirectionsOptions:)
+    open func isEqual(to directionsOptions: DirectionsOptions?) -> Bool {
+        guard let other = directionsOptions else { return false }
```
1ec5: This method should insist that the two objects are of exactly the same class. Otherwise, you can ask a MatchOptions if it’s equal to a RouteOptions, and it might say yes. That might be handy in some cases, but it’s better to do that sort of equivalence checking explicitly rather than relying on an equality method.

/ref https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172036916

MapboxDirections/Match/MBMatchingOptions.swift
```diff
@@ -0,0 +1,159 @@
+import Foundation
+
+@objc(MBMatchingOptions)
+open class MatchingOptions: DirectionsOptions {
```
1ec5: This class needs a documentation comment.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,86 @@
...
+    
+    /**
+     A number between 0 and 1 that indicates the Map Matching API’s confidence that the match is accurate. A higher confidence means the match is more likely to be accurate.
+     */
+    @objc open var confidence: Float
+    
+    
+    /**
+     Tracepoints on the road network that match the tracepoints in the matching options.
+     */
```
1ec5: The API documentation [says](https://www.mapbox.com/api-documentation/#match-response-object):

> If a trace point is omitted by map matching because it is an outlier, the entry will be `null`.

We should account for that edge case, then add this to the documentation:

> Any outlier tracepoint is omitted from the match. This array represents an outlier tracepoint is a `Tracepoint` object whose `Tracepoint.coordinate` property is `kCLLocationCoordinate2DInvalid`.

MapboxDirections/Match/MBMatchingOptions.swift
1ec5: Per chat, did you want to rename this class back to MatchOptions, for consistency with RouteOptions?

/ref https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r163955397

MapboxDirections/Match/MBMatchingOptions.swift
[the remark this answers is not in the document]
1ec5: Remove this property in favor of a new `timestamp` property on Tracepoint. (Sorry, I had been under the mistaken impression that Waypoint had a `location` property of type CLLocation. Wouldn’t be a bad idea.)

In `params`’ getter, map each waypoint to its `timestamp`. In Tracepoint, override `Waypoint(location:heading:name:)` to also store the location’s `timestamp`.

/ref https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r172037100

MapboxDirections/MBDirectionsResult.swift
```diff
@@ -0,0 +1,171 @@
+import Polyline
+
+/**
+ A `DirectionsResult` represents a result returned from either the Directions API or the Map Matching API.
```
1ec5: > A `DirectionsResult` represents a result returned from either the Mapbox Directions service.
> 
> You do not create instances of this class directly. Instead, you receive `Route` or `Match` objects when you request directions using the `Directions.calculate(_:completionHandler:)` or `Directions.calculateRoutes(matching:completionHandler:)` method.

“Mapbox Directions service” encompasses the Directions API and Map Matching API. At least that’s how it’s presented in the sidebar of the [API Documentation](https://www.mapbox.com/api-documentation/#directions) site.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+    
+    /**
+     A number between 0 and 1 that indicates the Map Matching API’s confidence that the match is accurate. A higher confidence means the match is more likely to be accurate.
+     */
+    @objc open var confidence: Float
+    
+    
+    /**
+     Tracepoints on the road network that match the tracepoints in the matching options.
```
1ec5: Nit: match options.

MapboxDirections/MBDirections.swift
```diff
@@ -183,6 +191,56 @@ open class Directions: NSObject {
...
+            }
+            completionHandler(response, nil)
+        }) { (error) in
+            completionHandler(nil, error)
+        }
+        task.resume()
+        return task
+    }
+    
+    @objc(calculateRoutesMatchOptions:completionHandler:)
```
1ec5: Rename this back to `calculateRoutesMatchingOptions:completionHandler:`.

MapboxDirections/Match/MBMatch.swift
```diff
@@ -0,0 +1,103 @@
...
+     
+     Any outlier tracepoint is omitted from the match. This array represents an outlier tracepoint is a `Tracepoint` object whose `Tracepoint.coordinate` property is `kCLLocationCoordinate2DInvalid`.
+     */
+    @objc open var tracepoints: [Tracepoint]
```
1ec5: 1ad509535565d440c74865639cb1d695a2589a23 removed the `Waypoint.waypointIndex` property in response to https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r173983435. But it doesn’t appear to have been replaced by anything, so it isn’t possible to tell which tracepoints should be treated like waypoints for the purpose of turning the match into a route.

Add a new property to Match: either a `waypointIndices` property of type IndexSet (for consistency with MatchOptions), or a `waypoints` property of type `[Tracepoint]` that holds whichever tracepoints should be treated like waypoints.

As tail work, we should explore having Waypoint inherit from Tracepoint instead of the other way around, since not all tracepoints are waypoints but all waypoints are tracepoints.

MapboxDirectionsTests/RoutableMatchTests.swift
```diff
@@ -0,0 +1,101 @@
+import XCTest
+import OHHTTPStubs
+@testable import MapboxDirections
+
+class RoutableMatchTest: XCTestCase {
+    override func tearDown() {
+        OHHTTPStubs.removeAllStubs()
+        super.tearDown()
+    }
+
```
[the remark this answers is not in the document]
1ec5: A reminder to add a `testCoding()` method.

/ref https://github.com/mapbox/MapboxDirections.swift/pull/236#discussion_r173975686

MapboxDirections/Match/MBMatch.swift
```diff
@@ -81,6 +88,8 @@ open class Match: DirectionsResult {
         self.tracepoints = tracepoints
     }
     
+    open var supportsSecureCoding = true
```
bsudekum: Ugh: 

```
Test Case '-[MapboxDirectionsTests.MatchTests testMatchWithNullTracepoints]' started.
<unknown>:0: error: -[MapboxDirectionsTests.MatchTests testMatchWithNullTracepoints] : failed: caught "NSInvalidUnarchiveOperationException", "Class 'MBMatch' has a superclass that supports secure coding, but 'MBMatch' overrides -initWithCoder: and does not override +supportsSecureCoding. The class must implement +supportsSecureCoding and return YES to verify that its implementation of -initWithCoder: is secure coding compliant."
```

bsudekum:
Ok, have this down to one last test failure: 

```
<unknown>:0: error: -[MapboxDirectionsTests.MatchTests testMatchWithNullTracepoints] : failed: caught "NSInvalidUnarchiveOperationException", "value for key 'directionsOptions' was of unexpected class 'MBMatchOptions'. Allowed classes are '{(
    MBDirectionsResult
)}'."
(
```

MapboxDirections/MBDirectionsResult.swift
```diff
@@ -0,0 +1,175 @@
...
+            } else {
+                return nil
+            }
+        })
+        
+        legs = decoder.decodeObject(of: [NSArray.self, RouteLeg.self], forKey: "legs") as? [RouteLeg] ?? []
+        distance = decoder.decodeDouble(forKey: "distance")
+        expectedTravelTime = decoder.decodeDouble(forKey: "expectedTravelTime")
+        
+        guard let options = decoder.decodeObject(of: [DirectionsResult.self], forKey: "directionsOptions") as? DirectionsOptions else {
```
1ec5: This object is expected to be of type DirectionsOptions, not DirectionsResult.

1ec5:
Opened #248 for documenting the new Map Matching API support.

PR #237, opened 2018-01-25
Add tunnel RoadClass

bsudekum:
The RoadClass tunnel has been added to the Directions API.

/cc @mapbox/navigation-ios


bsudekum:
@1ec5 In the navigation changelog?

1ec5:
Hah, too many repos!

Issue #238, opened 2018-02-09
Add type/modifier to VisualInstruction class


bsudekum:
Banner objects now include the proper type/modifier how the banner should be visualized.

![image](https://user-images.githubusercontent.com/1058624/36038746-632635d4-0d75-11e8-8bc6-80a36ae855bb.png)

/cc @frederoni @1ec5

PR #239, opened 2018-02-09
Add maneuver and modifier to visual instructions

bsudekum:
Closes: https://github.com/mapbox/MapboxDirections.swift/issues/238

Adds `maneuverType`  and `maneuverDirection` to the VisualInstruction class.

/cc @lyzidiamond @mapbox/navigation-ios


MapboxDirections/MBVisualInstruction.swift
```diff
@@ -41,6 +41,18 @@ open class VisualInstruction: NSObject, NSSecureCoding {
...
+     :nodoc:
+     The maneuver type for the `VisualInstruction`.
+     */
+    @objc public var maneuverType: ManeuverType
+    
+    /**
+     :nodoc:
+     The modifier type for the `VisualInstruction`.
+     */
+    @objc public var maneuverDirection: ManeuverDirection
```
bsudekum: I'm up for calling this something else.

bsudekum: Soon the secondary and primary components will each have type and modifier. How should these properties be renamed?

`primaryManeuverType` and `secondaryManeuverType`?

frederoni: How about wrapping primary and secondary in an instructions object of its own?
Not sure about the name but something along these lines:
```swift
instructions.primary.maneuverType
instructions.secondary.maneuverType
```

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -63,19 +82,25 @@ open class VisualInstruction: NSObject, NSSecureCoding {
             }
         }
         
-        self.init(distanceAlongStep: distanceAlongStep, primaryText: primaryText, primaryTextComponents: primaryTextComponents, secondaryText: secondaryText, secondaryTextComponents: secondaryTextComponents)
+        let maneuverType = ManeuverType(description: json["type"] as! String) ?? .none
+        let maneuverDirection = ManeuverDirection(description: json["modifier"] as! String)  ?? .none
```
frederoni: Should we unwrap the modifier gracefully or make it non-optional server-side?

bsudekum: This is something we have sort of struggled with deciding upon recently. It is a required key on the server, but as you have seen, there is a bug which omits it in some cases causing a crash.

Not really sure what the right answer here.

/cc @1ec5 @JThramer @lyzidiamond

MapboxDirections/MBVisualInstruction.swift
bsudekum: Does this make sense even though this property is an array of VisualInstructionsComponent?

https://github.com/mapbox/MapboxDirections.swift/blob/fc3e74bdf00e0eb1bf550e731fd5a1b00cfe9c9a/MapboxDirections/MBVisualInstruction.swift#L22-L28

Would each component have a type + modifier then?

MapboxDirections/MBVisualInstruction.swift
JThramer: 🤔 hmm, this is a stumper. If a missing `maneuverType` / `maneuverDirection` leads to significant undefined behavior later-on, then silently failing would be a bad idea. 

I think a chore switching all of the model logic over to `Codable` will help this issue out _a lot_ (as `Codable` will throw an error when it fails, which you can then turn into a understandable user error message), but if we want to do a `preconditionFailure(message:)` or something for now that spells out to the user exactly what is going on, that would be cool.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -41,41 +41,53 @@ open class VisualInstruction: NSObject, NSSecureCoding {
...
+            VisualInstructionComponent(json: $0, maneuverType: maneuverType, maneuverDirection: maneuverDirection)
         }
         
         var secondaryText: String?
         var secondaryTextComponents: [VisualInstructionComponent]?
         if let secondaryTextComponent = json["secondary"] as? JSONDictionary {
             secondaryText = secondaryTextComponent["text"] as? String
             secondaryTextComponents = (secondaryTextComponent["components"] as! [JSONDictionary]).map {
-                VisualInstructionComponent(json: $0)
+                VisualInstructionComponent(json: $0, maneuverType: maneuverType, maneuverDirection: maneuverDirection)
```
bsudekum: @mapbox/navigation-ios  I'm not very into this, but it does address https://github.com/mapbox/MapboxDirections.swift/pull/239#discussion_r167655759. I think these visual classes should be refactored to be more inline with @frederoni proposed API. This should happen as tail work after this PR is merged.

MapboxDirections/MBVisualInstruction.swift
1ec5: A client library should never silently swallow bugs that originate from the server. I’m OK with an assertion (or `preconditionFailure`) in this case for now, but ultimately an exception (that leads to the error handler being called) would be the ideal way to handle malformed server responses.

MapboxDirections/MBVisualInstruction.swift
1ec5: I think the data structure is already uncomfortably deeply nested. It might be a different story if we were working purely with Swift structs or tuples, but classes have a certain amount of mental overhead. How about `primaryManeuverType` and `secondaryManeuverType`, same for modifiers?

bsudekum:

frederoni:
Unwrapping type and modifier still crashes on the majority of the routes.

bsudekum:
@frederoni noting, the fix was not fully deployed until now.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -41,41 +41,53 @@ open class VisualInstruction: NSObject, NSSecureCoding {
...
+    
+    /**
+     :nodoc:
+     Indicates what side of a bidirectional road the driver must be driving on. Also referred to as the rule of the road.
+     */
+    @objc public var drivingSide: DrivingSide
+    
     /**
      :nodoc:
      Initialize a `VisualInstruction` from a dictionary.
```
vincethecoder: should we update the brief for this convenience init?

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -41,41 +41,53 @@ open class VisualInstruction: NSObject, NSSecureCoding {
...
      */
-    @objc public convenience init(json: [String: Any]) {
+    @objc public convenience init(json: [String: Any], drivingSide: DrivingSide) {
```
vincethecoder: Question, could we have an additional convenience init which specifies driving side, this way we can offer 2 initialization types.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -104,5 +123,6 @@ open class VisualInstruction: NSObject, NSSecureCoding {
         coder.encode(primaryTextComponents, forKey: "primaryTextComponents")
         coder.encode(secondaryText, forKey: "secondaryText")
         coder.encode(secondaryTextComponents, forKey: "secondaryTextComponents")
+        coder.encode(drivingSide, forKey: "drivingSide")
```
vincethecoder: Minor: possibly we could considers setting our keys to a constant variable within this `VisualInstruction` class. (Just a future maintenance precaution measure.)

bsudekum: Very true, might be solved with https://github.com/mapbox/MapboxDirections.swift/pull/221

PR #243, opened 2018-02-27
Add new types

bsudekum:
* Changes `.destination` -> `.text`
* Adds `.image`
* Changes how we parse out `.delimiter`

I'm thinking icon should become image.

/cc @mapbox/navigation-ios @lyzidiamond

PR #244, opened 2018-02-27
Add abbreviation support

bsudekum:
Adds support for abbreviations.

Going to update fixture and tests dependent on which ever can merge first between this PR and https://github.com/mapbox/MapboxDirections.swift/pull/243

/cc @mapbox/navigation-ios @danpaz


MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -49,6 +49,16 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
     
+    /**
+     An abbreviated version of the text for a given component.
+     */
+    @objc public var abbreviation: String?
+    
+    /**
+     The priority in which the component should be abbreviated. Lower numbers should be abbreviated first.
+     */
+    @objc public var abbreviationPriority: Int
```
bsudekum: Although this property is optional, objc does not support optional Ints. Because of this, the default value will be zero. Anyone have objections here?

danpaz: No concerns from me. Just noting that `abbr_priority: 0` is the highest priority, so you might have to add a check to excludes abbreviations where `abbr` is undefined when putting abbreviations together.

bsudekum: On second thought, I think it's important it is optional. Going to use NSNumber.

bsudekum: Changed the default value to `Int.max` since zero would be incorrect.

bsudekum:

PR #246, opened 2018-03-05
Changed access level of MBRouteOptions response function & MBRouteStep instructions parameter

wishtrip-dev:
* ### Changed access level for `response(from json: JSONDictionary)` function from `internal` to `public`
Our backend, similar to Mapbox backend, can create Routes to navigate. Accordingly, we are interested that the method responsible for parsing route JSON file should be public, because we receive the route JSON file from our server outside of the Mapbox framework. 
* ### `instructions` parameter changed from `let` to `var`
Several of our custom routes have special instructions, which cannot be loaded while creating Route objects. We would like `MBRouteStep`'s `instructions` parameter to be declared as a variable, allowing it to be changed dynamically after the Route object is created. 

We believe these changes will be useful for other customers of MapboxDirections.swift.

### DIFF
--- a/MapboxDirections/MBRouteOptions.swift
+++ b/MapboxDirections/MBRouteOptions.swift
@@ -477,7 +477,7 @@ open class RouteOptions: NSObject, NSSecureCoding, NSCopying{
      - parameter json: The API response in JSON dictionary format.
      - returns: A tuple containing an array of waypoints and an array of routes.
      */
-    internal func response(from json: JSONDictionary) -> ([Waypoint]?, [Route]?) {
+    public func response(from json: JSONDictionary) -> ([Waypoint]?, [Route]?) {
         var namedWaypoints: [Waypoint]?
         if let jsonWaypoints = (json["waypoints"] as? [JSONDictionary]) {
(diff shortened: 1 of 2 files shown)


frederoni:
Thanks for the PR.
In order to make it build and pass on CI
- Change the `RouteOptions.response(from:)` signature to take a `[String: Any]`. (We could make the typealias `JSONDictionary` public but I don't think MapboxDirections should be in the business of defining what a json dictionary is.
- Change the access level of `RouteOptionsV4.response(from:)` to public.

wishtrip-dev:
A better solution was found (i.e. using Route.init(json:waypoints:routeOptions:) to create Route object from custom route JSON). There's no need for pull request.

gregoireLem:
@wishtrip-dev do you have an exemple on how you created your objects required in Route.init ?

I have an error : Unexpectedly found nil while unwrapping an Optional value in RouteStep.init(son:) when calling Route.init ..

thx

1ec5:
`RouteStep(json:)` expects a JSON-serializable object that matches the [Mapbox Directions API](https://www.mapbox.com/api-documentation/#directions)’s documented response format. #198 would add a convenience layer for custom OSRM instances.

PR #250, opened 2018-03-19
Serialize speed limit annotations

bsudekum:
This adds a new annotation attribute, `segmentMaximumSpeedLimits` which represents the maximum speed limit for a segment.

/cc @mapbox/navigation-ios @karenzshea


MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -0,0 +1,118 @@
+import Foundation
+
+/**
+ A localized limit for measuring speed limits.
+ */
+@objc(MBSpeedLimit)
+public class SpeedLimit: NSObject, NSSecureCoding {
```
1ec5: Replace this class with a C struct to reduce overhead, and name it `MBSpeedLimit`. We can expect that the only fields it needs to contain are of type `double` and `MBSpeedUnit` (an enumeration), which can both be bridged from Objective-C to Swift. There would only be two reasons for making this a class: an instance method or a pointer-typed property, such as NSString.

Alternatively, drop support for iOS 8–9 and replace this class with the NSMeasurement class. `:daydream:`

MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -0,0 +1,118 @@
...
+    
+    /**
+     A unitless measure of speed which is dependent on the `MaximumSpeedLimit.speedUnits`.
+     
+     If a speed is not known, it will default to `Int.NSNotFound`.
+     */
+    @objc public var speed: Int = NSNotFound
```
1ec5: Speed limits can be non-integral, so `Float` or `Double` would be a more appropriate type. `NSNotFound` is only appropriate as a magic value for integral types, so use −1 instead.

1ec5: Rename this property to `value`. “Speed” is already part of the containing type’s name.

MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -0,0 +1,118 @@
...
+    
+    
+    /**
+     Units for `MaximumSpeedLimit.speed`.
+     */
+    @objc public var speedUnits: SpeedUnit = .none
```
1ec5: Rename this property to `unit`. “Speed” is already part of the containing type’s name.

MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -0,0 +1,118 @@
...
+    
+    
+    /**
+     `Bool` whether the maximum speed for a segment is known.
+     */
+    @objc public var speedIsUnknown: Bool = true
```
1ec5: A separate property would be a suboptimal way to represent an unknown value. `value == -1` is already sufficient to indicate that the speed limit is unknown. So we can formalize that fact: define a constant `MBSpeedUnknown` that has a `value` of –1. This is how `CLLocationCoordinate2D` and `kCLLocationCoordinate2DInvalid` work.

Down the line, if we need more symbolic values in the future, such as `variable`, then we can define an enumeration for those symbolic values and add a field of that type to the `Speed` struct. That would be useful for distinguishing between legal and advisory speed limits, for example.

MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -0,0 +1,118 @@
...
+    
+    /**
+     Indicates the road segment uses kilometers per hour for measuring speed limits.
+     */
+    case kilometersPerHour
+    
+    /**
+     Indicates the segment does not have a speed limit.
+    */
+    case none
```
1ec5: “No unit” is an unintuitive way to say “no value”. Remove this case and use `value == DBL_MAX` to represent an unlimited speed.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -186,6 +195,11 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     open let segmentCongestionLevels: [CongestionLevel]?
     
+    /**
+     An array containing the maximum speed limits for along each road segment in the route leg geometry.
```
1ec5: Typo:

> An array containing the maximum speed limit along each road segment along the route leg’s shape.

MapboxDirections/MBRouteLeg.swift
```diff
@@ -43,13 +44,19 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+                }
+            }
         }
         
         self.openStreetMapNodeIdentifiers = openStreetMapNodeIdentifiers
         self.segmentDistances = segmentDistances
         self.expectedSegmentTravelTimes = expectedSegmentTravelTimes
         self.segmentSpeeds = segmentSpeeds
         self.segmentCongestionLevels = congestionLevels
+        self.segmentMaximumSpeedLimits = maximumSpeedLimits
```
1ec5: I’m leaning towards calling this property `maximumSegmentSpeedLimits`, which would be more grammatical.

bsudekum: This sort of sounds like, the  _maximum number of segments_.

MapboxDirections/Attributes/MBSpeedLimit.swift
1ec5: On second thought, we’re mainly putting these objects in an NSArray, and C structs need to be boxed in NSValue before going into an NSArray. So that sort of defeats the purpose of making them structs. For now, let’s make the class as similar to NSMeasurement as possible and revisit using NSMeasurement once we drop iOS 8–9 support.

MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -1,5 +1,10 @@
 import Foundation
 
+/**
+ Speed value for a segment which is unknown.
+ */
+public let MBSpeedIsInvalid: Double = -1.0
```
1ec5: The suggestion in https://github.com/mapbox/MapboxDirections.swift/pull/250#discussion_r176269902 was to add a constant of type `MBSpeedLimit`, not a constant for the magic number alone. But we can only make it a constant if we make `MBSpeedLimit` a struct. Since we can’t, then we’ll need to create a class factory method on SpeedLimit instead:

```swift
class var invalid = SpeedLimit(value: -1, unit: .kilometersPerHour)
```

MapboxDirections/Attributes/MBSpeedLimit.swift
```diff
@@ -0,0 +1,110 @@
...
+    public init(value: Double, speedUnits: SpeedUnit) {
+        self.value = value
+        self.unit = speedUnits
+    }
+    
+    /**
+     Initialize a new `SpeedLimit` object from a JSON dictionary.
+     */
+    @objc public convenience init(json: [String: Any]) {
+        var speed = json["speed"] as? Double ?? SpeedLimit.invalid.value
```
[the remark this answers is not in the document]
1ec5: Use a literal `-1` here, then ensure in a test that the SpeedLimit that results from an invalid JSON value is equal to `SpeedLimit.invalid`.

1ec5:
The saga continues in #367.

1ec5:
I revived this PR because it already had tests and handled Autobahns correctly, unlike #367.

1ec5:
To do before merging:

* [ ] Merge #367
* [ ] Drop support for deploying backwards to iOS 9._x_, macOS 10.11._x_, tvOS 9._x_, and watchOS 2._x_: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r275962395
* [ ] Replace SpeedUnit with `Measurement<UnitSpeed>`: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r275962395
* [ ] Subclass `Measurement<UnitSpeed>` as `SignpostedSpeed`, factoring in the indicated sign style: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r277903175

1ec5:
The SPM build is failing with:

```
/Users/distiller/project/Sources/MapboxDirections/Extensions/Measurement.swift:3:11: error: 'Measurement' is only available on OS X 10.12 or newer
extension Measurement where UnitType: UnitSpeed {
          ^
```

which is weird, because I just merged in #379, which requires macOS 10.12 and above.

frederoni:
SPM doesn't depend on Xcode, it has its own place to specify [supported platforms](https://github.com/apple/swift-package-manager/blob/af308ac93ed01a5a5d8d4a66a12109ad9e8e0068/Sources/PackageDescription4/Package.swift#L278-L292) which I took the liberty of adding to this PR [here](https://github.com/mapbox/MapboxDirections.swift/pull/250/commits/318821c46d0ef4854a85dc687d4052cdb0939f37).

1ec5:
> * [ ] Convert each measurement to the unit of the corresponding step
> * [ ] Subclass `Measurement<UnitSpeed>` as `SignpostedSpeed`, factoring in the indicated sign style: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r277903175

Ideally MapboxDirections.swift would be responsible for cleaning up the model so that the speed limit attributes are readily usable in UI. Unfortunately, it’s a challenge to associate attributes with steps. Currently the navigation SDK associates congestion attributes with steps, but that’s memory intensive and a lot of inconvenient code.

1ec5:
87ee65908c9aa131c0b0a7f05d8a0fb35d4ff7b5 introduces a new property, `RouteLeg.segmentRangesByStep`, that makes it easier to get the elements in the `segment*` properties that correspond to a given step. That mostly addresses the concerns in https://github.com/mapbox/MapboxDirections.swift/pull/250#issuecomment-548067789. For speed limits, the navigation SDK can now use code like this to efficiently determine the speed limit to display:

```swift
let index = legProgress.currentStep.shape!.closestCoordinate(to: coordinate)!.index
let legProgress = routeProgress.currentLegProgress
let range = legProgress.leg.segmentRangesByStep[legProgress.stepIndex]
let speedUnit = legProgress.currentStep.speedLimitUnit
let speedLimit = legProgress.leg.segmentMaximumSpeedLimits[range][index].converted(to: speedUnit)
let speedLimitSignStandard = legProgress.currentStep.speedLimitSignStandard
```

PR #251, opened 2018-03-21
Release v0.19.0

bsudekum:
We should also add https://github.com/mapbox/MapboxDirections.swift/pull/250 to this release, a review would be great.

/cc @mapbox/navigation-ios


bsudekum:
Actually, this is ready now. Speeds in https://github.com/mapbox/MapboxDirections.swift/pull/250 will not be ready on the Directions API for roughly another week.

PR #252, opened 2018-04-02
Add exit, exitNumber component types

bsudekum:
Adda .exit, .exitNumber to MBVisualInstructionType.

/cc @1ec5 @JThramer

### DIFF
--- a/MapboxDirections/MBVisualInstructionType.swift
+++ b/MapboxDirections/MBVisualInstructionType.swift
@@ -25,13 +25,15 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
     
     /**
      The compoment contains the localized word for "exit".
+     
+     This component may appear before or after an `.exitNumber` component, depending on the language.
      */
     case exit
     
     /**
      A component contains an exit number.
      */
-    case exitNumber
+    case exitCode
     
     public init?(description: String) {
         let type: VisualInstructionComponentType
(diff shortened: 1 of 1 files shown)


MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -23,6 +23,16 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
      */
     case image
     
+    /**
+     The compoment contains the localized word for "exit".
+     */
```
1ec5: > This component may appear before or after an `.exitNumber` component, depending on the language.

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -23,6 +23,16 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
...
+    case exit
+    
+    /**
+     A component contains an exit number.
+     */
+    case exitNumber
```
1ec5: Rename this property to `exitCode` for consistency with `RouteStep.exitCode`:

https://github.com/mapbox/MapboxDirections.swift/blob/1711560166601f46466391a21da1ff79f2727323/MapboxDirections/MBRouteStep.swift#L882-L889

This property was so named to avoid confusion with a preexisting `exitIndex` and also because a freeway exit number may be alphabetic or alphanumeric.

PR #253, opened 2018-04-02
Make Lane initializer public

bsudekum:
Needed for https://github.com/mapbox/mapbox-navigation-ios/pull/1270#discussion_r178635704

/cc @1ec5 @frederoni

### DIFF
--- a/MapboxDirections/MBLane.swift
+++ b/MapboxDirections/MBLane.swift
@@ -10,7 +10,11 @@ public class Lane: NSObject, NSSecureCoding {
      */
     @objc public let indications: LaneIndication
     
-    internal init(indications: LaneIndication) {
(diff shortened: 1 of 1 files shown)


MapboxDirections/MBLane.swift
```diff
@@ -10,7 +10,11 @@ public class Lane: NSObject, NSSecureCoding {
      */
     @objc public let indications: LaneIndication
     
-    internal init(indications: LaneIndication) {
+    
+    /**
+     Initialize a new `Lane` from an array of  `LaneIndication`.
```
1ec5: `LaneIndication` is a bitmask, not an array.

> Initializes a new `Lane` using the given lane indications.

PR #257, opened 2018-04-06
Add allowsArrivingOnOppositeSide DirectionsOptions bool

bsudekum:
This adds a new option, `DirectionOptions.approachType`, for Indicating how a route considers from which side of the road to approach a waypoint.

/cc @1ec5 @coxchapman


MapboxDirections/MBApproachType.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+
+/**
+ Indicate how a route considers from which side of the road to approach a waypoint.
+ */
+@objc(MBApproachType)
+public enum ApproachType: UInt, CustomStringConvertible {
```
1ec5: “Type” is such an overloaded word that we should try very hard not to use it in any our Swift libraries’ public APIs. Some options:

* For the time being, the only two options imply that we could replace this enumeration with a simple `avoidsRoadCrossingOnArrival` or `allowsArrivingOnOppositeSide` flag.
* Call the enumeration `ApproachRestriction` and rename `unrestricted` to `none`.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -118,6 +124,13 @@ open class RouteOptions: DirectionsOptions {
      */
     @objc open var roadClassesToAvoid: RoadClasses = []
     
+    /**
+     Indicate how a route considers from which side of the road to approach a waypoint.
+     
+     By default, `ApproachType.unrestricted` is used meaning the route can approach a waypoint from either side of the road. If set to `ApproachType.curb`, the route will be returned so that on arrival, the waypoint will be found on the side that corresponds with the `DrivingSide` of the region in which the returned route is located.
```
1ec5: The fact that we end up documenting both of the enumeration’s cases here points strongly in the direction of replacing the enumeration with a simple Boolean-typed property. Additionally, this comment should explain _why_ the developer might want to set the option, not just what it does, and how this option differs from `allowsUTurnAtWaypoint`:

https://github.com/mapbox/MapboxDirections.swift/blob/038aa947fd137b9bd5c095eeb92cbaa0638fffab/MapboxDirections/MBRouteOptions.swift#L90

MapboxDirections/MBApproachType.swift
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     Approaches a waypoint from either side of road.
+     */
+    case unrestricted
+
+    /**
+     Approaches a waypoint from the same side of the road as the waypoint falls.
+     */
+    case curb
```
1ec5: A road has two curbs. This option forces the route to approach the curb closest to the waypoint. So rename this case `closerCurb` or `avoidsRoadCrossing`.

bsudekum:
@1ec5 updated to a simple bool.

bsudekum:
@1ec5 this is ready for review.

MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -404,6 +409,13 @@ open class DirectionsOptions: NSObject, NSSecureCoding, NSCopying {
      */
     @objc open var includesVisualInstructions = false
     
+    /**
+     Indicate how a route considers from which side of the road to approach a waypoint.
```
1ec5: This is a property, not an action method, so describe the effect rather than an action.

> A Boolean value that indicates whether each of the resulting routes must approach a waypoint from the same side of the road.

1ec5:
Contrary to https://github.com/mapbox/MapboxDirections.swift/pull/257#issue-180010648, we ended up adding a Boolean-typed `RouteOptions.allowsArrivingOnOppositeSide` property instead of the enumeration. In the future, if there are new arrival types besides the two currently supported by the Directions API, we can turn this property into a enumeration or add additional properties as appropriate.

PR #258, opened 2018-04-09
Add then component

bsudekum:
~This parses out the `then` text component for displaying information about the upcoming step.~

**UPDATES**
This parses out the `sub` component for displaying information about the upcoming step. Also, the new `lane` component type here provides additional lane information to assist visual representation. 

/cc @lyzidiamond @danpaz @mapbox/navigation-ios


MapboxDirections/MBVisualInstruction.swift
```diff
@@ -74,19 +88,32 @@ open class VisualInstruction: NSObject, NSSecureCoding {
             }
         }
         
-        self.init(distanceAlongStep: distanceAlongStep, primaryText: primaryText, primaryTextComponents: primaryTextComponents, secondaryText: secondaryText, secondaryTextComponents: secondaryTextComponents, drivingSide: drivingSide)
+        var thenText: String?
+        var thenTextComponents: [VisualInstructionComponent]?
+        if let thenTextComponent = json["then"] as? JSONDictionary {
+            thenText = thenTextComponent["text"] as? String
```
bsudekum: This key is being renamed to `sub`.

vincethecoder:
For future reference, here is a [link](https://api.mapbox.com/directions/v5/mapbox/driving-traffic/-77.015082%2C38.900279%3B-77.016183%2C38.900267.json?geometries=polyline&annotations=congestion&steps=true&overview=full&banner_instructions=true&access_token=pk.eyJ1IjoiZGFucGF0IiwiYSI6ImI0WThCVWMifQ.BP_j7Kiv5SNL-p-kTfXNxg) to a sample JSON response with a sub-component.

vincethecoder:
Sample JSON structure - for future reference:

screen shot 2018-05-16 at 2 12 03 pm

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -45,6 +45,20 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      */
     @objc public var abbreviationPriority: Int = NSNotFound
     
+    /**
+     An array indicating which directions you can go from a lane (left, right, or straight).
+     
+     If the value is `[.left", .straight]`, the driver can go straight or left from that lane. This is only set when the `component` is a lane.
+     */
+    @objc public var directions: [Any]?
```
vincethecoder: `Any` was used instead of `ManeuverDirection` -- since the `objc` declared property is an array of enum.

bsudekum: How about renaming this and making it a lane indication: https://github.com/mapbox/MapboxDirections.swift/blob/master/MapboxDirections/MBLaneIndication.h

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -82,13 +101,19 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
      */
-    @objc public init(type: VisualInstructionComponentType, text: String?, imageURL: URL?, abbreviation: String?, abbreviationPriority: Int) {
+    @objc public init(type: VisualInstructionComponentType, text: String?, imageURL: URL?, abbreviation: String?, abbreviationPriority: Int, directions: [String]? = nil, active: Bool = false) {
         self.text = text
         self.imageURL = imageURL
         self.type = type
         self.abbreviation = abbreviation
         self.abbreviationPriority = abbreviationPriority
+        if let directions = directions {
+            self.directions = directions.map { ManeuverDirection(description: $0) ?? .none }
```
bsudekum: I don't think `directions ` is an array of ManeuverDirctions. Try using the `LaneIndication` initializer here.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -45,6 +45,20 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
+     If the value is `[.left", .straight]`, the driver can go straight or left from that lane. This is only set when the `component` is a lane.
+     */
+    @objc public var directions: [Any]?
+    
+    /**
+     The boolean that indicates whether the component is a lane and can be used to complete the upcoming maneuver.
+     
+     If multiple lanes are active, then it used to complete the upcoming maneuver. This value is set to `false` by default.
+     */
+    @objc public var active: Bool = false
```
bsudekum: This property name does not really convey the link between it's name and whether or not the lane usable.

MapboxDirections/MBVisualInstructionComponent.swift
vincethecoder: @bsudekum yes (according to the recently published api-docs), when `type == .lane` we do have an array of directions. The suggested `LaneIndication` indicator could be a computed property as `type` is defined when initialized.

bsudekum: @vincethecoder I'm saying that internally in this library, these are LaneIndications and not ManeuverDirections.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -45,6 +45,20 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      */
     @objc public var abbreviationPriority: Int = NSNotFound
     
+    /**
+     An array indicating which directions you can go from a lane (left, right, or straight).
+     
+     If the value is `[.left", .straight]`, the driver can go straight or left from that lane. This is only set when the `component` is a lane.
+     */
+    @objc public var directions: LaneIndication
```
bsudekum: This should be an array of `LaneIndication`.

bsudekum: Let's also rename this `indications` to [follow other patterns](https://github.com/mapbox/MapboxDirections.swift/blob/42ec86382b180277394e3c1717c6c1992f86b84a/MapboxDirections/MBLane.swift#L11) in this lib.

vincethecoder: @bsudekum Yep, it's already an array... 😸

MapboxDirections/MBVisualInstructionBanner.swift
```diff
@@ -21,6 +21,11 @@ open class VisualInstructionBanner: NSObject, NSSecureCoding {
      */
     @objc public let secondaryInstruction: VisualInstruction?
     
+    /**
+     The information about the next maneuver (the one after the upcoming maneuver). This detail gives the driver heads up when the current step is short.
+     */
+    @objc public let subInstruction: VisualInstruction?
```
1ec5: Rename this property to `subinstruction` (and elsewhere in this PR). Precedents for this include `NSCompoundPredicate.subpredicates`.

MapboxDirections/MBVisualInstructionBanner.swift
```diff
@@ -21,6 +21,11 @@ open class VisualInstructionBanner: NSObject, NSSecureCoding {
      */
     @objc public let secondaryInstruction: VisualInstruction?
     
+    /**
+     The information about the next maneuver (the one after the upcoming maneuver). This detail gives the driver heads up when the current step is short.
```
1ec5: > A visual instruction that is presented simultaneously to provide information about an additional maneuver that occurs in rapid succession.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -45,6 +45,20 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      */
     @objc public var abbreviationPriority: Int = NSNotFound
     
+    /**
+     An array indicating which directions you can go from a lane (left, right, or straight).
+     
+     If the value is `[.left", .straight]`, the driver can go left or straight from that lane. This is only set when the `component` is a `lane`.
```
1ec5: There are a couple typos in the first sentence. How about:

> If the value contains both `LaneIndication.left` and `LaneIndication.straightAhead`…

1ec5: > This property is set to `nil` unless the `type` property is set to `VisualInstructionComponentType.lane`.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -45,6 +45,20 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
...
+     */
+    @objc public var indications: LaneIndication
+    
+    /**
+     The boolean that indicates whether the component is a lane and can be used to complete the upcoming maneuver.
+     
+     If multiple lanes are active, then they can all be used to complete the upcoming maneuver. This value is set to `false` by default.
+     */
+    @objc public var isLaneActive: Bool = false
```
1ec5: I think we should strive for more similarities with how lane indications are normally exposed. It’s poor API design to expose multiple mutually exclusive properties that are optionally set depending on each other. Consider making VisualInstructionComponent an umbrella class or protocol for subclasses such as LaneInstructionComponent – `isActive` can then be a property on that subclass.

1ec5: Rename this property to `isActiveLane`, or `isActive` if you end up refactoring the class. Also refine the getter and setter names in Objective-C: `-activeLane` should be the getter, and `-setActiveLane:` should be the setter.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -82,13 +107,17 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      - parameter imageURL: The URL to an image representation of this component.
      - parameter abbreviation: An abbreviated representation of `text`.
      - parameter abbreviationPriority: The priority for which the component should be abbreviated.
+     - parameter indications: The possibile directions to go from a lane component.
+     - parameter isLaneActive: The flag to indicate that the upcoming maneuver can be completed with a lane component.
      */
-    @objc public init(type: VisualInstructionComponentType, text: String?, imageURL: URL?, abbreviation: String?, abbreviationPriority: Int) {
+    @objc public init(type: VisualInstructionComponentType, text: String?, imageURL: URL?, abbreviation: String?, abbreviationPriority: Int, indications: LaneIndication = LaneIndication(), isLaneActive: Bool = false) {
```
1ec5: This is a lot of arguments for an initializer. Since these parameters aren’t always relevant, remove them and make the corresponding properties non-constant, so that the caller can set the properties after initializing the object.

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -6,6 +6,11 @@ import Foundation
 @objc(MBVisualInstructionComponentType)
 public enum VisualInstructionComponentType: Int, CustomStringConvertible {
     
+    /**
+     The step does not have a particular visual instruction component type associated with it.
```
1ec5: What is the developer supposed to do with a `none` instruction? Hide it?

MapboxDirections/MBVisualInstructionComponent.swift
1ec5: Better yet, instead of a bunch of lane instruction components, build up a new `Intersection` object that can have a bunch of `approachLanes` and `usableApproachLanes`.

MapboxDirections/MBVisualInstructionBanner.swift
vincethecoder: @1ec5 I understand your preference for `subinstruction` based off, `NSCompoundPredicate.subpredicates`. In this context, it feels out of place when sub instruction variable reads differently from, `primaryInstruction` and `secondaryInstruction`. Probably, we should go for consistency here. 
https://github.com/mapbox/MapboxDirections.swift/pull/258/files#diff-8aa869dddbf0e52feb01bd7652980dfbL42

MapboxDirections/MBVisualInstructionComponent.swift
vincethecoder: @1ec5 FYI, `LaneIndication` is an array of lane indications. For obj-C compatibility, we shouldn't set this to an optional array.

MapboxDirections/MBVisualInstructionComponent.swift
vincethecoder: @1ec5 I don't think that's necessary. We could create a convenience initializer which accepts these additional parameters. Since `sub` components have a distinct visual instruction structure, we can get away with a convenience initializer in this case.

MapboxDirections/MBVisualInstructionComponent.swift
1ec5: LaneIndication is an OptionSet, not an array. It just happens that Swift uses the same syntax for option sets as it does for arrays. We should say the property is set to the empty set (or 0 in Objective-C) in this case, and we should describe the property as an option set, not an array.

MapboxDirections/MBVisualInstructionBanner.swift
1ec5: “Sub” isn’t a word.

vincethecoder: @1ec5 okay, if that's your notion, then we're better off with a readable word like `tertiary` --> `tertiaryInstruction`. Which makes this consistent.

MapboxDirections/MBVisualInstructionComponent.swift
vincethecoder: Okay. Noted. In which case, we will indicate that this OptionSet is by default an empty set. Hence, we keep this as a type `isActiveLane` as a type `LaneIndication`.

MapboxDirections/MBComponent.swift
```diff
@@ -0,0 +1,45 @@
+import Foundation
+
+/**
+The parent component that comprises what should be displayed in the instruction banner.
+ */
+@objc(MBComponent)
+open class Component: NSObject, NSSecureCoding {
```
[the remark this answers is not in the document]
vincethecoder: @bsudekum we can make it a protocol.

MapboxDirections/MBComponent.swift
```diff
@@ -1,45 +1,16 @@
...
 /**
-The parent component that comprises what should be displayed in the instruction banner.
+The component protocol that comprises what the instruction banner should display.
  */
 @objc(MBComponent)
-open class Component: NSObject, NSSecureCoding {
-    
-    open static var supportsSecureCoding: Bool = true
-    
+public protocol Component: class, NSSecureCoding {
```
1ec5: Component is a pretty generic term, and a protocol’s name should describe some quality about its conforming classes. Since the two conforming classes are VisualInstructionComponent and LaneIndicationComponent, both of which are intended for visual display, let’s call the protocol something like Visualizable. (This isn’t a great name – any better ideas?)

MapboxDirections/MBComponent.swift
```diff
@@ -1,45 +1,16 @@
...
     /**
      The plain text representation of this component.
      */
-    @objc public let text: String?
+    @objc var text: String? { get }
```
1ec5: This property only makes sense for VisualInstructionComponent, not for LaneIndicationComponent, so remove it from this protocol (but keep it as part of VisualInstructionComponent).

MapboxDirections/MBComponent.swift
```diff
@@ -1,45 +1,16 @@
...
-            return nil
-        }
-        self.type = type
-    }
-    
-    @objc public func encode(with coder: NSCoder) {
-        coder.encode(text, forKey: "text")
-        coder.encode(type, forKey: "type")
-    }
+    @objc var type: VisualInstructionComponentType { get }
```
1ec5: Similarly, this property should be removed from the protocol because it’s only applicable to visual instruction components.

MapboxDirections/MBComponent.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+/**
+The component protocol that comprises what the instruction banner should display.
+ */
+@objc(MBComponent)
+public protocol Component: class, NSSecureCoding {
```
1ec5: Component is a pretty generic term, and a protocol’s name should describe some quality about its conforming classes. Since the two conforming classes are VisualInstructionComponent and LaneIndicationComponent, both of which are intended for visual display, let’s call the protocol something like Visualizable. (This isn’t a great name – any better ideas?)

MapboxDirections/MBComponent.swift
```diff
@@ -0,0 +1,16 @@
...
+    /**
+     The plain text representation of this component.
+     */
+    @objc var text: String? { get }
```
1ec5: This property only makes sense for VisualInstructionComponent, not for LaneIndicationComponent, so remove it from this protocol (but keep it as part of VisualInstructionComponent).

MapboxDirections/MBComponent.swift
```diff
@@ -0,0 +1,16 @@
...
+    
+    /**
+     The type of visual instruction component. You can display the component differently depending on its type.
+     */
+    @objc var type: VisualInstructionComponentType { get }
```
1ec5: Similarly, this property should be removed from the protocol because it’s only applicable to visual instruction components.

MapboxDirections/MBLaneIndicationComponent.swift
```diff
@@ -0,0 +1,67 @@
...
+     The boolean that indicates whether the component is a lane and can be used to complete the upcoming maneuver.
+     
+     If multiple lanes are active, then they can all be used to complete the upcoming maneuver. This value is set to `false` by default.
+     */
+    @objc public var isUsable: Bool = false
+    
+    // MARK: Component/NSSecureCoding Protocols Variables
+    @objc public var text: String?
+    
+    @objc public var type: VisualInstructionComponentType
```
1ec5: These two properties would be deleted.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -56,15 +62,57 @@ open class VisualInstruction: NSObject, NSSecureCoding {
...
+                }
+            } else {
+                components = dictionary.map { record in
+                    let type = VisualInstructionComponentType(description: record["type"] as? String ?? "") ?? .text
+                    let imageURL = URL(string: (record["imageBaseURL"] as? String) ?? "")
+                    return VisualInstructionComponent(type: type,
+                                                      text: record["text"] as? String,
+                                                  imageURL: imageURL,
+                                              abbreviation: record["abbr"] as? String,
+                                      abbreviationPriority: record["abbr_priority"] as? Int ?? NSNotFound)
```
1ec5: Only align parameters by the colon in Objective-C, not in Swift. Use Xcode’s reindent command to reformat this code.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -11,20 +9,7 @@ import Foundation
 /**
  A component of a `VisualInstruction` that represents a single run of similarly formatted text or an image with a textual fallback representation.
  */
-@objc(MBVisualInstructionComponent)
```
1ec5: This class should remain bridgeable to Objective-C.

MapboxDirections/MBLaneIndicationComponent.swift
```diff
@@ -0,0 +1,67 @@
+/**
+A component that represents a lane  representation of an instruction.
+ */
+open class LaneIndicationComponent: Component {
```
1ec5: This class needs to be bridgeable to Objective-C, with a class name that begins with the library’s class prefix `MB`.

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -30,6 +30,11 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
      */
     case exit
     
+    /**
+     The component contains information about which lanes can be used to complete the maneuver.
+     */
+    case lane
```
1ec5: This case becomes unnecessary because LaneIndicationComponent no longer has a `type` property.

1ec5:
Also, please update the PR description with a summary of the API changes contained in the PR. Thanks!

MapboxDirections/MBComponent.swift
vincethecoder: Probably, something like `ComponentRepresentable`?

MapboxDirections/MBLaneIndicationComponent.swift
```diff
@@ -0,0 +1,50 @@
...
+        
+        guard let active = decoder.decodeObject(forKey: "active") as? Bool else {
+            return nil
+        }
+        self.isUsable = active
+    }
+    
+    public func encode(with coder: NSCoder) {
+        coder.encode(indications, forKey: "directions")
+        coder.encode(isUsable, forKey: "active")
```
bsudekum: Nit, we've followed a convention of naming the key the same name as the variables used. Rename this to something like: `coder.encode(isUsable, forKey: "isUsable")` for consistency.

MapboxDirections/MBVisualInstructionBanner.swift
```diff
@@ -85,6 +98,7 @@ open class VisualInstructionBanner: NSObject, NSSecureCoding {
         coder.encode(distanceAlongStep, forKey: "distanceAlongStep")
         coder.encode(primaryInstruction, forKey: "primary")
         coder.encode(secondaryInstruction, forKey: "secondary")
+        coder.encode(tertiaryInstruction, forKey: "sub")
```
bsudekum: Same thing here, `coder.encode(tertiaryInstruction, forKey: "tertiaryInstruction")`.

MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -45,6 +31,13 @@ open class VisualInstructionComponent: NSObject, NSSecureCoding {
      */
     @objc public var abbreviationPriority: Int = NSNotFound
     
+    // MARK: Component/NSSecureCoding Protocols Variables
+    @objc public var text: String?
+    
+    @objc public var type: VisualInstructionComponentType
```
bsudekum: These properties could use some documentation.

MapboxDirections/MBComponentRepresentable.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+/**
+The component representable protocol that comprises what the instruction banner should display.
+ */
+@objc(MBComponentRepresentable)
+public protocol MBComponentRepresentable: class, NSSecureCoding { }
```
JThramer: Type prefixes in swift are somewhat of an anti-pattern, unless your team has a special reason (and convention) for using them. They were [a convention](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingBasics.html#//apple_ref/doc/uid/20001281-1002226-BBCJECED) in Objective-C because all namespaces in that language are global, but with swift's introduction of [modules](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID4), this pattern is more or less deprecated. This is because even if you call your class `UIViewController`, `MyApp.UIViewController()` is different from `UIKit.UIViewController()`.

 Naming the OBJ-C bridge `MBComponentRepresentable` is a good call though and should be kept.

MapboxDirections/MBVisualInstructionBanner.swift
```diff
@@ -21,6 +21,11 @@ open class VisualInstructionBanner: NSObject, NSSecureCoding {
      */
     @objc public let secondaryInstruction: VisualInstruction?
     
+    /**
+     A visual instruction that is presented simultaneously to provide information about an additional maneuver that occurs in rapid succession.
+     */
+    @objc public let tertiaryInstruction: VisualInstruction?
```
JThramer: This is good (abstracted because it's _either_ a lane, _or_ a turn component), but it may be nice to specifically call out how this property is used in our current implementation.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -56,15 +56,39 @@ open class VisualInstruction: NSObject, NSSecureCoding {
...
     public convenience init(json: [String: Any]) {
         let text = json["text"] as? String
-        let maneuverType = ManeuverType(description: json["type"] as! String) ?? .none
-        let maneuverDirection = ManeuverDirection(description: json["modifier"] as! String)  ?? .none
-        let textComponents = (json["components"] as! [JSONDictionary]).map {
-            VisualInstructionComponent(json: $0)
+        let componentsDictionary = json["components"] as? [JSONDictionary]
+        var components = [MBComponentRepresentable]()
+        
+        var maneuverType: ManeuverType = .none
```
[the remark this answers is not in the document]
vincethecoder: @JThramer this pre-existing and basically refactored from https://github.com/mapbox/MapboxDirections.swift/pull/258/files/916531f27e054235e599a9f6fb9e9a8955533655#diff-361a2ac3a3f3ee2f7a0fe4bbdd15f1deL59

The deduction from the existing logic is, even if we provided a fail-able initializer, the nil value will also always return `.none` -- so it's a choice between being proactive vs. reactive.

MapboxDirections/MBLaneIndicationComponent.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     An array indicating which directions you can go from a lane (left, right, or straight).
+     
+     If the value is `[LaneIndication.left", LaneIndication.straightAhead]`, the driver can go left or straight ahead from that lane. This is only set when the `component` is a `lane`.
+     */
+    @objc public var indications: LaneIndication
+    
+    /**
+     The boolean that indicates whether the component is a lane and can be used to complete the upcoming maneuver.
```
bsudekum: I think since this is a LaneIndicationComponent, we already know it's a lane. I think this is more about whether the lane can be used to complete the maneuver.

MapboxDirections/MBLaneIndicationComponent.swift
```diff
@@ -0,0 +1,50 @@
...
+        self.indications = indications
+        
+        guard let isUsable = decoder.decodeObject(forKey: "isUsable") as? Bool else {
+            return nil
+        }
+        self.isUsable = isUsable
+    }
+    
+    public func encode(with coder: NSCoder) {
+        coder.encode(indications, forKey: "directions")
```
bsudekum: Rename this to `coder.encode(indications, forKey: "indications")`

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -56,15 +56,39 @@ open class VisualInstruction: NSObject, NSSecureCoding {
     @objc(initWithJSON:)
     public convenience init(json: [String: Any]) {
         let text = json["text"] as? String
-        let maneuverType = ManeuverType(description: json["type"] as! String) ?? .none
-        let maneuverDirection = ManeuverDirection(description: json["modifier"] as! String)  ?? .none
-        let textComponents = (json["components"] as! [JSONDictionary]).map {
-            VisualInstructionComponent(json: $0)
+        let componentsDictionary = json["components"] as? [JSONDictionary]
```
bsudekum: Inline this line into the if let down below instead of pulling this out into a variable.

MapboxDirections/MBVisualInstructionBanner.swift
```diff
@@ -21,6 +21,13 @@ open class VisualInstructionBanner: NSObject, NSSecureCoding {
      */
     @objc public let secondaryInstruction: VisualInstruction?
     
+    /**
+     A visual instruction that is presented simultaneously to provide information about an additional maneuver that occurs in rapid succession.
+     
+     This instruction could either contain the visual layout information or the lane information about the next maneuver after the upcoming maneuver.
```
bsudekum: `This instruction could either contain the visual layout information or the lane information about the upcoming maneuver.`

MapboxDirectionsTests/IntructionsTests.swift
```diff
@@ -83,22 +83,37 @@ class SpokenInstructionsTests: XCTestCase {
...
+        let laneIndicationComponent = visualInstructions?.first?.tertiaryInstruction?.components.first as! LaneIndicationComponent
+        XCTAssertNotNil(visualInstructions)
+        XCTAssertEqual(visualInstructions?.first?.tertiaryInstruction?.text, "")
+        XCTAssertEqual(visualInstructions?.first?.distanceAlongStep, 120.7)
+        XCTAssertEqual(visualInstructions?.first?.tertiaryInstruction?.finalHeading, 180.0)
+        XCTAssertEqual(visualInstructions?.first?.tertiaryInstruction?.maneuverType, .reachFork)
+        XCTAssertEqual(visualInstructions?.first?.tertiaryInstruction?.maneuverDirection, .right)
+        XCTAssertEqual(laneIndicationComponent.isUsable, true)
+        XCTAssertEqual(laneIndicationComponent.indications, [.left, .straightAhead])
+        XCTAssertEqual(visualInstructions?.first?.drivingSide, .right)
```
bsudekum: So does this step have both lane information and then information? I think the implementation on the server prevents this from ever occurring.

vincethecoder: Will re-check the JSON response and relay my finding.

MapboxDirectionsTests/IntructionsTests.swift
```diff
@@ -84,21 +84,90 @@ class SpokenInstructionsTests: XCTestCase {
...
+            XCTAssertEqual(task.state, .completed)
+        }
+        
+        XCTAssertNotNil(route)
+        XCTAssertEqual(route!.routeIdentifier, "cjikck25m00v279ms1knttdgc")
+        
+        let step = route!.legs.first!.steps.first!
+        let visualInstructions = step.instructionsDisplayedAlongStep
+        
+        let tertiaryInstruction = visualInstructions?.first?.tertiaryInstruction
```
bsudekum: @vincethecoder getting close! I think the final piece is to now add a test for a tertiaryInstruction where it's properties are non-nil. IE test for a then instruction instead of a lane instruction.

vincethecoder: @bsudekum accounted for here: https://github.com/mapbox/MapboxDirections.swift/pull/258/files#diff-c645476c0ab60998ed8e475352ee8bdcR174

PR #259, opened 2018-04-10
Expose banner degrees

frederoni:
Exposes degrees on banner instructions (https://www.mapbox.com/api-documentation/#banner-instruction-object)

This PR also refactors the structure of the visual instructions in order to resemble the api response more accurately and make it easier to add new components.

<details>
Before:

```json
{
  "primaryText": "primaryText",
  "secondaryText": "secondaryText",
  "primaryTextComponents": [],
  "secondaryTextComponents": []
}
```

After:
```json
VisualInstructionBanner: {
    "primaryInstruction": VisualInstruction {
        "textComponents": [VisualInstructionComponent]
    }
}
```

</details>
<br>

@bsudekum 👀


frederoni:
@1ec5 @bsudekum do you remember the reasoning behind why we’d flatten

```json
{
  "primary": {
    "text": "primaryText",
    "components": []
  },
  "secondary": {
    "text": "secondaryText",
    "components": []
  }
}
```

into

```json
{
  "primaryText": "primaryText",
  "secondaryText": "secondaryText",
  "primaryTextComponents": [],
  "secondaryTextComponents": []
}
```

If we’d follow the same structure, we should probably add a `primaryDegrees` and a `secondaryDegrees` but I'd rather use the nested structure that api-directions is using.
On the other hand, we would only use the degrees of the primary instruction so mapping it to its visual instruction would also be somewhat correct.

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -2,127 +2,99 @@ import Foundation
...
+     Initialize A `VisualInstructionBanner`.
      */
-    @objc public init(distanceAlongStep: CLLocationDistance, primaryText: String, primaryTextComponents: [VisualInstructionComponent], secondaryText: String?, secondaryTextComponents: [VisualInstructionComponent]?, drivingSide: DrivingSide) {
-        self.distanceAlongStep = distanceAlongStep
-        self.primaryText = primaryText
-        self.primaryTextComponents = primaryTextComponents
-        self.secondaryText = secondaryText
-        self.secondaryTextComponents = secondaryTextComponents
-        self.drivingSide = drivingSide
+    @objc public init(text: String?, maneuverType: ManeuverType, maneuverDirection: ManeuverDirection, textComponents: [VisualInstructionComponent], degrees: CLLocationDegrees = 180) {
```
bsudekum: Is a default degree necessary?

bsudekum:
Looking good. For prosperity, noting that we'll need to continue to pin to the third digit here: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation.podspec#L43

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -2,127 +2,99 @@ import Foundation
...
-            let secondaryManeuverType = ManeuverType(description: secondaryTextComponent["type"] as! String) ?? .none
-            let secondaryManeuverDirection = ManeuverDirection(description: secondaryTextComponent["modifier"] as! String)  ?? .none
-            secondaryTextComponents = (secondaryTextComponent["components"] as! [JSONDictionary]).map {
-                VisualInstructionComponent(maneuverType: secondaryManeuverType, maneuverDirection: secondaryManeuverDirection, json: $0)
-            }
-        }
-        
-        self.init(distanceAlongStep: distanceAlongStep, primaryText: primaryText, primaryTextComponents: primaryTextComponents, secondaryText: secondaryText, secondaryTextComponents: secondaryTextComponents, drivingSide: drivingSide)
-    }
+    @objc public var degrees: CLLocationDegrees = 180
```
bsudekum: Why not make this optional and nil by default?

MapboxDirections/MBVisualInstruction.swift
frederoni: If it wasn't for ObjC bridging, I'd prefer `degrees: CLLocationDegrees? = nil` but 180 seems appropriate for default value as it indicates `.straightAhead`.

MapboxDirections/MBVisualInstruction.swift
frederoni: See https://github.com/mapbox/MapboxDirections.swift/pull/259#discussion_r180805941

MapboxDirections/MBVisualInstruction.swift
```diff
@@ -2,127 +2,103 @@ import Foundation
...
+    @objc public let textComponents: [VisualInstructionComponent]
     
     /**
      :nodoc:
-     Indicates what side of a bidirectional road the driver must be driving on. Also referred to as the rule of the road.
+     The degrees at which you will be exiting a roundabout, assuming 180 indicates going straight through the roundabout.
+     Note that this property is irrelevant unless the `maneuverType` is
      */
-    @objc public var drivingSide: DrivingSide
+    @objc public var degrees: CLLocationDegrees = 180
```
1ec5: Degrees is the unit of measurement, which is already implied by `CLLocationDegrees`; `heading` would be a better name, more consistent with the rest of the API.

1ec5: Yes, `finalHeading` sounds good: #265.

PR #263, opened 2018-04-13
Update podspec, readme for v0.20.0

1ec5:
This is a followup to #260 that updates the CocoaPods podspec and readme for v0.20.0.

/cc @frederoni @akitchen


1ec5:
Duplicate of #261.

PR #267, opened 2018-04-17
Automate one step in documentation generation

### LINKED ISSUE #262: Documentation generation script should update redirect
[This documentation](https://github.com/mapbox/MapboxDirections.swift/blob/fb62bdcab91e87d865ddb0f75c4bfd3f9bd8e696/docs/README.md) covers the steps required to generate documentation for a particular release. Unfortunately, this step is often forgotten:

https://github.com/mapbox/MapboxDirections.swift/blob/fb62bdcab91e87d865ddb0f75c4bfd3f9bd8e696/docs/README.md#L8

[document.sh](https://github.com/mapbox/MapboxDirections.swift/blob/master/scripts/document.sh) should have a flag that automatically updates these files.

/cc @frederoni @captainbarbosa

frederoni:
Fixes #262 

Simplifies the documentation generation by removing one step.

@1ec5

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,5 +5,4 @@ Run `scripts/document.sh`
 ## Update the documentation site:
 1. Clone [mapbox-navigation-ios](https://github.com/mapbox/mapbox-navigation-ios) to a mapbox-navigation-ios-docs folder alongside your MapboxDirections.swift clone, and check out the `mb-pages` branch.
 1. In your main MapboxDirections.swift clone, check out the release branch and run `OUTPUT=../mapbox-navigation-ios-docs/directions/X.X.X scripts/document.sh`, where _X.X.X_ is the new SDK version.
-1. In mapbox-navigation-ios-docs, edit [directions/index.html](https://github.com/mapbox/mapbox-navigation-ios/blob/mb-pages/directions/index.html) and directions/docsets/Mapbox.xml to refer to the new SDK version.
 1. Commit and push your changes to the mapbox-navigation-ios `mb-pages` branch.
--- a/scripts/document.sh
+++ b/scripts/document.sh
@@ -41,3 +41,7 @@ jazzy \
 
 find ${OUTPUT} -name *.html -exec \
     perl -pi -e 's/BRANDLESS_DOCSET_TITLE/Directions.swift $1/, s/MapboxDirections.swift\s+(Docs|Reference)/MapboxDirections.swift $1/' {} \;
+
+# Replace version numbers
+sed -i '' -e 's|url=[^0-9.]*\([0-9.]*\)|url='${RELEASE_VERSION}'|g' ${OUTPUT}/../index.html
+sed -i '' -e 's|[0-9.]\([0-9.]\)\([0-9]\)|{x}|g; s|{x}{x}|'${RELEASE_VERSION}'|g' ${OUTPUT}/../docsets/MapboxDirections.xml


scripts/document.sh
```diff
@@ -41,3 +41,7 @@ jazzy \
 
 find ${OUTPUT} -name *.html -exec \
     perl -pi -e 's/BRANDLESS_DOCSET_TITLE/Directions.swift $1/, s/MapboxDirections.swift\s+(Docs|Reference)/MapboxDirections.swift $1/' {} \;
+
+# Replace version numbers
+sed -i '' -e 's|url=[^0-9.]*\([0-9.]*\)|url='${RELEASE_VERSION}'|g' ${OUTPUT}/../index.html
+sed -i '' -e 's|[0-9.]\([0-9.]\)\([0-9]\)|{x}|g; s|{x}{x}|'${RELEASE_VERSION}'|g' ${OUTPUT}/../docsets/MapboxDirections.xml
```
1ec5: Sometimes we publish prereleases, which need published documentation but shouldn’t be featured as the current version just yet. Perhaps detect a prerelease and bail, or add an option to disable updating the redirect?

docs/README.md
```diff
@@ -5,5 +5,4 @@ Run `scripts/document.sh`
 ## Update the documentation site:
 1. Clone [mapbox-navigation-ios](https://github.com/mapbox/mapbox-navigation-ios) to a mapbox-navigation-ios-docs folder alongside your MapboxDirections.swift clone, and check out the `mb-pages` branch.
 1. In your main MapboxDirections.swift clone, check out the release branch and run `OUTPUT=../mapbox-navigation-ios-docs/directions/X.X.X scripts/document.sh`, where _X.X.X_ is the new SDK version.
```
1ec5: Maybe we should refactor the script so that `OUTPUT` is the path to directions/ and the version subfolder is determined automatically.

frederoni:
@1ec5 this PR is ready for round two.

scripts/document.sh
```diff
@@ -15,13 +15,11 @@ fi
 
 
 OUTPUT=${OUTPUT:-documentation}
+DEFAULT_VERSION="x.x.x"
```
1ec5: Can the version default to something more accurate, like the `git describe` thing we used to use?

scripts/document.sh
```diff
@@ -32,12 +30,32 @@ jazzy \
...
-find ${OUTPUT} -name *.html -exec \
+find "${OUTPUT}/${VERSION}" -name *.html -exec \
     perl -pi -e 's/BRANDLESS_DOCSET_TITLE/Directions.swift $1/, s/MapboxDirections.swift\s+(Docs|Reference)/MapboxDirections.swift $1/' {} \;
+
+function parseSemver() {
+    local RE='[^0-9]*\([0-9]*\)[.]\([0-9]*\)[.]\([0-9]*\)\([0-9A-Za-z-]*\)'
+    eval $2=`echo $1 | sed -e "s#$RE#\1#"` # major
+    eval $3=`echo $1 | sed -e "s#$RE#\2#"` # minor
+    eval $4=`echo $1 | sed -e "s#$RE#\3#"` # patch
+    eval $5=`echo $1 | sed -e "s#$RE#\4#"` # prerelease
```
1ec5: This is intense. 😀 I guess I would’ve tried to use a Bash array for this, but as long as the string being evaluated is coming purely from the script, then this should be safe.

Issue #270, opened 2018-04-20
Remove AttributeOptions.openStreetMapNodeIdentifier


1ec5:
`AttributeOptions.openStreetMapNodeIdentifier` was added and documented back in #118 when the Directions API’s `annotation` parameter was still in flux: https://github.com/mapbox/MapboxDirections.swift/pull/118#discussion_r110547410. The parameter no longer accepts the `nodes` value, so we should remove this `AttributeOptions` case.

/cc @bsudekum @danpat @karenzshea

1ec5:
Fixed in #272.

PR #272, opened 2018-04-23
Remove Node Identifier Attribute

JThramer:
Implements #270. I'm somewhat unclear about how to document this change, as there is no `CHANGELOG.MD` in this project.

/cc @mapbox/navigation-ios


MapboxDirections/MBAttribute.h
```diff
@@ -3,36 +3,29 @@
 /**
  Attributes are metadata information for a route leg.
  
- When most of the attributes are specified, the resulting route leg contains one attribute value for each segment in leg, where a segment is the straight line between two coordinates in the route leg’s full geometry. When the `MBAttributeOpenStreetMapNodeIdentifier` attribute is specified, the corresponding `RouteLeg` property contains one more value than each of the other attributes.
+ When most of the attributes are specified, the resulting route leg contains one attribute value for each segment in leg, where a segment is the straight line between two coordinates in the route leg’s full geometry.
```
1ec5: Replace “most of the attributes are specified” with “any of the attributes is specified”, now that there’s no longer an exception to the rule.

Issue #271, opened 2018-04-23
DirectionsOptions should include waypoint names in URL parameter


1ec5:
DirectionsOptions.params should add a `waypoint_names` parameter based on all the `Waypoint` objects’ `name` properties.

/ref mapbox/mapbox-navigation-ios#1123
/cc @bsudekum

1ec5:
Fixed in #273.

PR #273, opened 2018-05-08
New query parameter, waypoint names added to directions options.

vincethecoder:
`DirectionsOptions.params` getter now has a new URLQueryItem added to the array based on mapping all the waypoints to their names.

Addresses #271

/ref mapbox/mapbox-navigation-ios#1123
/cc @1ec5, @bsudekum


MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -444,6 +444,11 @@ open class DirectionsOptions: NSObject, NSSecureCoding, NSCopying {
             params.append(URLQueryItem(name: "annotations", value: attributesStrings))
         }
         
+        if !waypoints.compactMap({ $0.name }).isEmpty {
+            let names = waypoints.map { $0.name ?? "" }.joined(separator: ";")
```
bsudekum: Would it be possible to do this map and compactMap once?

1ec5: No, because an intermediate waypoint can have no name, but it would still need an empty entry in the semicolon-delimited list.

PR #275, opened 2018-05-10
Changing MBVisualInstructionComponentType.image -> .icon

JThramer:
Updating `MBVisualInstructionComponentType` to reflect new changes on API. This is to reflect that the icon component may have only text, and thus should be rendered as a generic route shield. Supports mapbox/mapbox-navigation-ios#1190.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxDirections/MBVisualInstructionType.swift
+++ b/MapboxDirections/MBVisualInstructionType.swift
@@ -19,9 +19,9 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
     case text
     
     /**
-     Component contains an image that should be rendered.
+     Component either contains an image that should be rendered, or text that should be rendered as a generic icon.
      */
-    case image
+    case icon
     
     /**
      The compoment contains the localized word for "exit".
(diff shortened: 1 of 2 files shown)


bsudekum:
@JThramer this looks like just the enum is changing here, is that what you are going for?

MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -19,9 +19,9 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
     case text
     
     /**
-     Component contains an image that should be rendered.
+     Component either contains an image that should be rendered, or text that should be rendered as a generic icon.
      */
-    case image
+    case icon
```
1ec5: I’m not sure this change is strictly necessary. “Image” and “icon” are roughly synonymous in this context. We went with `image` for consistency with the `VisualInstructionComponent.imageURL` property, which is still present regardless of the API change.

JThramer:
@bsudekum @1ec5 The idea here in my mind was that it was kinda silly to have a component type with the name of `image` if it may not actually contain an image.

JThramer:
Consensus was gained that this change is superfluous. Closing.

PR #281, opened 2018-06-21
Decode polyline6

frederoni:
We're decoding a geometry string as `polyline` even if we specify `polyline6` in the route options.

@bsudekum @1ec5


MapboxDirections/MBRoute.swift
```diff
@@ -38,7 +38,8 @@ open class Route: DirectionsResult {
         case let geometry as JSONDictionary:
             coordinates = CLLocationCoordinate2D.coordinates(geoJSON: geometry)
         case let geometry as String:
-            coordinates = decodePolyline(geometry, precision: 1e5)!
+            let precision = routeOptions.shapeFormat == .polyline ? 1e5 : 1e6
```
frederoni: Probably better to switch on `routeOptions.shapeFormat`?

bsudekum: Yeah, probably. Although the line above, `case let geometry as String:` takes care of the case where it's geojson.

frederoni: We need to plumb this through `MBRouteStep` as well. 🤔

frederoni:
Ready for review

PR #288, opened 2018-08-23
Added allowsArrivingOnOppositeSide

### LINKED ISSUE #280: Support approaches URL query parameter
The Directions API has an `approaches` URL query parameter that should be exposed as an option in this library. In #257, we attempted to expose this option as a single flag on RouteOptions. However, we ended up reverting that implementation in #279, because each waypoint can have a distinct value in `approaches`. Instead, `allowsArrivingOnOppositeSide` should be a flag on Waypoint.

/cc @bsudekum

frederoni:
Fixes #280 support `approaches` URL query parameter.

/cc @1ec5 @akitchen

### DIFF
--- a/MapboxDirections/MBWaypoint.swift
+++ b/MapboxDirections/MBWaypoint.swift
@@ -70,6 +70,7 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
         heading = decoder.decodeDouble(forKey: "heading")
         headingAccuracy = decoder.decodeDouble(forKey: "headingAccuracy")
         name = decoder.decodeObject(of: NSString.self, forKey: "name") as String?
+        allowsArrivingOnOppositeSide = decoder.decodeBool(forKey: "allowsArrivingOnOppositeSide")
     }
     
     open func encode(with coder: NSCoder) {
@@ -79,6 +80,7 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
         coder.encode(heading, forKey: "heading")
         coder.encode(headingAccuracy, forKey: "headingAccuracy")
         coder.encode(name, forKey: "name")
+        coder.encode(allowsArrivingOnOppositeSide, forKey: "allowsArrivingOnOppositeSide")
     }
     
     open func copy(with zone: NSZone?) -> Any {
@@ -144,6 +146,13 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
      This parameter does not affect the route, but you can set the name of a waypoint you pass into a `RouteOptions` object to help you distinguish one waypoint from another in the array of waypoints passed into the completion handler of the `Directions.calculate(_:completionHandler:)` method.
      */
     @objc open var name: String?
+    
+    /**
+     A boolean value indicating whether arriving on opposite side is allowed.
+     
+     This property has no effect if `RouteOptions.includesSteps` is set to `false`.
+     */
+    @objc open var allowsArrivingOnOppositeSide = true
     
     @objc open override var description: String {
         return name ?? "<latitude: \(coordinate.latitude); longitude: \(coordinate.longitude)>"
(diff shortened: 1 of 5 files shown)


MapboxDirections/MBWaypoint.swift
```diff
@@ -17,11 +17,13 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
         
         It is recommended that the value of this parameter be greater than the `horizontalAccuracy` property of a `CLLocation` object obtained from a `CLLocationManager` object. There is a high likelihood that the user may be located some distance away from a navigable road, for instance if the user is currently on a driveway or inside a building.
      - parameter name: The name of the waypoint. This parameter does not affect the route but may help you to distinguish one waypoint from another.
+     - parameter allowsArrivingOnOppositeSide: A boolean value indicating whether arriving on opposite side is allowed.
      */
-    @objc public init(coordinate: CLLocationCoordinate2D, coordinateAccuracy: CLLocationAccuracy = -1, name: String? = nil) {
+    @objc public init(coordinate: CLLocationCoordinate2D, coordinateAccuracy: CLLocationAccuracy = -1, name: String? = nil, allowsArrivingOnOppositeSide: Bool = true) {
```
frederoni: Should we have a convenience initializer to make Obj-C less verbose by default?

akitchen: Seems like a good idea 👍

MapboxDirections/MBWaypoint.swift
```diff
@@ -9,6 +9,16 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
         return true
     }
     
+    /// :nodoc:
```
1ec5: Why aren’t these initializers documented? Because they’re only available for Objective-C? I think we have some examples of that elsewhere in MapboxDirections.swift or MapboxStatic.swift, where we still document the Objective-C-only methods but say they’re Objective-C-only.

MapboxDirections/MBWaypoint.swift
```diff
@@ -17,11 +27,13 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
         
         It is recommended that the value of this parameter be greater than the `horizontalAccuracy` property of a `CLLocation` object obtained from a `CLLocationManager` object. There is a high likelihood that the user may be located some distance away from a navigable road, for instance if the user is currently on a driveway or inside a building.
      - parameter name: The name of the waypoint. This parameter does not affect the route but may help you to distinguish one waypoint from another.
+     - parameter allowsArrivingOnOppositeSide: A boolean value indicating whether arriving on opposite side is allowed.
      */
-    @objc public init(coordinate: CLLocationCoordinate2D, coordinateAccuracy: CLLocationAccuracy = -1, name: String? = nil) {
+    @objc public init(coordinate: CLLocationCoordinate2D, coordinateAccuracy: CLLocationAccuracy = -1, name: String? = nil, allowsArrivingOnOppositeSide: Bool = true) {
```
1ec5: Is it necessary to compound the proliferation of initializer arguments? This (for now) relatively obscure feature has a meaningful default, unlike `coordinateAccuracy` or `name`. Waypoint was originally a struct, which is why we put lots of arguments in the initializer, but that isn’t to say that every property needs to have an argument.

MapboxDirections/MBWaypoint.swift
```diff
@@ -145,6 +161,13 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
      */
     @objc open var name: String?
     
+    /**
+     A boolean value indicating whether arriving on opposite side is allowed.
+     
+     This parameter can only be used in combination with `RouteOptions.includesSteps` = true.
```
1ec5: > This property has no effect if `RouteOptions.includeSteps` is set to `false`.

MapboxDirectionsTests/RouteOptionsTests.swift
```diff
@@ -80,6 +80,38 @@ class RouteOptionsTests: XCTestCase {
...
+            Waypoint(coordinate: coordinate, coordinateAccuracy: 0),
+            wp1,
+            Waypoint(coordinate: coordinate, coordinateAccuracy: 0)
+        ]
+        
+        let routeOptions = RouteOptions(waypoints: waypoints)
+        routeOptions.includesSteps = true
+        let params = routeOptions.params
+        let approaches = params.filter { $0.name == "approaches" }.first!
+        XCTAssert(approaches.value! == "unrestricted;curb;unrestricted", "waypoints[1] should be restricted to curb")
```
1ec5: Nit: use `XCTAssertEqual(_:_:)` for a more precise assertion message.

PR #303, opened 2018-10-23
Offline Tile downloader

frederoni:
Introduced Offline Tile downloader

Added two new public methods:

- `Directions.fetchAvailableVersions(completionHandler:)`
- `Directions.downloadTiles(in:version:completionHandler:)`

<br>

- [x] Documentation
- [x] Bundle a small region for unit testing
- [x] Unit test downloading
- [x] Deploy api-routing-tiles to production
- [ ] ~~Unit test download progress~~ (#306)
- [ ] ~~Unit test resume download~~ (#306)

cc @1ec5 @JThramer @akitchen


MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,174 @@
...
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(versionString)
+    }
+}
+
+@objc(MBOfflineDirections)
+public class OfflineDirections: NSObject, URLSessionDownloadDelegate {
```
frederoni: We could probably fold this into `Directions` if we mention `offline` in the signature and/or documentation.
@1ec5 what’s your take on this?

MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,139 @@
...
+        
+        let configuration = URLSessionConfiguration.default
+        let url = tilesURL(for: boundingBox, version: version)
+        let session = URLSession(configuration: configuration, delegate: self, delegateQueue: nil)
+        
+        return session.downloadTask(with: url)
+    }
+    
+    public func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didResumeAtOffset fileOffset: Int64, expectedTotalBytes: Int64) {
+        // TODO: resume download when range headers are supported
```
akitchen: Can we add support for this in advance of the feature availability on the server?

MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,139 @@
...
+}
+
+@objc(MBOfflineDirectionsProtocol)
+public protocol OfflineDirectionsProtocol {
+    
+    /**
+     Fetches the available versions.
+     */
+    @discardableResult
+    func availableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask
```
1ec5: Since this method returns a task rather than the offline versions themselves, rename it to `getAvailableOfflineVersions(completionHandler:)`.

MapboxDirections/MBBoundingBox.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import CoreLocation
+
+/**
+ A bounding box represents a geographic region.
+ */
+@objc(MBBoundingBox)
+public class BoundingBox: NSObject, Codable {
```
1ec5: The map SDK has a similar struct named [`MGLCoordinateBounds`](https://github.com/mapbox/mapbox-gl-native/blob/f50d8dbe3afa2556b1ef404e39ef897e0093cd0c/platform/darwin/src/MGLGeometry.h#L69-L75), and MapboxGeocoder.swift has a CLRegion subclass named [RectangularRegion](https://github.com/mapbox/MapboxGeocoder.swift/blob/0af564695d0153bbfddeae729bac8ae22e89334d/MapboxGeocoder/MBRectangularRegion.swift). It would be great if we could make this class a struct, for value semantics; otherwise, a subclass of CLRegion would at least make some of the capabilities more discoverable.

MapboxDirections/MBDirections.swift
```diff
@@ -130,6 +96,12 @@ open class Directions: NSObject {
     /// The Mapbox access token to associate the request with.
     internal let accessToken: String
     
+    /// The handler that reports the download progress of an offline tile pack
+    internal var offlineProgressHandler: OfflineDownloaderProgressHandler?
+    
+    /// The handler that reports the completion of an offline tile pack
+    internal var offlineCompletionHandler: OfflineDownloaderCompletionHandler?
```
1ec5: It’s a bit awkward that these properties are here in the main class definition because of an extension’s conformance to OfflineDirectionsProtocol.

More significantly, these properties make Directions stateful. Directions originally was stateful, going so far as to store the URLSessionTask for the API request, similar to MapKit’s MKDirections. But that made it impossible for a single Directions object to process multiple requests simultaneously, and it also meant that the object had to be held strongly for the duration of the request. #47 moved Directions to [a stateless singleton model](https://github.com/mapbox/MapboxDirections.swift/blob/master/CHANGELOG.md#v060) that has provided a good deal of flexibility to this library. If we’re going to return to the old model, Directions would need to encapsulate the URLSessionTask objects and provide a method for canceling the request.

MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,139 @@
...
+public typealias OfflineDownloaderCompletionHandler = (_ location: URL?,_ response: URLResponse?, _ error: Error?) -> Void
+public typealias OfflineDownloaderProgressHandler = (_ bytesWritten: Int64, _ totalBytesWritten: Int64, _ totalBytesExpectedToWrite: Int64) -> Void
+public typealias OfflineVersionsHandler = (_ version: [Version]?, _ error: Error?) -> Void
+
+struct AvailableVersionsResponse: Codable {
+    let availableVersions: [Version]
+}
+
+@objc(MBVersion)
+public class Version: NSObject, Codable {
```
1ec5: There would be somewhat less boilerplate if we made Version a struct instead. It would have to be defined in Objective-C in order to support both languages.

MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,139 @@
...
+    @discardableResult
+    func downloadTiles(for boundingBox: BoundingBox, version: Version, progressHandler: @escaping OfflineDownloaderProgressHandler, completionHandler: @escaping OfflineDownloaderCompletionHandler) -> URLSessionDownloadTask
+}
+
+extension Directions: OfflineDirectionsProtocol, URLSessionDownloadDelegate {
+    
+    func availableVersionsURL() -> URL {
+        
+        let url = apiEndpoint.appendingPathComponent("route-tiles/v1").appendingPathComponent("versions")
+        var components = URLComponents(string: url.absoluteString)
```
1ec5: Nit: use `URLComponents(url:resolvingAgainstBaseURL:)` for better type safety. (×2)

MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,139 @@
...
+}
+
+@objc(MBVersion)
+public class Version: NSObject, Codable {
+    let versionString: String
+    let versionDate: Date
+    
+    static var dateFormatter: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-dd"
```
1ec5: On the off-chance that there needs to be two versions on the same day, would the API ever assign a more precise timestamp?

MapboxDirections/MBBoundingBox.swift
frederoni: Hm, yeah, it's unfortunate that MapboxDirections doesn't depend on those. Would be nice to share a core set of data types, similar to what gl-native is doing.

I went back and forth with CLRegion but didn't find it adding any values CoordinateBounds makes sense for consistency. How about `{MB}/CoordinateBounds`?

MapboxDirections/OfflineDirections.swift
frederoni: Good point. The version should be represented as a String according to the api team.

MapboxDirections/MBBoundingBox.swift
frederoni: Turf has a [BoundingBox](https://github.com/mapbox/turf-swift/blob/master/Sources/Turf/BoundingBox.swift) but we are partially striving for turf-js terminology there.

1ec5: GeoJSON terminology makes more sense for a library that’s very much about manipulating GeoJSON, such as Turf, but unless we have this library depend on Turf – not necessarily a bad idea – I kind of like CoordinateBounds for consistency with the map SDK. I agree that CLRegion isn’t adding much in this case; MapboxGeocoder.swift has that to be more similar to MapKit’s and Core Location’s geocoders.

> Swift struct wouldn't bridge to Obj-C.

A C struct does bridge to Swift, so implement it in an Objective-C header.

MapboxDirections/OfflineDirections.swift
1ec5: Fetch is fine.

MapboxDirections/MBDirections.swift
1ec5: MapboxDirections.swift takes the approach of exposing the request URL publicly, so you can hook the library into any networking library, not just URLSession but also AFNetworking etc.

MapboxDirections/OfflineDirections.swift
```diff
@@ -0,0 +1,91 @@
...
+        }
+        
+        return task
+    }
+    
+    @discardableResult
+    @objc
+    public func downloadTiles(for boundingBox: BoundingBox,
+                              version: OfflineVersion,
+                              session: URLSession? = URLSession.shared,
```
1ec5: Even more flexible would be to expose `url(forTilesIn:version:)` publicly, so that developers can use the URL with either a custom URLSession or another networking library such as AFNetworking.

PR #311, opened 2018-11-13
Coordinate Truncation

### LINKED ISSUE #309: Truncate precision in `CLLocationCoordinate2D` to relevant decimal places
When crafting a custom `RouteOptions` request , the user has the ability to use the full `Double` bitwidth to describe `CLLocationCoordinate2D` waypoints. This causes us to potentially emit really long requests when we end up requesting a route.

We only pay attention to the the fifth or sixth decimal place in the API (depending on the API version I think? cc @1ec5), so we should truncate our values to match at request time.

/cc @mapbox/navigation-ios

JThramer:
Fixes #309. Adds number formatter that truncates coordinates to relevant number of decimal places, depending on `shapeFormat` setting.

### DIFF
--- a/MapboxDirections/MBWaypoint.swift
+++ b/MapboxDirections/MBWaypoint.swift
@@ -106,6 +106,24 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
      */
     @objc open var coordinateAccuracy: CLLocationAccuracy = -1
     
+    internal enum CoordinatePrecision: Int {
+        case five = 5
+        case six = 6
+     }
+    
+    internal func jsonRepresentation(precision: CoordinatePrecision) -> String? {
+        let formatter = NumberFormatter()
+        formatter.maximumFractionDigits = precision.rawValue
+        
+        guard let lat = formatter.string(for: coordinate.latitude),
+            let long = formatter.string(for: coordinate.longitude) else {
+                return nil
+        }
+        
+        return "\(long),\(lat)"
+    }
+    
+    
     // MARK: Getting the Direction of Approach
     
     /**
--- a/MapboxDirections/MBWaypoint.swift
+++ b/MapboxDirections/MBWaypoint.swift
@@ -111,17 +111,6 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
         case six = 6
      }
     
-    internal func jsonRepresentation(precision: CoordinatePrecision) -> String? {
-        let formatter = NumberFormatter()
-        formatter.maximumFractionDigits = precision.rawValue
-        
-        guard let lat = formatter.string(for: coordinate.latitude),
-            let long = formatter.string(for: coordinate.longitude) else {
-                return nil
-        }
-        
-        return "\(long),\(lat)"
-    }
     
     
     // MARK: Getting the Direction of Approach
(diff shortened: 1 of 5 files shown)


JThramer:
cc @mapbox/navigation-ios

MapboxDirections/MBWaypoint.swift
```diff
@@ -106,6 +106,24 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
...
     @objc open var coordinateAccuracy: CLLocationAccuracy = -1
     
+    internal enum CoordinatePrecision: Int {
+        case five = 5
+        case six = 6
+     }
+    
+    internal func jsonRepresentation(precision: CoordinatePrecision) -> String? {
+        let formatter = NumberFormatter()
+        formatter.maximumFractionDigits = precision.rawValue
```
1ec5: We’d get better performance by keeping just one number formatter around instead of recreating it twice for each waypoint. Given the large number of waypoints required to trigger the error being fixed, the difference might be measurable.

Additionally, the default `NumberFormatter.locale` depends on the application locale. So if the system language is set to Spanish, all requests will produce errors, because Spanish uses a comma as the decimal separator.

Use the standard library math functions instead:

```swift
extension Double {
    func rounded(to precision: Double) -> Double {
        return Double((self * precision).rounded()) / precision
    }
}

"\(coordinate.longitude.rounded(to: 1e6)),\(coordinate.latitude.rounded(to: 1e6))"
```

MapboxDirections/MBWaypoint.swift
```diff
@@ -106,6 +106,24 @@ open class Waypoint: NSObject, NSCopying, NSSecureCoding {
      */
     @objc open var coordinateAccuracy: CLLocationAccuracy = -1
     
+    internal enum CoordinatePrecision: Int {
+        case five = 5
+        case six = 6
+     }
+    
+    internal func jsonRepresentation(precision: CoordinatePrecision) -> String? {
```
1ec5: The return value goes directly into a URL query parameter, not JSON. What we’re implementing here is really `String(describing:precision:)` for a `CLLocationCoordinate2D`.

MapboxDirections/MBWaypoint.swift
akitchen: I agree, NumberFormatters are known to be expensive to create (or perhaps they only used to be?) 

this is worth profiling, and possibly statically memoizing an instance or using @1ec5’s suggestion

1ec5: Formatters are generally expensive to create because the actual formatting is done by CLDR, which is generic enough to support lots of languages and configurations. As far as I know, NumberFormatter doesn’t do a whole lot of caching between instances, because each instance’s configuration is mutable and each configuration option can drastically affect output.

Anyways, formatters are only for output that’s meant for display to the user. This string is going into a URL query parameter that is interpreted as two standard POSIX doubles. So to fix the localization issue I pointed out, we’d need to set `locale` to `en_US_POSIX`. (It isn’t enough to set the decimal separator, because some languages like Arabic use non-Western numerals.) But at that point, we might as well use the standard library math functions, because we don’t benefit at all from NumberFormatter.

PR #314, opened 2018-11-20
No Response Fix

### LINKED ISSUE #310: Directions.calculate can fail silently.
There are situations where the API can send us back a `200` (or some other status code that `URLSession` does not catch, such as `413`) that contains a response, but no routes.

This effectively forces the user to sanity-check that `routes` (in the completion handler) is not `nil`. This shouldn't have to happen. If we don't see a naturally-generated `URLError` come back from the response, we should synthesize one that represents the erroneous state of "No navigation solution came back from the service."

/cc @mapbox/navigation-ios

JThramer:
Is there anyone there?

Fixes #310, where a situation can arise where an API request returns an error, but that error is not matched by our blacklist and thus the SDK returns the response without an `NSError`, but also without a `[Route]` object. This should never happen.

/cc @mapbox/navigation-ios


MapboxDirections/MBDirections.swift
```diff
@@ -1,7 +1,7 @@
 typealias JSONDictionary = [String: Any]
 
 /// Indicates that an error occurred in MapboxDirections.
-public let MBDirectionsErrorDomain = "MBDirectionsErrorDomain"
+public let MBDirectionsErrorDomain = "com.mapbox.directions.ErrorDomain"
```
JThramer: Before anyone asks, this is a great (and rather rare) example of "Do as I say, not as I do" from Apple:

> You can create your own error domains and error codes for use in your own frameworks, or even in your own applications. It is recommended that the string constant for the domain be of the form com.company.framework_or_app.ErrorDomain.

[Doc Link](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorObjectsDomains/ErrorObjectsDomains.html#//apple_ref/doc/uid/TP40001806-CH202-CJBGAIBJ)

akitchen:
Thanks for turning this around. Seems like something that should really have a test case, however. Any chance you could add one?

MapboxDirections/MBDirections.swift
```diff
@@ -226,7 +226,7 @@ open class Directions: NSObject {
      - returns: The data task for the URL.
      - postcondition: The caller must resume the returned task.
      */
-    fileprivate func dataTask(with url: URL, data: Data? = nil, completionHandler: @escaping (_ json: JSONDictionary) -> Void, errorHandler: @escaping (_ error: NSError) -> Void) -> URLSessionDataTask {
+    fileprivate func dataTask(with url: URL, data: Data? = nil, completionHandler: @escaping (_ response: HTTPURLResponse?, _ json: JSONDictionary) -> Void, errorHandler: @escaping (_ error: NSError) -> Void) -> URLSessionDataTask {
```
1ec5: `response` appears to be unused, but I guess it could be useful in the future.

1ec5:
```
Test Case '-[MapboxDirectionsTests.DirectionsTests testKnownBadResponse]' started.
/Users/distiller/project/MapboxDirectionsTests/DirectionsTests.swift:61: error: -[MapboxDirectionsTests.DirectionsTests testKnownBadResponse] : XCTAssertTrue failed - Wrong type of error recieved
Test Case '-[MapboxDirectionsTests.DirectionsTests testKnownBadResponse]' failed (0.009 seconds).
```

PR #322, opened 2018-12-06
Ignore empty image URLs

1ec5:
Fixed an issue where `VisualInstructionComponent(json:)` would set `VisualInstructionComponent.imageURL` to an invalid URL when the JSON representation includes an empty image URL. `URL(string:)` already returns `nil` for invalid URLs, but appending the resolution and file extension onto an empty string made the string into a valid URL – a relative file URL that would be invalid for an HTTP connection.

/cc @mapbox/navigation-ios @kevinkreiser

### DIFF
--- a/VisualInstructionComponentTests.swift
+++ b/VisualInstructionComponentTests.swift
@@ -1,0 +1,13 @@
+import XCTest
+import MapboxDirections
+
+class VisualInstructionComponentTests: XCTestCase {
+    func testJSONInitialization() {
+        let component = VisualInstructionComponent(json: [
+            "text": "Take a hike",
+            "imageBaseURL": "",
+        ])
+        XCTAssertEqual(component.text, "Take a hike")
+        XCTAssertNil(component.imageURL)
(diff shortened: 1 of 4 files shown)


VisualInstructionComponentTests.swift
```diff
@@ -0,0 +1,13 @@
...
+import MapboxDirections
+
+class VisualInstructionComponentTests: XCTestCase {
+    func testJSONInitialization() {
+        let component = VisualInstructionComponent(json: [
+            "text": "Take a hike",
+            "imageURL": "",
+        ])
+        XCTAssertEqual(component.text, "Take a hike")
+        XCTAssertNil(component.imageURL)
```
akitchen: I'm a little confused by this. Your test passes even with the code change reverted.

1ec5: The test was wrong. Fixed in f4a6425c8abe7357eaf89692c32cbba78ee9b112.

PR #326, opened 2018-12-12
Waypoint targets

1ec5:
Added the `Waypoint.targetCoordinate` property corresponding to the Directions API’s `waypoint_targets` property, which specifies a more specific destination location for arrival instructions.

Fixed an issue where the `Waypoint.allowsArrivingOnOppositeSide` property was not copied when copying a `Waypoint` object. Added unit tests of copying and coding waypoints, as well as specifying `targetCoordinate` and other waypoint properties in conjunction with a RouteOptions object.

Added documentation associating some properties with their API counterparts for discoverability.

/cc @mapbox/navigation-ios @danesfeder @danpaz

### DIFF
--- a/MapboxDirections/MBDirectionsOptions.swift
+++ b/MapboxDirections/MBDirectionsOptions.swift
@@ -328,7 +328,7 @@ open class DirectionsOptions: NSObject, NSSecureCoding, NSCopying {
      An array of directions query strings to include in the request URL.
      */
     internal var queries: [String] {
-        return waypoints.compactMap { return "\($0.coordinate.longitude.rounded(to: 1e6)),\($0.coordinate.latitude.rounded(to: 1e6))" }
+        return waypoints.compactMap{ $0.coordinate.stringForRequestURL }
     }
     
     internal var path: String {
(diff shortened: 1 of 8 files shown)


MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -328,7 +328,7 @@ open class DirectionsOptions: NSObject, NSSecureCoding, NSCopying {
      An array of directions query strings to include in the request URL.
      */
     internal var queries: [String] {
-        return waypoints.compactMap { return "\($0.coordinate.longitude.rounded(to: 1e6)),\($0.coordinate.latitude.rounded(to: 1e6))" }
+        return waypoints.compactMap{ $0.coordinate.stringForRequestURL }
```
1ec5: Note the following edge case: if the developer specifies three waypoints and one of them is invalid, this library previously sent a request with all three waypoints, all but guaranteeing an error in the response. Now it sends only two of the waypoints, which may return a valid response (ignoring the invalid coordinate). However, if the invalid waypoint had the `name` or `targetCoordinate` property set, then there would be a mismatch between the path and query parameters in the request URL, resulting in a different error than before in the response.

1ec5:
Tests were failing because of a broken file reference; fixed in 6a2a7077d64f632e72755b1f46d8e88539f74c9d.

PR #340, opened 2019-02-07
Per-waypoint via point behavior

### LINKED ISSUE #334: Add waypoint indices option for Directions API
The Directions API will soon expose a `waypoints` parameter similar to the Map Matching API’s: Project-OSRM/osrm-backend#5345. To support it, we could move `MatchOptions.waypointIndices` and its support code to DirectionsOptions, allowing RouteOptions to inherit it:

https://github.com/mapbox/MapboxDirections.swift/blob/c350b60489c33845551cf0163a5481d32c940479/MapboxDirections/Match/MBMatchOptions.swift#L49-L56

However, this would be a good opportunity to replace the parallel index set with a more object-oriented property on Waypoint that determines whether the individual waypoint is explicitly persisted in the route response. I’m not sure what to name it: `isPersistent`? `isIncludedInCalculatedRoute`? `isRepresentedByArrivalAndDepartureSteps`? `separatesTwoLegs`? `isIncludedInInstructions`? `isImpliedByCalculatedRoute`? `isViaPoint`?

Based on this property, we’ll have to revisit the assumption that `RouteOptions.waypoints` lines up with the response’s waypoints:

https://github.com/mapbox/MapboxDirections.swift/blob/c350b60489c33845551cf0163a5481d32c940479/MapboxDirections/MBRoute.swift#L27-L29

1ec5:
Added the `Waypoint.separatesLegs` property, which you can set to `false` to create a route that travels “via” the waypoint but doesn’t stop there. Deprecated the `MatchOptions.waypointIndices` property in favor of `Waypoint.separatesLegs`, which also works with `RouteOptions`.

Fixed unset properties in  `Waypoint` objects that are included in a calculated `Route`s or `Match`es by copying objects.

[](https://commons.wikimedia.org/wiki/File:Loch_nan_uamh_viaduct.jpg)

Fixes #334.

/cc @mapbox/navigation-ios

### DIFF
--- a/MapboxDirections/Match/MBMatchOptions.swift
+++ b/MapboxDirections/Match/MBMatchOptions.swift
@@ -49,21 +49,23 @@ open class MatchOptions: DirectionsOptions {
     /**
      An index set containing indices of two or more items in `coordinates`. These will be represented by `Waypoint`s in the resulting `Match` objects.
 
-     Use this property when the `includesSteps` property is `true` or when `coordinates` represents a trace with a high sample rate. If this property is `nil`, the resulting `Match` objects contain a waypoint for each coordinate in the match options.
+     Use this property when the `DirectionsOptions.includesSteps` property is `true` or when `coordinates` represents a trace with a high sample rate. If this property is `nil`, the resulting `Match` objects contain a waypoint for each coordinate in the match options.
 
      If specified, each index must correspond to a valid index in `coordinates`, and the index set must contain 0 and the last index (one less than `endIndex`) of `coordinates`.
      */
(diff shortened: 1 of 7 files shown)


MapboxDirections/Match/MBMatchOptions.swift
```diff
@@ -164,3 +173,8 @@ open class MatchOptions: DirectionsOptions {
         return (waypoints, routes)
     }
 }
+
+private protocol MatchOptionsDeprecations {
```
1ec5: This is [a neat trick](https://stackoverflow.com/a/45743766/4585461) for suppressing intentional implementation-related usage of deprecated class members.

PR #353, opened 2019-02-21
Adjust naming of offline tile downloading methods

### LINKED ISSUE #352: Offline download task methods masquerade as action methods
`Directions.fetchAvailableOfflineVersions(completionHandler:)` and `Directions.downloadTiles(in:version:session:completionHandler:)` are named as if they actually start off an HTTP request, but all they do is create and return a URLSessionDownloadTask. That, combined with the fact that a completion handler will typically be a bit lengthy, means it’s extremely easy for the caller to forget to `resume()` the task. Nothing happens, and it takes many breakpoints and a modicum of hairpulling to figure out why. These methods should either be renamed to  `availableVersionsDownloadTask(completionHandler:)` and `downloadTask(forDownloadingTilesIn:version:session:completionHandler:)`, respectively, or they should resume the tasks before returning them, as `calculate(_:completionHandler:)` does.

/cc @mapbox/navigation-ios

akitchen:
Fixes #352

### DIFF
--- a/Sources/MapboxDirections/OfflineDirections.swift
+++ b/Sources/MapboxDirections/OfflineDirections.swift
@@ -19,7 +19,7 @@ public protocol OfflineDirectionsProtocol {
      - parameter completionHandler: A closure of type `OfflineVersionsHandler` which will be called when the request completes
      */
     @objc(fetchAvailableOfflineVersionsWithCompletionHandler:)
-    func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> ()
+    func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask
     
     /**
      Downloads offline routing tiles of the given version within the given coordinate bounds using the shared URLSession. The tiles are written to disk at the location passed into the `completionHandler`.
@@ -28,8 +28,8 @@ public protocol OfflineDirectionsProtocol {
      - parameter version: The version to download. Version is represented as a String (yyyy-MM-dd-x)
      - parameter completionHandler: A closure of type `OfflineDownloaderCompletionHandler` which will be called when the request completes
      */
-    @objc(downloadTilesInCoordinateBounds:withVersion:completionHandler:)
-    func downloadTiles(in coordinateBounds: CoordinateBounds, version: OfflineVersion, completionHandler: @escaping OfflineDownloaderCompletionHandler) -> ()
+    @objc(downloadTilesInCoordinateBounds:version:completionHandler:)
+    func downloadTiles(in coordinateBounds: CoordinateBounds, version: OfflineVersion, completionHandler: @escaping OfflineDownloaderCompletionHandler) -> URLSessionDownloadTask
 }
 
 extension Directions: OfflineDirectionsProtocol {
@@ -60,10 +60,11 @@ extension Directions: OfflineDirectionsProtocol {
      
      - parameter completionHandler: A closure of type `OfflineVersionsHandler` which will be called when the request completes
      */
-    public func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> () {
+    @discardableResult
+    public func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask {
         let task = availableVersionsDataTask(completionHandler: completionHandler)
-
-        return task.resume()
+        task.resume()
+        return task
     }
 
     /**
@@ -71,6 +72,7 @@ extension Directions: OfflineDirectionsProtocol {
      
      - parameter completionHandler: A closure of type `OfflineVersionsHandler` which will be called when the request completes
      */
+    @discardableResult
     @objc(availableVersionsDataTaskWithCompletionHandler:)
     public func availableVersionsDataTask(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask {
         let task = URLSession.shared.dataTask(with: availableVersionsURL) { (data, response, error) in
@@ -101,11 +103,13 @@ extension Directions: OfflineDirectionsProtocol {
      - parameter version: The version to download. Version is represented as a String (yyyy-MM-dd-x)
      - parameter completionHandler: A closure of type `OfflineDownloaderCompletionHandler` which will be called when the request completes
      */
+    @discardableResult
     public func downloadTiles(in coordinateBounds: CoordinateBounds,
                               version: OfflineVersion,
-                              completionHandler: @escaping OfflineDownloaderCompletionHandler) -> () {
+                              completionHandler: @escaping OfflineDownloaderCompletionHandler) -> URLSessionDownloadTask {
         let task = routingTilesDataTaskWith(coordinateBounds, version: version, completionHandler: completionHandler)
-        return task.resume()
+        task.resume()
+        return task
     }
 
     /**
--- a/Sources/MapboxDirections/OfflineDirections.swift
+++ b/Sources/MapboxDirections/OfflineDirections.swift
@@ -62,19 +62,6 @@ extension Directions: OfflineDirectionsProtocol {
      */
     @discardableResult
     public func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask {
-        let task = availableVersionsDataTask(completionHandler: completionHandler)
-        task.resume()
-        return task
-    }
-
-    /**
-     Creates and returns a URLSessionDataTask for downloading available offline routing tile versions. The caller is responsible for calling `-resume` on the task.
-     
-     - parameter completionHandler: A closure of type `OfflineVersionsHandler` which will be called when the request completes
-     */
-    @discardableResult
-    @objc(availableVersionsDataTaskWithCompletionHandler:)
-    public func availableVersionsDataTask(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask {
         let task = URLSession.shared.dataTask(with: availableVersionsURL) { (data, response, error) in
             if let error = error {
                 return completionHandler(nil, error)
@@ -93,6 +80,8 @@ extension Directions: OfflineDirectionsProtocol {
             }
         }
         
+        task.resume()
+        
         return task
     }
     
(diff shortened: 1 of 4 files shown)


MapboxDirectionsTests/OfflineDirectionsTests.swift
```diff
@@ -24,14 +24,13 @@ class OfflineDirectionsTests: XCTestCase {
             return OHHTTPStubsResponse(jsonObject: jsonObject, statusCode: 200, headers: ["Content-Type": "application/json"])
         }
         
-        directions.fetchAvailableOfflineVersions { (versions, error) in
+        directions.fetchAvailableOfflineVersions(completionHandler: { (versions, error) in
```
frederoni: does trailing closure syntax still work here?

frederoni:
The flaky GeoJSON tests may be caused by unsorted JSON.

MapboxDirectionsTests/OfflineDirectionsTests.swift
akitchen: yes, it does. I think there was a battle with autocomplete during the refactor. updated.

akitchen:
> The flaky GeoJSON tests may be caused by unsorted JSON.

I ran it again and it passed. Can't seem to reproduce locally. 

where is the non-determinisim? there seems to be a timeout in the test, which could indicate the OHHTTPStubs isn't yet bootstrapped properly, or there's a race. I think we've seen this before with NSURLSession-based network stubbing in tests.

I don;t think this should block merge of this PR (assuming we're 👍 otherwise) but we do need to get to the bottom of it.

frederoni:
> where is the non-determinisim?

I can't find the non-deterministic URL here, but it MapboxStatic.swift, the OHHTTPStub parsed GeoJSON when setting up the router which would lead to undeterministic route/url. 

We worked around that in the test case by adding:
https://github.com/mapbox/MapboxStatic.swift/blob/3ab49ce2f8f51baee2895f5bef5b26721c41b5a4/MapboxStatic/Overlay.swift#L399-L425

MapboxDirections/OfflineDirections.swift
```diff
@@ -14,20 +14,22 @@ struct AvailableVersionsResponse: Codable {
...
     /**
-     Fetches the available versions.
+     Fetches the available routing tile versions. Typically, the most recent version should be passed into `downloadTiles(in:version:completionHandler:)`.
+
+     - parameter completionHandler: Executes when the download completes or fails.
      */
-    @discardableResult
-    func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask
+    @objc(fetchAvailableOfflineVersionsWithCompletion:)
+    func fetchAvailableOfflineVersions(completionHandler: @escaping OfflineVersionsHandler) -> ()
```
1ec5: I think it’s still useful to return a data task, as the “calculate” methods in Directions do. That way the application can cancel the task if necessary.

MapboxDirections/OfflineDirections.swift
```diff
@@ -14,20 +14,22 @@ struct AvailableVersionsResponse: Codable {
...
+     Fetches routing tiles needed to generate a route within the given bounding box.
      
      - parameter coordinateBounds: The region of the pack to be downloaded.
      - parameter version: The version of the pack to be downloaded.
-     - parameter completionHandler: Informs when the download is completed or failed. The offline pack may be moved from the temporary directory and to a persistent store at this point.
+     - parameter completionHandler: Executes when the download completes or fails. The offline pack may be moved from the temporary directory to a persistent store at this point.
      */
-    @discardableResult
-    func downloadTiles(in coordinateBounds: CoordinateBounds, version: OfflineVersion, session: URLSession?, completionHandler: @escaping OfflineDownloaderCompletionHandler) -> URLSessionDownloadTask
+    @objc(downloadTilesInCoordinateBounds:withVersion:completion:)
```
1ec5: Nit: the usual parameter name in Objective-C is either `completionHandler` or `completionBlock`.

MapboxDirections/OfflineDirections.swift
```diff
@@ -84,24 +90,23 @@ extension Directions: OfflineDirectionsProtocol {
...
-                              session: URLSession? = URLSession.shared,
-                              completionHandler: @escaping OfflineDownloaderCompletionHandler) -> URLSessionDownloadTask {
-        
-        let urlSession = session ?? URLSession.shared
+                              completionHandler: @escaping OfflineDownloaderCompletionHandler) -> () {
+        let task = routingTilesDataTaskWith(coordinateBounds, version: version, completionHandler: completionHandler)
+        return task.resume()
+    }
+
+    @objc(routingTilesDataTaskWithCoordinateBounds:version:completion:)
```
1ec5: If we make this public, let’s add documentation so it shows up in jazzy.

MapboxDirections/OfflineDirections.swift
akitchen: really? OK, let's go with `completionHandler` since that's what's in the swift

MapboxDirections/OfflineDirections.swift
akitchen: 👍

akitchen: do you agree that it's worth it to expose this publicly? I wanted to address #352 but also preserve finer-grained control for developers

akitchen:
I'm not so sure the JSON keys sort order is a factor here - the stub is matching on:
```
        let queryParams: [String: String?] = [
            "alternatives": "true",
            "geometries": String(describing: shapeFormat),
            "overview": "full",
            "steps": "true",
            "continue_straight": "true",
            "access_token": BogusToken,
        ]
        stub(condition: isHost("api.mapbox.com")
            && isPath("/directions/v5/mapbox/driving/-122.42,37.78;-77.03,38.91.json")
            && containsQueryParams(queryParams)) { _ in
[...]
```
... so I interpret the expectation timeout to mean that it isn't set up, or that internal async within the URL loading system tearing down the previous stubs caught this one in its dragnet as well.

Since these stubs are so specific and localized, I think we can let them live longer and tear them down at the end of this test _class_'s execution. Small commit coming to test this theory.

akitchen:
Interestingly, the test still failed with https://github.com/mapbox/MapboxDirections.swift/pull/353/commits/eeb41aae1873f9c84041ea2bb511050a5d68cc67 (which passes locally). 🤔

MapboxDirections/OfflineDirections.swift
akitchen: oh, I missed this comment earlier. does that render this PR (and #352) moot @1ec5 ?

akitchen:
I reverted the change to test teardown since it seems to incur the same level of flakiness. We can move the discussion of JSON key sorting into #354 and try to keep this focused and get it merged, or not merged, depending on https://github.com/mapbox/MapboxDirections.swift/pull/353#discussion_r259162288 (cc @1ec5 )

MapboxDirections/OfflineDirections.swift
1ec5: No, I think the important thing is that this method resumes the task. It’s fine that it also returns the task.

MapboxDirections/OfflineDirections.swift
1ec5: Sure, that’s fine.

MapboxDirections/OfflineDirections.swift
```diff
@@ -14,20 +14,22 @@ struct AvailableVersionsResponse: Codable {
...
-     - parameter version: The version of the pack to be downloaded.
-     - parameter completionHandler: Informs when the download is completed or failed. The offline pack may be moved from the temporary directory and to a persistent store at this point.
+     - parameter coordinateBounds: The bounding box
+     - parameter version: The version to download. Version is represented as a String (yyyy-MM-dd-x)
+     - parameter completionHandler: A closure of type `OfflineDownloaderCompletionHandler` which will be called when the request completes
      */
-    @discardableResult
-    func downloadTiles(in coordinateBounds: CoordinateBounds, version: OfflineVersion, session: URLSession?, completionHandler: @escaping OfflineDownloaderCompletionHandler) -> URLSessionDownloadTask
+    @objc(downloadTilesInCoordinateBounds:withVersion:completionHandler:)
+    func downloadTiles(in coordinateBounds: CoordinateBounds, version: OfflineVersion, completionHandler: @escaping OfflineDownloaderCompletionHandler) -> ()
```
1ec5: Either the argument label in Swift should be `with:` or the selector piece in Objective-C should be just `version:`.

andrlee: @1ec5 should we plan a major release for this?

Sources/MapboxDirections/OfflineDirections.swift
```diff
@@ -54,14 +56,25 @@ extension Directions: OfflineDirectionsProtocol {
...
+     Creates and returns a URLSessionDataTask for downloading available offline routing tile versions. The caller is responsible for calling `-resume` on the task.
+     
+     - parameter completionHandler: A closure of type `OfflineVersionsHandler` which will be called when the request completes
+     */
+    @discardableResult
+    @objc(availableVersionsDataTaskWithCompletionHandler:)
+    public func availableVersionsDataTask(completionHandler: @escaping OfflineVersionsHandler) -> URLSessionDataTask {
         let task = URLSession.shared.dataTask(with: availableVersionsURL) { (data, response, error) in
             if let error = error {
                 return completionHandler(nil, error)
```
1ec5: `availableVersionsDataTask(_:)` still confuses me; I don’t really see the point in keeping it separate. In general, for any kind of API request, the developer would only ever need two kinds of methods:

* A method that encapsulates requesting and processing a response, taking a completion handler and returning a (discardable) task for cancellation.
* A method or property that provides access to the request payload itself without side effects.

These requirements are satisfied by `fetchAvailableOfflineVersions(_:)` and `availableVersionsURL`, and by `downloadTiles(in:version:)` and `tilesURL(for:version:)`. `availableVersionsDataTask(_:)` and `routingTilesDataTaskWith(_:version:completionHandler:)` are confusing hybrids, because they look like accessors or factory methods but take completion handlers, as if they’re the methods that perform the task. In fact, `completionHandler` is documented as being “called when the request completes”, but calling the method by itself doesn’t cause the completion handler to ever get called.

We can leave these methods in, but we should probably remove `@discardableResult` because the methods do nothing on their own.

frederoni: I agree. The hybrid methods don't improve readability or versatility so I folded them into their counterparts.

frederoni:
ready for review
cc @1ec5

Issue #51, opened 2016-06-06
Polyline-encoded geometry omitted from v5 steps


1ec5:
Polyline-encoded geometries in v5 responses are correctly decoded at the route level, but at the step level, we’re assuming GeoJSON encoding and omitting the geometry if it’s polyline-encoded.

/cc @frederoni

1ec5:
Meanwhile, if you request GeoJSON geometry, you get no route-level geometry. 😆

PR #362, opened 2019-03-23
Add support for Swift Package Manager

### LINKED ISSUE #234: Support Swift Package Manager
Add a Package.swift file so that [Swift Package Manager](https://swift.org/package-manager/) can incorporate this library into a project.

Blocked by raphaelmor/Polyline#35.

/cc @frederoni @captainbarbosa

frederoni:
~Fixes 318 - Support for static library using CocoaPods~
Fixes #234 - Swift Package Manager on macOS

- [x] Audit Obj-C bridgeability
- [ ] ~Deprecate MapboxDirections.swift on CocoaPods public repo in favor of MapboxDirections~ (Reverted pod rename to minimize the changes)

cc @1ec5


1ec5:
> I wonder if we can replace all extensible and bitmask Obj-C enums. 🤔

Unfortunately, that won’t be feasible as long as this library supports Objective-C. Would it be possible to maintain parallel types, with CocoaPods and Carthage using the Objective-C version and SPM using the Swift version?

frederoni:
> Would it be possible to maintain parallel types

Great idea. That should be relatively straightforward, and only four duplicated data types in the shim, which shouldn't be too inconvenient to maintain.

Package.swift
```diff
@@ -0,0 +1,28 @@
...
+    name: "MapboxDirections",
+    products: [
+        // Products define the executables and libraries produced by a package, and make them visible to other packages.
+        .library(
+            name: "MapboxDirections",
+            targets: ["MapboxDirections"]),
+        ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        .package(url: "../Polyline", .branch("fred/spm4.2"))
```
1ec5: Remember to update this line before merging.

Sources/MapboxDirections/MBLane.swift
```diff
@@ -3,19 +3,30 @@ import Foundation
...
+    #if SWIFT_PACKAGE
+    public let indications: LaneIndication
+    #else
     @objc public let indications: LaneIndication
+    #endif
     
     /**
      Initializes a new `Lane` using the given lane indications.
      */
+    #if SWIFT_PACKAGE
```
frederoni: I'm not entirely satisfied with this workaround, having to provide duplicated initializers.
What I would like to see is a pre-processor macro:
```
#if SWIFT_PACKAGE
#define OBJC_NO_SPM
#define OBJC_NO_SPM(x)
#else
#define OBJC_NO_SPM @objc
#define OBJC_NO_SPM(x) @objc(x)
#endif
```
but an Obj-C macro would not work with SPM and SPM doesn't support these macros.

`@objcMembers` solves roughly half of the issues, since it overrides the `@objc`-flag if it's not able to bridge the to Obj-C, but it only works for some methods and properties.

frederoni:
OHHTTPStubs doesn't provide support for SPM so we're not running tests on the SPM flavor yet.

Package.swift
```diff
@@ -0,0 +1,28 @@
...
+    name: "MapboxDirections",
+    products: [
+        // Products define the executables and libraries produced by a package, and make them visible to other packages.
+        .library(
+            name: "MapboxDirections",
+            targets: ["MapboxDirections"]),
+        ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        .package(url: "https://github.com/raphaelmor/Polyline.git", .branch("master"))
```
frederoni: @1ec5 Would you mind making a new release of Polyline?

Changes since [v4.2.0](https://github.com/raphaelmor/Polyline/releases/tag/v4.2.0):
 Added support for Swift Package Manager ([#51](https://github.com/raphaelmor/Polyline/pull/51))

Sources/MapboxDirections/MBLaneIndicationComponent.swift
```diff
@@ -1,22 +1,24 @@
+import Foundation
+
+
 /**
 A component that represents a lane  representation of an instruction.
  */
-@objc(MBLaneIndicationComponent)
```
frederoni: Revert the Obj-C prefix and add an SPM version of this type

Package.swift
1ec5: [All set](https://github.com/raphaelmor/Polyline/releases/tag/v4.2.1).

Sources/MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -219,6 +221,7 @@ public enum InstructionFormat: UInt, CustomStringConvertible {
 
  You do not create instances of this class directly. Instead, create instances of `MatchOptions` or `RouteOptions`.
  */
+@objcMembers
```
1ec5: Huh, didn’t know about this annotation. Good find. There are still some bare `@objc` keywords in the class; are they redundant now?

Sources/MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -382,7 +390,12 @@ open class DirectionsOptions: NSObject, NSSecureCoding, NSCopying {
 
      This property should be set to `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`. The default value of this property is `MBDirectionsProfileIdentifierAutomobile`, which specifies driving directions.
      */
+    #if SWIFT_PACKAGE
+    open var profileIdentifier: MBDirectionsProfileIdentifier
+    #else
```
1ec5: Is this conditional compilation still necessary now that the overall class is annotated with `@objcMembers`?

Sources/MapboxDirections/MBDirectionsOptions.swift
frederoni: `@objc` is still needed for refinements. It's also useful for compile-time errors of unbridgeable types.
Only using `@objcMembers` would fail silently if a type all of a sudden doesn't bridge to Objective-C. Haven't figured out a best practice yet, but as long as we have Obj-C bridging tests, a missing `@objc` annotation is fine since it would get caught by the test at compile-time.

Sources/MapboxDirections/MBDirectionsOptions.swift
frederoni: good catch, not needed anymore.

Sources/MapboxDirections/MBRouteOptions.swift
```diff
@@ -1,3 +1,15 @@
+import Foundation
+import CoreLocation
+
+#if SWIFT_PACKAGE
+public enum MBDirectionsProfileIdentifier: String {
+    case automobile = "mapbox/driving"
+    case automobileAvoidingTraffic = "mapbox/driving-traffic"
+    case cycling = "cycling"
```
1ec5:
```suggestion
    case cycling = "mapbox/cycling"
```

Sources/MapboxDirections/MBRouteOptions.swift
```diff
@@ -1,3 +1,15 @@
...
+    case walking = "walking"
```
1ec5:
```suggestion
    case walking = "mapbox/walking"
```

frederoni:
We can put Obj-C++ inside a separate compat dependency in SPM and compile the Objective-C code along with the Swift framework and therefore support Objective-C w/o  having to maintain duplicated types for the SPM version.

Sources/MapboxDirections/MBDirectionsOptions.swift
```diff
@@ -418,7 +425,7 @@ open class DirectionsOptions: NSObject, NSSecureCoding, NSCopying {
 
      By default, no attribute options are specified. It is recommended that `routeShapeResolution` be set to `.full`.
      */
-    @objc open var attributeOptions: AttributeOptions = []
+    open var attributeOptions: MBAttributeOptions = []
```
1ec5: Can the type remain as `AttributeOptions`, or does it have to be downgraded to `MBAttributeOptions`?

PR #363, opened 2019-04-01
Fix support for Xcode 9 -> 10.2

frederoni:
Upgraded to CircleCI 2.1

Running tests on
Xcode 9.4.1 iOS 9.3
Xcode 10.0 iOS 12.0
Xcode 10.2 iOS 12.2

Despite the branch name, this PR doesn't migrate to Swift 5, it merely validates compatibility with various versions of Xcode.

cc @1ec5


MapboxDirectionsTests/V5Tests.swift
```diff
@@ -60,7 +60,7 @@ class V5Tests: XCTestCase {
         }
         XCTAssertNotNil(task)
         
-        waitForExpectations(timeout: 2) { (error) in
+        waitForExpectations(timeout: 5) { (error) in
```
frederoni: Even locally, 2 seconds would fail 1/16 times for the 3.2 MB GeoJSON

Cartfile.private
```diff
@@ -1,2 +1,2 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 4.4
-github "AliSoftware/OHHTTPStubs" ~> 6.0
+github "linksmt/OHHTTPStubs" "563f48d3fab84ef04639649c770b00f4fa502cca"
```
1ec5: Is there a PR for this change or something else we can keep track of for when master gets updated? AliSoftware/OHHTTPStubs#295 is already in [v7.0.0](https://github.com/AliSoftware/OHHTTPStubs/releases/tag/7.0.0); is this for something else?

MapboxDirectionsTests/V5Tests.swift
1ec5: Yikes, somehow I didn’t realize our test fixtures were that big. This isn’t a performance test, so we should probably think about breaking this test case up into minimal test cases the next time we need to fix it.

Cartfile.private
1ec5: Oh, looks like this commit also pulls in AliSoftware/OHHTTPStubs#286. The CircleCI configuration isn’t testing on watchOS anyways. So I think we can pin to the main pod’s v7._x_.

frederoni: Yea, the reason why I switched to the fork here is the missing watchOS scheme in the original. We don't test on watchOS but we do compile for watchOS. We could switch back but it would require a workaround to run carthage bootstrap and ignore the .private and .resolved files.

PR #366, opened 2019-04-09
Add support for Linux

frederoni:
_this PR is targeting the `fred/spm` branch_

Outlined support for Linux.


1ec5:
Out of curiosity, what’s left before this PR could be ready for review?

frederoni:
The majority of the changes here is to avoid method dispatching on Linux. If or when we drop support for Objective-C, the necessary changes to support Linux would be a minor shim for CoreLocation data types, roughly 60 lines of code. Another blocker is https://github.com/raphaelmor/Polyline/pull/55

It will be easier to start this PR from scratch so I'm closing this for now.

PR #367, opened 2019-04-16
Add support for speed limits via maxspeed annotations.

avi-c:
Include SDK/API support for speed limits from the Directions API via the maxspeed annotations in the directions response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes to MapboxDirections.swift
 
+## master
+
+* Added support for maxspeed annotations from the Directions API. ([#367](https://github.com/mapbox/MapboxDirections.swift/pull/367))
+
 ## v0.30.0
 
 * `Directions.fetchAvailableOfflineVersions(completionHandler:)` and `Directions.downloadTiles(in:version:completionHandler:)` now resumes the data task before returning it to conform to its naming conventions and avoid confusion. ([#353](https://github.com/mapbox/MapboxDirections.swift/pull/353))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## master
 
-* Added support for maxspeed annotations from the Directions API. ([#367](https://github.com/mapbox/MapboxDirections.swift/pull/367))
+* Added `AttributeOptions.maximumSpeedLimit` for getting maximum posted speed limits in the `RouteLeg.segmentMaximumSpeedLimits` property. ([#367](https://github.com/mapbox/MapboxDirections.swift/pull/367))
 
 ## v0.30.0
 
--- a/Directions Example/ViewController.swift
+++ b/Directions Example/ViewController.swift
@@ -76,6 +76,8 @@ class ViewController: UIViewController, MBDrawingViewDelegate {
         wp2.allowsArrivingOnOppositeSide = false
         let options = RouteOptions(waypoints: [wp1, wp2])
         options.includesSteps = true
+        options.routeShapeResolution = .full
+        options.attributeOptions = [.congestionLevel, .maximumSpeedLimit]
         
         Directions(accessToken: MapboxAccessToken).calculate(options) { (waypoints, routes, error) in
             guard error == nil else {
--- a/MapboxDirections.xcodeproj/project.pbxproj
+++ b/MapboxDirections.xcodeproj/project.pbxproj
@@ -274,6 +274,7 @@
 		F448F8401DDCC709000BC343 /* Polyline.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F448F83C1DDCC6F3000BC343 /* Polyline.framework */; };
 		F448F8411DDCC70C000BC343 /* OHHTTPStubs.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F448F83B1DDCC6F3000BC343 /* OHHTTPStubs.framework */; };
 		F448F8421DDCC70D000BC343 /* Polyline.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F448F83C1DDCC6F3000BC343 /* Polyline.framework */; };
+		F4B0022A22650A9F00CF44FF /* MapboxDirections.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = DA6C9D881CAE442B00094FBC /* MapboxDirections.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		F4D785EF1DDD82C100FF4665 /* RouteStepTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4D785EE1DDD82C100FF4665 /* RouteStepTests.swift */; };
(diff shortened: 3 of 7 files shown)


1ec5:
This is a revival of #250.

MapboxDirections/MBAttribute.h
```diff
@@ -35,4 +35,9 @@ typedef NS_OPTIONS(NSUInteger, MBAttributeOptions) {
      This attribute requires `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`. Any other profile identifier produces `CongestionLevel.unknown` for each segment along the route.
      */
     MBAttributeCongestionLevel = (1 << 4),
+
+    /**
+     Doc me.
```
1ec5: Do this. 😉

MapboxDirections/MBSpeedLimit.swift
```diff
@@ -0,0 +1,112 @@
+//
+//  MBSpeedLimit.swift
+//  MapboxDirections
+//
+//  Created by Avi Cieplinski on 4/15/19.
+//  Copyright © 2019 Mapbox. All rights reserved.
```
1ec5: Remove this file header block.

MapboxDirections/MBSpeedLimit.swift
```diff
@@ -0,0 +1,112 @@
...
+//
+
+import Foundation
+
+@objc(MBSpeedLimit)
```
1ec5: This class needs a documentation comment; otherwise, the entire class will be missing from the generated jazzy docset.

MapboxDirections/MBSpeedLimit.swift
```diff
@@ -0,0 +1,112 @@
...
+public class SpeedLimit: NSObject {
+
+    @objc public static let invalid = Double(-1)
+
+    /**
+     A unitless measure of speed which is dependent on the `SpeedLimit.speedUnit`.
+
+     By default, the speed will be unknown and equal to `MBSpeedIsInvalid` which is -1.
+
+     If the speed is none, the value will be equal to `Double.greatestFiniteMagnitude`.
```
1ec5:
```suggestion
     If there is no maximum speed limit, the property is set to `Double.greatestFiniteMagnitude`.
```

MapboxDirections/MBSpeedLimit.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     Indicates the road segment uses miles per hour for measuring speed limits.
+     */
+    case milesPerHour
+
+    /**
+     Indicates the road segment uses kilometers per hour for measuring speed limits.
+     */
+    case kilometersPerHour
```
1ec5: Argh, it would be great if we could just use `Measurement` along with `UnitSpeed.kilometersPerHour`, but alas we still support iOS 9.0 (along with macOS 10.11.0, tvOS 9.0, and watchOS 2.0). I’m chomping at the bit to drop iOS 9 support.

/cc @mapbox/navigation-ios

MapboxDirections/MBSpeedLimit.swift
```diff
@@ -0,0 +1,112 @@
...
+        }
+        self.init(rawValue: unit.rawValue)
+    }
+
+    public var description: String {
+        switch self {
+        case .milesPerHour:
+            return "mph"
+        case .kilometersPerHour:
+            return "kph"
```
1ec5: The Directions API seems to use `km/h` rather than `kph` to indicate kilometers per hour. The initializer is already using `km/h`, so this property doesn’t round-trip.

```suggestion
            return "km/h"
```

MapboxDirections/MBSpeedLimit.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Created by Avi Cieplinski on 4/15/19.
+//  Copyright © 2019 Mapbox. All rights reserved.
+//
+
+import Foundation
+
+@objc(MBSpeedLimit)
+public class SpeedLimit: NSObject {
+
+    @objc public static let invalid = Double(-1)
```
1ec5: This member needs a documentation comment.

1ec5: I think we should offer a way for the developer to easily check whether the speed limit as a whole is invalid (unset), not just its value. We could define:

```swift
@objc public static let invalid = SpeedLimit(value: -1, unit: .kilometersPerHour)
```

plus an `isEqual(_:)` implementation, so that the developer could check:

```swift
if maximumSpeed == .invalid { … }
```

/ref https://github.com/mapbox/MapboxDirections.swift/pull/250#discussion_r176279345

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes to MapboxDirections.swift
 
+## v0.27.4
+
+* Added support for maxspeed annotations from the Directions API. ([#367](https://github.com/mapbox/MapboxDirections.swift/pull/367))
```
1ec5:
```suggestion
* Added `AttributeOptions.maximumSpeedLimit` for getting maximum posted speed limits in the `RouteLeg.segmentMaximumSpeedLimits` property. ([#367](https://github.com/mapbox/MapboxDirections.swift/pull/367))
```

MapboxDirections/MBRouteLeg.swift
```diff
@@ -173,6 +182,15 @@ open class RouteLeg: NSObject, NSSecureCoding {
      */
     public let segmentCongestionLevels: [CongestionLevel]?
 
+    /**
+     An array containing the maximum speed along each road segment in the route leg geometry.
```
1ec5: Distinguish between maximum actual or expected speed and maximum posted speed limit:

```suggestion
     An array containing the maximum posted speed limit along each road segment in the route leg geometry.
```

MapboxDirections/MBRouteLeg.swift
```diff
@@ -173,6 +182,15 @@ open class RouteLeg: NSObject, NSSecureCoding {
...
+
+     Speed limit data is available in [a number of countries and territories worldwide](https://docs.mapbox.com/help/how-mapbox-works/directions/).
```
1ec5: It might be worth mentioning that the returned speed limits may be [advisory speed limits](https://en.wikipedia.org/wiki/Advisory_speed_limit) on entrance and exit ramps where legal speed limits aren’t posted.

MapboxDirections/MBSpeedLimit.swift
1ec5: A proposed change to the Directions API would add information about locally relevant units to each step, making this class somewhat redundant. We could store the speed limit annotations as `CLLocationSpeed`s (concerting everything to meters per second). Then client code could losslessly convert back to kilometers per hour or miles per hour as indicated by the step.

1ec5:
I revived #250 and rebased and retargeted this PR atop it, because that PR already had tests and handled Autobahns correctly.

1ec5:
To do before merging:

* [ ] Drop support for deploying backwards to iOS 9._x_, macOS 10.11._x_, tvOS 9._x_, and watchOS 2._x_: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r275962395
* [ ] Replace SpeedUnit with `Measurement<UnitSpeed>`: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r275962395
* [ ] Subclass `Measurement<UnitSpeed>` as `SignpostedSpeed`, factoring in the indicated sign style: https://github.com/mapbox/MapboxDirections.swift/pull/367#discussion_r277903175

PR #368, opened 2019-04-24
Migrate to Swift 4.2

1ec5:
A few minor modifications to migrate to Swift 4.2.

/cc @frederoni


MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -1851,7 +1851,6 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = DEBUG;
 				SWIFT_SWIFT3_OBJC_INFERENCE = Off;
-				SWIFT_VERSION = 4.0;
```
frederoni: It's convenient to set these at the project level and inherit on the target level. 👍

PR #370, opened 2019-05-10
Walking parameters

### LINKED ISSUE #371: Add walking bias options
The `mapbox/walking` profile now supports [three new parameters](https://docs.mapbox.com/api/navigation/#optional-parameters-for-the-mapboxwalking-profile), for now in the Directions API but soon in the Map Matching API as well:

* `walking_speed` – expose as `DirectionsOptions.speed`, of type `CLLocationSpeed` (measured in meters per second)
* `walkway_bias` – `DirectionsOptions.walkwayBias`, of type `Double` or `Float`
* `alley_bias` – `DirectionsOptions.alleyBias`, of type `Double` or `Float`

/cc @mapbox/navigation-ios @kevinkreiser

frederoni:
This PR adds walking parameters (`alleyPriority`, `walkwayPriority`, and `speed`) to `RouteOptions`.

- [x] Audit Obj-C bridging

Fixes #371.

cc @1ec5 @kevinkreiser 👀


MapboxDirections/MBRouteOptions.swift
```diff
@@ -111,6 +118,32 @@ open class RouteOptions: DirectionsOptions {
      Currently, you can only specify a single road class to avoid.
      */
     @objc open var roadClassesToAvoid: RoadClasses = []
+    
+    /**
+     A bias which determines whether the route should prefer or avoid the use of alleys. The
+     allowed range of values is from -1.0 to 1.0, where -1 indicates preference to avoid
+     alleys, 1 indicates preference to favor alleys, and 0 indicates no preference.
```
kevinkreiser: :pray: thank you for grabbing the correct docs!

frederoni:
Initially in https://github.com/mapbox/MapboxDirections.swift/pull/370/commits/68dfa3f2fc5ccdce4d77587dc823b7e5e492b5c9, I started with a clear separation between route options and walking options by subclassing directions options because walking parameters and driving parameters are mutually exclusive but not yet enforced on the server-side. However, currently, as this PR stands, the walking parameters are folded into RouteOptions but the relevant query items are parsed for the corresponding profile identifier.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -119,9 +155,21 @@ open class RouteOptions: DirectionsOptions {
...
+            queryItems.append(URLQueryItem(name: "alley_bias", value: String(alleyBias)))
+            queryItems.append(URLQueryItem(name: "walkway_bias", value: String(walkwayBias)))
+            queryItems.append(URLQueryItem(name: "walking_speed", value: String(walkingSpeed)))
+        case .cycling:
+            break
+        default:
+            break
         }
 
         if !roadClassesToAvoid.isEmpty {
```
frederoni: `"exclude"` (roadClassesToAvoid) should only be appended if profile identifier is any of `[.automobile, .automobileAvoidingTraffic, .cycling]` for now.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -119,9 +155,21 @@ open class RouteOptions: DirectionsOptions {
             URLQueryItem(name: "alternatives", value: String(includesAlternativeRoutes)),
             URLQueryItem(name: "continue_straight", value: String(!allowsUTurnAtWaypoint))
         ])
-
-        if includesExitRoundaboutManeuver {
-            queryItems.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
+        
+        switch profileIdentifier {
+        case .automobile, .automobileAvoidingTraffic:
```
1ec5: These switch statements might become more invasive over time. Maybe we can extend the MBDirectionsProfileIdentifier enumeration so that a given profile knows which parameters it supports?

MapboxDirections/MBRouteOptions.swift
```diff
@@ -111,6 +118,35 @@ open class RouteOptions: DirectionsOptions {
...
+     */
+    @objc open var walkwayBias: Double = 0
+    
+    /**
+     Walking speed in meters per second. Must be between 0.14 and 6.94 meters per second.
+     
+     Defaults to 1.42
+     This property has no effect unless profile identifier is set to `MBDirectionsProfileIdentifier.walking`.
+     */
+    @objc open var walkingSpeed: CLLocationSpeed = 1.42
```
1ec5: Should we leave open the possibility that the cycling profile may support specifying a maximum attainable speed in the future? Naming the option this way means we’d later have to add a `cyclingSpeed` property, and neither would be applicable in the driving profiles. We should avoid a situation where DirectionsOptions or RouteOptions is saddled with too many options that are mutually exclusive. We could name the property `speed` or `expectedSpeed`, still with the note that it’s ignored for profiles other than walking.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -111,6 +118,35 @@ open class RouteOptions: DirectionsOptions {
...
+     -1.0 to 1.0, where -1 indicates indicates preference to avoid walkways, 1 indicates preference
+     to favor walkways, and 0 indicates no preference.
+     
+     Defeaults to 0.
+     This property has no effect unless profile identifier is set to `MBDirectionsProfileIdentifier.walking`.
+     */
+    @objc open var walkwayBias: Double = 0
+    
+    /**
+     Walking speed in meters per second. Must be between 0.14 and 6.94 meters per second.
```
1ec5: Let’s add constants or enumeration cases for these preset values (same for the other properties). Otherwise, there’s a risk that magic numbers could float around in obscure client-side code, only to break when things change on the server side.

MapboxDirections/MBRouteOptions.swift
```diff
@@ -119,9 +155,21 @@ open class RouteOptions: DirectionsOptions {
...
             URLQueryItem(name: "continue_straight", value: String(!allowsUTurnAtWaypoint))
         ])
-
-        if includesExitRoundaboutManeuver {
-            queryItems.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
+        
+        switch profileIdentifier {
+        case .automobile, .automobileAvoidingTraffic:
+            if includesExitRoundaboutManeuver {
+                queryItems.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
```
1ec5: All profiles support this parameter, and roundabouts do occasionally occur along bike/hike trails. Better that this library pass through the parameter so that the navigation SDK can assume its presence.

1ec5:
d57c1e0e7ee6b6cb56d18f18ea5c9bf43c2f188e renamed `alleyBias`, `walkwayBias`, and `walkingSpeed` to `alleyPriority`, `walkwayPriority`, and `speed`, respectively. `alleyPriority` and `walkwayPriority` now make use of a new `MBDirectionsPriority` extensible enumeration, which allows us to define preset values without proliferating global constants or hard-coding magic number literals.

PR #382, opened 2019-09-27
Remove Objective-C compatibility; adopt Swift language features

### LINKED ISSUE #134: Adopt JSONDecoder

### LINKED ISSUE #228: Please refactor your unwrapping optionals
I use MapboxDirections 0.15.1. After response my crashed with an unwrapping optional issue in MBRouteStep line 600.
`let drivingSide = DrivingSide(description: json["driving_side"] as! String) ?? .right`
I resolve the issue by:
`let drivingSide = DrivingSide(description: json["driving_side"] as? String ?? "") ?? .right`
I can imaging that they are many other optional issues in the project.

### LINKED ISSUE #381: Discontinue Objective-C compatibility
We’re discontinuing support for applications written purely in Objective-C to use the iOS navigation SDK. The same change will take place in the iOS navigation SDK: mapbox/mapbox-navigation-ios#2231. Applications written in Objective-C will need to implement compatibility shims in Swift to continue to interoperate with this library.

Since #47, we’ve half-heartedly supported Objective-C by ensuring bridgeability of key APIs, but there are many gaps in coverage, and the public API in Objective-C feels like a translation from another language. A renewed focus on Swift will eliminate a major source of technical debt. We’ll be able to write more idiomatic, less error-prone Swift code without compromising the developer experience in both languages.

Discontinuing Objective-C support will have the following immediate impacts on the codebase:

* `@objc` attributes and `MB` class prefixes will disappear.
* Methods whose sole purpose is to bridge to Objective-C will disappear.
* Enumeration declarations will move from Objective-C headers to Swift files.
* The CMapboxDirections target will be removed from the SPM distribution.
* References to Objective-C compatibility will be removed from readmes and jazzy guides.

We’ll hopefully be able to take advantage of some Swift language features in the process:

* Convert value classes to structures or enumerations with associated types.
* Convert abstract classes to protocols with default implementations.
* Move global constants under structures and classes.
* Replace magic default values with optionals.
* Replace NSCoding with Codable.

/cc @mapbox/navigation-ios

JThramer:
This is the MapboxDirections.swift-specific companion PR to mapbox/mapbox-navigation-ios#2230.

To-do: 
- [x] Adopt `Codable` (#221)

Fixes #134, fixes #228, and fixes #381.


1ec5:
This fixes #381.

1ec5:
…and fixes #134.

Directions Example/ViewController.swift
```diff
@@ -5,7 +5,7 @@ import Mapbox
 
 // A Mapbox access token is required to use the Directions API.
 // https://docs.mapbox.com/help/how-mapbox-works/access-tokens/#creating-and-managing-access-tokens
-let MapboxAccessToken = "<# your Mapbox access token #>"
+let MapboxAccessToken = "pk.eyJ1IjoiamVycmFkIiwiYSI6ImNqMXBleG44bzAwMDAzM3JpYm55ZWh3YWQifQ.ZDMyGh4XLiB505JdxwIo3Q"
```
JThramer: Remove this <now defunct> public token

Sources/MapboxDirections/MBDirectionsResult.swift
```diff
@@ -1,77 +1,130 @@
 import Foundation
-import Polyline
+import struct Polyline.Polyline
 import CoreLocation
+import struct Turf.LineString
 
 
+
+//protocol DirectionsResult: Codable {
```
JThramer: Remove this commented code.

Sources/MapboxDirections/MBDirectionsResult.swift
```diff
@@ -80,37 +133,8 @@ open class DirectionsResult: NSObject, NSSecureCoding {
...
-     - precondition: `coordinates` must be large enough to hold `coordinateCount` instances of `CLLocationCoordinate2D`.
-     
-     - note: This initializer is intended for Objective-C usage. In Swift code, use the `coordinates` property.
-     */
-    @objc open func getCoordinates(_ coordinates: UnsafeMutablePointer<CLLocationCoordinate2D>) {
-        for i in 0..<(self.coordinates?.count ?? 0) {
-            coordinates.advanced(by: i).pointee = self.coordinates![i]
-        }
-    }
+    //    public let coordinates: [CLLocationCoordinate2D]?
```
JThramer: Remove this commented code.

JThramer: Was this an oversight?

Sources/MapboxDirections/MBDirectionsResult.swift
```diff
@@ -119,11 +143,11 @@ open class DirectionsResult: NSObject, NSSecureCoding {
      
      To determine the name of the route, concatenate the names of the route’s legs.
      */
-    @objc public let legs: [RouteLeg]
+    public let legs: [RouteLeg]
     
-    @objc open override var description: String {
-        return legs.map { $0.name }.joined(separator: " – ")
-    }
+    //    open override var description: String {
```
JThramer: Commented code

Sources/MapboxDirections/MBIntersection.swift
```diff
@@ -59,89 +56,87 @@ public class Intersection: NSObject, NSSecureCoding {
...
-            outletRoadClasses = nil
-        }
+    private enum CodingKeys: String, CodingKey {
+        case outletIndexes = "entry"
+        case headings = "bearings"
+        case location
+        case approachIndex = "in"
+        case outletIndex = "out"
+        case lanes
+//        case approachLanes
```
JThramer: remove commented code

Sources/MapboxDirections/MBRouteOptions.swift
```diff
@@ -47,22 +39,47 @@ open class RouteOptions: DirectionsOptions {
...
-     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. This parameter, if set, should be set to `MBDirectionsProfileIdentifierAutomobile`, `MBDirectionsProfileIdentifierAutomobileAvoidingTraffic`, `MBDirectionsProfileIdentifierCycling`, or `MBDirectionsProfileIdentifierWalking`. `MBDirectionsProfileIdentifierAutomobile` is used by default.
-     */
-    public required init(waypoints: [Waypoint], profileIdentifier: MBDirectionsProfileIdentifier? = nil) {
-        super.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
-        self.allowsUTurnAtWaypoint = ![MBDirectionsProfileIdentifier.automobile.rawValue, MBDirectionsProfileIdentifier.automobileAvoidingTraffic.rawValue].contains(self.profileIdentifier.rawValue)
+    private enum CodingKeys: String, CodingKey {
+        case allowsUTurnAtWaypoint
+        case includesAlternativeRoutes
+       // case includesSteps
```
JThramer: remove commented code

Sources/MapboxDirections/MBRouteOptions.swift
```diff
@@ -167,196 +164,110 @@ open class RouteOptions: DirectionsOptions {
...
-            includesExitRoundaboutManeuver == other.includesExitRoundaboutManeuver,
-            roadClassesToAvoid == other.roadClassesToAvoid,
-            alleyPriority == other.alleyPriority,
-            walkwayPriority == other.walkwayPriority,
-            speed == other.speed else { return false }
-        return true
+        return params + super.urlQueryItems
     }
+    
+//    // MARK: NSCopying
```
JThramer: remove commented code

Sources/MapboxDirections/MBVisualInstructionBanner.swift
```diff
@@ -2,107 +2,143 @@ import Foundation
...
+
+//    /**
+//     Initializes a new visual instruction banner object based on the given JSON dictionary representation and a driving side.
+//
+//     - parameter json: A JSON object that conforms to the [primary or secondary banner](https://docs.mapbox.com/api/navigation/#banner-instruction-object) format described in the Directions API documentation.
+//     - parameter drivingSide: The side of the road the user should drive on. This value should be consistent with the containing route step.
+//     */
+//    
+//    public convenience init(json: [String: Any], drivingSide: DrivingSide) {
+//        let distanceAlongStep = json["distanceAlongGeometry"] as! CLLocationDistance
```
JThramer: remove commented code

Sources/MapboxDirections/Match/MBMatch.swift
```diff
@@ -2,113 +2,116 @@ import Foundation
...
+}
 /**
  A `Match` object defines a single route that was created from a series of points that were matched against a road network.
  
  Typically, you do not create instances of this class directly. Instead, you receive match objects when you pass a `MatchOptions` object into the `Directions.calculate(_:completionHandler:)` or `Directions.calculateRoutes(matching:completionHandler:)` method.
  */
-@objc(MBMatch)
 open class Match: DirectionsResult {
+    #warning("check this")
+//    init(matchOptions: MatchOptions, legs: [RouteLeg], tracepoints: [Tracepoint], distance: CLLocationDistance, expectedTravelTime: TimeInterval, coordinates: [CLLocationCoordinate2D]?, confidence: Float, speechLocale: Locale?, waypointIndices: IndexSet) {
```
JThramer: remove commented code, also check that

JThramer: pretty sure we're good here, but should double-check

Sources/MapboxDirections/Match/MBMatch.swift
```diff
@@ -2,113 +2,116 @@ import Foundation
...
+        try container.encode(confidence, forKey: .confidence)
+        try container.encode(tracepoints, forKey: .tracepoints)
+        try container.encode(matchOptions, forKey: .matchOptions)
+        try super.encode(to: encoder)
     }
     
+//    /**
+//     Initializes a new match object with the given JSON dictionary representation and tracepoints.
+//     
+//     - parameter json: A JSON dictionary representation of the route as returned by the Mapbox Map Matching API.
```
JThramer: remove commented code

Sources/MapboxDirections/Match/MBMatch.swift
```diff
@@ -2,113 +2,116 @@ import Foundation
...
-        guard let tracepoints = decoder.decodeObject(of: [NSArray.self, Tracepoint.self], forKey: "tracepoints") as? [Tracepoint] else {
-            return nil
-        }
-        self.tracepoints = tracepoints
-        
-        waypointIndices = decoder.decodeObject(of: NSIndexSet.self, forKey: "waypointIndices") as IndexSet?
-        
-        super.init(coder: decoder)
-    }
+//    public required init?(coder decoder: NSCoder) {
```
JThramer: remove commented code

Sources/MapboxDirections/Match/MBTracepoint.swift
```diff
@@ -5,42 +5,54 @@ import CoreLocation
...
+        try super.encode(to: encoder)
     }
+
     
-    @objc(isEqualToTracepoint:)
-    open func isEqual(to other: Tracepoint?) -> Bool {
-        guard let other = other else { return false }
-        return super.isEqual(to: other) && type(of: self) == type(of: other) &&
-            alternateCount == other.alternateCount
+//    // MARK: Objective-C equality
```
JThramer: remove commented code

Tests/MapboxDirectionsTests/RouteOptionsTests.swift
```diff
@@ -165,8 +119,8 @@ class RouteOptionsTests: XCTestCase {
         let destination = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 39.12971, longitude: -84.51638), name: "UC")
         destination.targetCoordinate = CLLocationCoordinate2D(latitude: 39.13115, longitude: -84.51619)
         let options = RouteOptions(waypoints: [origin, destination])
-        
-        XCTAssertEqual(options.queries, ["-84.47182,39.15031", "-84.51638,39.12971"])
+        XCTAssertEqual(options.coordinates, "-84.47182,39.15031;-84.51638,39.12971")
+//        XCTFail("Get the path")
```
JThramer: remove comment

Sources/MapboxDirections/MBDirectionsResult.swift
JThramer: Answer: No.

Sources/MapboxDirections/MBVisualInstructionType.swift
```diff
@@ -33,39 +33,11 @@ public enum VisualInstructionComponentType: Int, CustomStringConvertible {
...
-            type = .exitCode
-        default:
-            return nil
-        }
-        self.init(rawValue: type.rawValue)
-    }
+    /**
+    A component contains a lane.
+     */
+    case lane
```
1ec5: This case would never occur because the component would be deserialized as a LaneInstructionComponent instead of a VisualInstructionComponent.

Sources/MapboxDirections/Extensions/Codable.swift
```diff
@@ -0,0 +1,70 @@
...
+    
+    var value: Codable? {
+        return t ?? u
+    }
+    
+    var coordinates: [CLLocationCoordinate2D] {
+        if let geo = value as? Geometry {
+            return geo.coordinates
+        } else if let geo = value as? String {
+            return decodePolyline(geo, precision: options.shapeFormat == .polyline6 ? 1e6 : 1e5)!
```
1ec5: The current UncertainCodable implementation (and the associated-value enumeration concept we’re bouncing around) would potentially store a Polyline-encoded string and convert it to coordinates on the fly. Given how frequently the navigation SDK gets `coordinates` on a given step, I’m not sure we want to convert on the fly; it could have a significant impact on certain operations like off-route detection.

Sources/MapboxDirections/Extensions/GeoJSON.swift
```diff
@@ -0,0 +1,49 @@
+import Foundation
+import CoreLocation
+
+struct Geometry: Codable {
```
1ec5: This file’s path implies that it contains extensions to Turf’s GeoJSON types, whereas in fact it defines a new private Geometry type. Move this file somewhere else and call it Geometry.swift.

Sources/MapboxDirections/Extensions/Codable.swift
1ec5: `PolyLineString`, the associated-value enumeration, just stores what the JSON representation does (plus an optional precision). PolyLineString is nothing but an intermediate representation to facilitate conversion from the JSON representation to LineString-backed storage.

Sources/MapboxDirections/MBIntersection.swift
```diff
@@ -1,16 +1,14 @@
 import Foundation
-import CoreLocation
```
1ec5: SPM builds are failing because Swift for Linux doesn’t include Core Location. (Apparently penguin tracking isn’t a particularly commercially viable proposition.)

```
/Users/distiller/project/Sources/MapboxDirections/MBIntersection.swift:11:26: error: use of undeclared type 'CLLocationCoordinate2D'
    public let location: CLLocationCoordinate2D
                         ^~~~~~~~~~~~~~~~~~~~~~
```

Sources/MapboxDirections/Extensions/Codable.swift
```diff
@@ -0,0 +1,55 @@
...
+            let encodedPolyline = encodeCoordinates(lineString.coordinates, precision: precision)
+            self = .polyline(encodedPolyline, precision: precision)
+        }
+    }
+    
+    init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let options = decoder.userInfo[.options] as? DirectionsOptions
+        switch options?.shapeFormat {
+        case .geoJSON:
```
1ec5: We need to make sure the SPM build bot is building in Swift 5:

```
/Users/distiller/project/Sources/MapboxDirections/Extensions/Codable.swift:37:15: error: enum case 'geoJSON' not found in type 'RouteShapeFormat?'
        case .geoJSON:
              ^
```

Sources/MapboxDirections/MBDirections.swift
```diff
@@ -129,143 +137,238 @@ open class Directions: NSObject {
...
-        let task = dataTask(forCalculating: options, completionHandler: { (json) in
+        let completion = completionHandler as MapboxRouteCompletionHandler
+        let fetchStart = Date()
+        let request = urlRequest(forCalculating: options)
+        let requestTask = URLSession.shared.dataTask(with: request) { (possibleData, possibleResponse, possibleError) in
             let responseEndDate = Date()
-            let response = options.response(from: json)
-            if let routes = response.1 {
-                self.postprocess(routes, fetchStartDate: fetchStartDate, responseEndDate: responseEndDate, uuid: json["uuid"] as? String)
+            guard let response = possibleResponse, ["application/json", "text/html"].contains(response.mimeType) else {
```
1ec5: Why does this method accept HTML-formatted responses while the other `calculate` methods only accept JSON?

Directions Example/ViewController.swift
1ec5: mapbox/MapboxDirections.swift#385 implements the recommended approach for injecting the access token from a non-version-controlled location, to head off mishaps like this in the future.

Tests/MapboxDirectionsTests/RouteStepTests.swift
```diff
@@ -106,59 +106,125 @@ class RoadTests: XCTestCase {
...
-        XCTAssertEqual(unarchivedStep.destinations ?? [], step.destinations ?? [])
-        XCTAssertEqual(unarchivedStep.instructionsSpokenAlongStep ?? [], step.instructionsSpokenAlongStep ?? [])
-        XCTAssertEqual(unarchivedStep.instructionsDisplayedAlongStep ?? [], step.instructionsDisplayedAlongStep ?? [])
-        XCTAssertEqual(unarchivedStep.drivingSide, DrivingSide.left)
+        XCTAssert(roundTripJSON == pass)
+    }
+}
+
+
+fileprivate let routeStepJSON = """
```
1ec5: This massive multiline-stringified JSON source code should be reverted back to what we had before. The aim is to unit-test RouteStep, not integration-test nearly the entire library.

1ec5:
48fab7b0d99ea297f3d36429a01d64c59382e819 consolidates Component, ComponentSerializer, VisualInstructionComponent, and LaneIndicationComponent into a single codable associated-value enumeration, VisualInstruction.Component. Each case of the enumeration other than `.lane` has a TextRepresentation and/or ImageRepresentation. This structure accurately models the mutual exclusivity of several of the properties in the JSON object.

Sources/MapboxDirections/MBVisualInstructionComponent.swift
```diff
@@ -79,27 +79,30 @@ public extension VisualInstruction {
             let container = try decoder.container(keyedBy: CodingKeys.self)
             let kind = try container.decode(Kind.self, forKey: .kind)
             
-            let text = try container.decode(String.self, forKey: .text)
+            let text = try container.decodeIfPresent(String.self, forKey: .text)
```
1ec5: `text` is documented as being required in the API response, so we should be calling `decode(_:forKey:)` here.

1ec5: Ah, right, lane components don’t have `text`. That’s an omission in the documentation.

Sources/MapboxDirections/MBRoute.swift
```diff
@@ -1,82 +1,28 @@
...
-            coordinates = decodePolyline(geometry, precision: 1e6)!
-        default:
-            coordinates = nil
-        }
-        
-        self.init(legs: [leg], distance: distance, expectedTravelTime: expectedTravelTime, coordinates: coordinates, speechLocale: nil, options: options)
+    public override var directionsOptions: DirectionsOptions {
+        return routeOptions as DirectionsOptions
     }
+    public var routeOptions: RouteOptions
```
JThramer: Document this.

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+
+class MapMatchingResponse: Decodable {
```
1ec5: This name is confusing: MapMatchingResponse and MatchResponse both decode Map Matching API responses, but MapMatchingResponse produces Routes while MatchResponse produces Matches. Each API should have only one response decoder, and it should be up to the `calculate(_:completionHandler:)` method or some helper method to convert Matches to Routes.

The next step after that would be to make MatchResponse and RouteResponse public and return those objects in the completion handlers. That would give us a place to put stuff like response timestamps and UUIDs instead of duplicating them on each individual result.

Sources/MapboxDirections/Match/MBMatch.swift
```diff
@@ -1,114 +0,0 @@
...
-     
-     Any outlier tracepoint is omitted from the match. This array represents an outlier tracepoint is a `Tracepoint` object whose `Tracepoint.coordinate` property is `kCLLocationCoordinate2DInvalid`.
-     */
-    @objc open var tracepoints: [Tracepoint]
-    
-    
-    /**
-     Index of the waypoint inside the matched route.
-     */
-    @objc open var waypointIndices: IndexSet?
```
1ec5: This property is distinct from the `DirectionsOptions.waypointIndices` property that was justifiably removed. This property is meant to indicate which of the items in `tracepoints` is a (leg-separating) waypoint. If we remove this property, then we need to make sure the relevant tracepoints’ `separatesLegs` properties are set.

Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -0,0 +1,182 @@
...
+    
+    // MARK: Creating a Directions Result
+    
+    public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        legs = try container.decode([RouteLeg].self, forKey: .legs)
+        distance = try container.decode(CLLocationDistance.self, forKey: .distance)
+        expectedTravelTime = try container.decode(TimeInterval.self, forKey: .expectedTravelTime)
+        
+        _directionsOptions = decoder.userInfo[.options] as! DirectionsOptions
```
1ec5: This PR doesn’t completely fix #228 because it still contains some force-unwrapping of user info dictionary entries. This is problematic because decoding from JSON is now part of the public API for all types. Any user info dictionary entries we rely on for decoding need to be documented as preconditions, and the absence of them should throw an error instead of crashing. Both the error-throwing and documentation are addressed in #388, but we should try to bring those changes over to this PR because they’re more relevant to the decoding changes that are taking place here.

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
1ec5: → #391

Sources/MapboxDirections/DirectionsResult.swift
1ec5: c5e733ddd67e80a18316b133823982298f7dcdb0 brings over the error-throwing and documentation from #388.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -0,0 +1,254 @@
...
+        
+        switch kind {
+        case .delimiter:
+            self = .delimiter(text: textRepresentation)
+        case .text:
+            self = .text(text: textRepresentation)
+        case .image:
+            var imageBaseURL: URL?
+            if let imageBaseURLString = try container.decodeIfPresent(String.self, forKey: .imageBaseURL) {
+                imageBaseURL = URL(string: imageBaseURLString)
```
1ec5: This test fails because `imageBaseURLString` goes out of scope, somehow making `imageBaseURL` nil too.

```
Test Case '-[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent]' started.
/Users/distiller/project/Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift:39: error: -[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent] : XCTAssertEqual failed: ("Optional("https://s3.amazonaws.com/mapbox/shields/v3/us-42@1x.png")") is not equal to ("Optional("https://s3.amazonaws.com/mapbox/shields/v3/us-42@1x.svg")")
/Users/distiller/project/Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift:40: error: -[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent] : XCTAssertEqual failed: ("Optional("https://s3.amazonaws.com/mapbox/shields/v3/us-42@3x.png")") is not equal to ("Optional("https://s3.amazonaws.com/mapbox/shields/v3/us-42@3x.svg")")
Test Case '-[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent]' failed (0.010 seconds).
```

I think it may be an optimization bug in the compiler, because if I change the code to this instead, the debugger is unable to resolve `imageBaseURL` within the `if` statement:

```swift
            var imageRepresentation: ImageRepresentation
            if let imageBaseURLString = try container.decodeIfPresent(String.self, forKey: .imageBaseURL) {
                let imageBaseURL = URL(string: imageBaseURLString)
                imageRepresentation = ImageRepresentation(imageBaseURL: imageBaseURL)
            } else {
                imageRepresentation = ImageRepresentation(imageBaseURL: nil)
            }
```

1ec5: I got around that by decomposing the code a bit more, but something about this String is still tripping up `URL(string:)`, causing it to create a URL structure that itself is `nil` even though all its members are present. Meanwhile, **NS**URL is copacetic:

```
(lldb) p imageBaseURLString
(String) $R22 = "https://s3.amazonaws.com/mapbox/shields/v3/us-42"
(lldb) p URL(string: imageBaseURLString)
(URL?) $R24 = nil
(lldb) p URL(string: imageBaseURLString)?.absoluteString
(String?) $R26 = "https://s3.amazonaws.com/mapbox/shields/v3/us-42"
(lldb) po NSURL(string: imageBaseURLString)
▿ Optional<NSURL>
  - some : https://s3.amazonaws.com/mapbox/shields/v3/us-42
```

1ec5: I can easily reproduce the issue with similar test code anywhere in the example application:

```swift
        let componentJSON = [
            "text": "US 42",
            "type": "icon",
            "imageBaseURL": "https://s3.amazonaws.com/mapbox/shields/v3/us-42",
        ]
        let componentData = try! JSONSerialization.data(withJSONObject: componentJSON, options: [])
        let component = try! JSONDecoder().decode(VisualInstruction.Component.self, from: componentData)
        switch component {
        case .image(let image, _):
            image.imageBaseURL
            image.imageBaseURL?.absoluteString
            image.imageURL(scale: 1, format: .svg)
            image.imageURL(scale: 1, format: .svg)?.absoluteString
        default:
            preconditionFailure()
        }
```

The issue also reproduces if I copy the entire VisualInstruction type hierarchy into the example application as MyVisualInstruction and run the same code. However, I’m unable to reproduce the issue by copying the same code into a playground independent of MapboxDirections.swift.

1ec5: @fabian-guerra pointed out that the actual failure was a file extension mismatch rather than the URL being `nil`, which led to the fix in 53c71654c96c1b47f6910905391c220bb7ff40e7.

What’s still weird is that, when stopped inside `VisualInstruction.Component.init(decoder:)`, the debugger always prints `URL?` instances as `nil` and `URL` instances as `{}`, even if they actually have `absoluteString` set. I’ve been unable to reproduce this issue outside of that specific context, however.

Tests/MapboxDirectionsTests/WaypointTests.swift
```diff
@@ -103,4 +103,36 @@ class WaypointTests: XCTestCase {
...
+    func testTracepointEquality() {
+        let left = Tracepoint(coordinate: CLLocationCoordinate2D(), countOfAlternatives: 0, name: nil)
+        XCTAssertEqual(left, left)
+        
+        let right = Tracepoint(coordinate: CLLocationCoordinate2D(), countOfAlternatives: 0, name: nil)
+        XCTAssertEqual(left, right)
+        
+        // FIXME: Only Waypoint.==(_:_:) ever gets called: <https://stackoverflow.com/a/28794214/4585461>. This will be moot once Tracepoint becomes a struct that doesn’t inherit from Waypoint: <https://github.com/mapbox/MapboxDirections.swift/pull/388>.
+//        right = Tracepoint(coordinate: CLLocationCoordinate2D(latitude: 1, longitude: 1), countOfAlternatives: 0, name: nil)
+//        XCTAssertNotEqual(left, right)
```
1ec5: Unfortunately, it isn’t currently possible to fully compare two Tracepoints; the comparison only considers Waypoint properties. It [isn’t possible for a subclass to override `==`](https://stackoverflow.com/a/28794214/4585461) in the same manner that we had been used to overriding `isEqual(_:)`. Even comparing `Tracepoint() == Tracepoint()` calls `Waypoint.==(_:_:)` directly without involving `Tracepoint.==(_:_:)`, as seen in the code coverage report.

Fortunately, #388 would address this design flaw by converting Waypoint and Tracepoint to structs and relating them through a protocol rather than inheritance.

Unfortunately, this problem also affects the abstract DirectionsResult class. Equality is very important to a library such as MapboxDirections.swift. This makes the case for abolishing abstract classes in favor of protocols even stronger.

1ec5: #392 tracks eliminating the DirectionsOptions and DirectionsResult abstract classes.

codecov.yml
```diff
@@ -1,6 +1,6 @@
 ignore:
   - "Directions Example"
-  - "Tests"
+  - "MapboxDirectionsTests"
```
1ec5: #360 originally intended to omit MapboxDirectionsTests from code coverage metrics, but when #362 renamed this folder to Tests, test code inflated test coverage by around 11%. We’ll need to adjust reapply this change in a separate commit outside of this PR to avoid getting dinged by the (justifiable) omission of test code from code coverage metrics.

Issue #386, opened 2019-11-03
Match.tracepoints is inaccurate when there are submatches


1ec5:
When none of the input coordinates in MatchOptions separates legs, and the Map Matching API isn’t confident about a match between two input coordinates, it terminates one match at the first input coordinate, adds multiple “submatches” to the next input coordinate, and finally adds a match from that input coordinate onward.

This library gained Map Matching API compatibility to support a “[Bring Your Own Route](https://blog.mapbox.com/take-the-scenic-route-with-our-new-bring-your-own-route-feature-d3500a6a91c3)” workflow in which the API’s response is converted into Route objects, as if it came from the Directions API. The assumption is that each match spans the entire distance between the first and last input coordinates, but that isn’t necessarily what the API returns. If the API returns submatches, every Match object’s `tracepoints` property is inaccurate and there’s no way to reconstruct what the tracepoints should be.

/cc @mapbox/navigation-ios

PR #388, opened 2019-11-14
Revamp waypoint types

### LINKED ISSUE #333: Add snapped distance property to Waypoint
Directions API v5 has added a new `distance` property to Waypoint objects in the response, as requested in #320. This property contains the straight-line distance from the requested waypoint coordinate to the returned waypoint coordinate, which is snapped to the route line. The distance can already be calculated regardless of this property, but the API provides it for consistency with the rest of the response, to ensure that the same distance formula is used throughout.

We’ll need to add a new instance property to the Waypoints class. I think `distance` would be an ambiguous name, considering that the same class is used for both requests and responses. Perhaps we should create a subclass of Waypoint specifically for the returned waypoint data, much like the Tracepoint class for Map Matching API responses. Then we could indicate both the requested and returned coordinates, as well as the `snappedDistance`. Here’s where we create the Waypoints as part of the serialized response:

https://github.com/mapbox/MapboxDirections.swift/blob/c350b60489c33845551cf0163a5481d32c940479/MapboxDirections/MBRouteOptions.swift#L168

Technically we reuse the requested waypoints wholesale if none are provided in the response, but that only happens when using Directions API v4:

https://github.com/mapbox/MapboxDirections.swift/blob/c350b60489c33845551cf0163a5481d32c940479/MapboxDirections/MBRouteOptions.swift#L172

1ec5:
Replaced the Waypoint and Tracepoint classes with separate classes for requests and responses to better match the design of the Directions API and Map Matching API and avoid the messy situation of classes with multiple mutually exclusive properties.

Now you create a RouteOptions or MatchOptions using a series of DirectionsOptions.Waypoint instances (also known as RouteOptions.Waypoint or MatchOptions.Waypoint). The `RouteCompletionHandler` and `MatchCompletionHandler` closures and the response types represent waypoints as DirectionsResult.Waypoint (also known as RouteOptions.Waypoint) and tracepoints as `Match.Tracepoint`.

Also renamed the `Tracepoint.alternateCount` property to `Tracepoint.countOfAlternatives` and added the snapped distance to the DirectionsResult.Waypoint type.

Fixes #333.

/cc @mapbox/navigation-ios


1ec5:
Some tests are failing because the test fixtures need to updated. They were generated back before the Directions and Map Matching APIs started returning `distance` as part of each snapped waypoint.

Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -29,7 +29,10 @@ open class DirectionsResult: Codable {
         distance = try container.decode(CLLocationDistance.self, forKey: .distance)
         expectedTravelTime = try container.decode(TimeInterval.self, forKey: .expectedTravelTime)
         
-        _directionsOptions = decoder.userInfo[.options] as! DirectionsOptions
+        guard let directionsOptions = decoder.userInfo[.options] as? DirectionsOptions else {
+            throw DirectionsCodingError.missingOptions
```
1ec5: This PR continues to assume that options are passed in as a precondition for decoding, but it throws errors instead of force-unwrapping dictionary entries. Ideally, #382 would incorporate these changes to fix #228.

1ec5:
A major problem with converting Waypoint to a struct is that a given waypoint appears in multiple places in the response. The solution would be to move `RouteLeg.source` to Route and, once #391 is implemented, make `Response.waypoints` a computed property based on `Route.source` and `RouteLeg.destination`.

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
...
+
+* Replaced the `Waypoint` and `Tracepoint` classes with separate classes for requests and responses. Now you create a `RouteOptions` or `MatchOptions` using a series of `DirectionsOptions.Waypoint` instances (also known as `RouteOptions.Waypoint` or `MatchOptions.Waypoint`). The `RouteCompletionHandler` and `MatchCompletionHandler` closures and the response types represent waypoints as `DirectionsResult.Waypoint` (also known as `RouteOptions.Waypoint`) and tracepoints as `Match.Tracepoint`. ([#382](https://github.com/mapbox/MapboxDirections.swift/pull/382))
+* Replaced the `Route.coordinates` property with `Route.shape` and the `RouteStep.coordinates` property with `RouteStep.shape`. The `Route.coordinateCount` and `RouteStep.coordinateCount` properties have been removed, but you can use the `LineString.coordinates` property to get the array of `CLLocationCoordinate2D`s. ([#382](https://github.com/mapbox/MapboxDirections.swift/pull/382))
+* `RouteLeg.source` and `RouteLeg.destination` are now optional. They can be `nil` when the `RouteLeg` object is decoded individually from JSON. ([#382](https://github.com/mapbox/MapboxDirections.swift/pull/382))
+* The `
```
1ec5: Finish this sentence.

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -36,7 +36,8 @@ class MapMatchingResponse: Decodable {
             // Postprocess each route.
             for route in routes {
                 // Imbue each route’s legs with the leg-separating waypoints refined above.
-                route.legSeparators = waypoints.filter { $0.separatesLegs }
+                // TODO: Filter these waypoints by whether they separate legs, based on the options, if given.
```
Udumft, line 39: @1ec5 , so here we should filter out `waypoints` so that we leave only ones corresponding `options.legSeparators` (if options are present)? Is it correct?

Sources/MapboxDirections/MapMatching/MatchResponse.swift
```diff
@@ -17,12 +17,14 @@ class MatchResponse: Codable {
...
+        tracepoints = try container.decodeIfPresent([Match.Tracepoint?].self, forKey: .tracepoints)
         matches = try container.decodeIfPresent([Match].self, forKey: .matches)
         
-        if let points = self.tracepoints {
-            matches?.forEach {
-                $0.tracepoints = points
+        if let tracepoints = self.tracepoints, let matches = matches {
+            for match in matches {
+                // TODO: Filter on matchings_index.
+                // TODO: Push tracepoints down to individual legs to reflect waypoint_index.
```
Udumft, line 26: Can someone point me to some kind of a doc or explanation about `tracepoints` in the response and `matchings_index`, `waypoint_index` too?

CHANGELOG.md
Udumft: Some context would be helpful to do it :)

1ec5: I think I had been going through the PR looking for breaking changes to call out but got distracted. If you see anything else that has changed in the public API besides the first item, feel free to add it here; otherwise, we can remove this item.

Sources/MapboxDirections/MapMatching/MatchResponse.swift
1ec5: Here’s the [Map Matching API documentation](https://docs.mapbox.com/api/navigation/#map-matching), which is more relevant than before, now that this PR hews more closely to the Map Matching API response format.

It isn’t necessary to fix #386 as part of this PR, but I think understanding that issue helped me understand these properties better.

Sources/MapboxDirections/MapMatching/Tracepoint.swift
```diff
@@ -1,40 +1,47 @@
...
+        public var correction: CLLocationDistance
+        
+        // MARK: Determining the Degree of Confidence
+        
+        /**
+         Number of probable alternative matchings for this tracepoint. A value of zero indicates that this point was matched unambiguously.
+         */
+        public var countOfAlternatives: Int
+        
+        public var name: String? // ??? matching response otherwise won't be able to decode waypoints names.
```
Udumft, line 30: If we don't supply a `name` in a `Tracepoint`, Map Matching response decoding won't pass this data to `RouteResponse` and in it's turn resulting `Waypoint`s won't have name too.  I see that `Tracepoint` is identical to `Route.Waypoint`, why do we need this struct at all?

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -27,15 +27,15 @@ public struct RouteResponse {
 
 extension RouteResponse: Codable {
     enum CodingKeys: String, CodingKey {
-        case code
-        case message
-        case error
+        case code // << do we need these?
+        case message // <<
+        case error // <<
```
Udumft, line 32: These values are now decoded using `ResponseDisposition`, but It is only used internally while parsing the response and is not passed to the user. I assume this is what we want, and these keys inside `RouteResponse` are not needed?

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -91,37 +91,63 @@ extension RouteResponse: Codable {
...
+                    if decodedWaypoint == waypoints.last! ||
+                        decodedWaypoint == waypoints.first! {
+                        return decodedWaypoint
+                    }
+                    
+                    guard let index = optionsLegSeparators.firstIndex(of: waypointInOptions) else {
+                        return nil
+                    }
+                    optionsLegSeparators.remove(at: index)
+                    return decodedWaypoint
```
Udumft, line 142: I hope I understood correctly how filtering should be applied. I decided to stick to `options.legSeparators` and filter out decoded waypoints that correspond it.

Sources/MapboxDirections/MapMatching/MatchResponse.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+class MatchResponse: Codable {
```
Udumft, line 3: `MatchResponse` seems removed in some other PR. I assume it means that that tracepoint filtering should occur in `MapMatchingResponse` instead?

Sources/MapboxDirections/MapMatching/MatchResponse.swift
Udumft: According to that doc, our `Match` and `Tracepoint` objects do not reflect the structure described. Is there any special reason why? Or that's what I can safely take care of?

Sources/MapboxDirections/MapMatching/Tracepoint.swift
```diff
@@ -1,40 +1,58 @@
...
+        
+        /***
+         The index of the match object in matchings that the sub-trace was matched to.
+         */
+        public var matchingIndex: Int = 0 // TODO: not sure about default value
+        
+        /***
+         The index of the waypoint inside the matched route.
+         */
+        public var waypointIndex: Int = 0 // TODO: not sure about default value
```
Udumft, line 39: There were not originally in the doc, but looks like it is required to have.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -58,12 +58,28 @@ extension RouteResponse: Codable {
...
+                for (index, route) in routes.enumerated() {
+                    let legSeparators = sorted.filter {
+                        return $0.0.matchingIndex == index
+                    }.map { $0.1 }
+                    
+                    assert(legSeparators.count == route.legs.count + 1)
+                    
+                    route.legSeparators = legSeparators
+                }
+            }
```
Udumft, line 82: This part is how I understand [Map Matching Doc](https://docs.mapbox.com/api/navigation/#map-matching). The idea is to set leg separators explicitly via `matchings_index` and `waypoint_index`. Not sure if it is required though or will have any effect at all since `routes` and `waypoints` are passed to the user as public properties and waypoints are not modified in this code.
A help/review would be required here.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -58,12 +58,28 @@ extension RouteResponse: Codable {
...
+            if let routes = routes {
+                let sorted = zip(filtered, waypoints!).sorted {
+                    $0.0.waypointIndex < $1.0.waypointIndex
+                }
+                for (index, route) in routes.enumerated() {
+                    let legSeparators = sorted.filter {
+                        return $0.0.matchingIndex == index
+                    }.map { $0.1 }
+                    
+                    assert(legSeparators.count == route.legs.count + 1)
```
Udumft, line 78: I assume this is something we can expect to be guaranteed?

Sources/MapboxDirections/RouteResponse.swift
Udumft: Still actual

Sources/MapboxDirections/MapMatching/MatchResponse.swift
1ec5: The Map Matching API response format avoids repetition to minimize data transfer, and it’s optimized for a file format (JSON) that has no concept of pointer references. As a result, the response is rife with parallel structures and indices. At the same time, the Directions API response is influenced by the OSRM Router service’s heavily nested response format (maneuvers inside steps inside legs inside routes). As part of developing the “bring your own route” workflow, the Map Matching API adopted most of the Directions API’s format so that the client could duck-type matches as routes. So Map Matching API responses end up with a mix of indexable parallel structures in some places and repetitive, nested structures in other places.

Originally, this library only supported the Directions API, so it adopted that API’s highly nested structure. We also avoided index-based structures, even where the Directions API had them, because of potential out-of-bounds issues and other misuse. Pointer references are more convenient to work with than indices and don’t result in excessive memory consumption. It’s also much more convenient for objects to be self-contained: if the application only cares about the match, it can pass around or archive an individual Match object instead of the entire MatchResponse plus a match index.

In the long term, I think we will want to go all-in on indexed parallel structures, once a future version of the Directions API adopts Valhalla’s more consistently indexed-based structure. When we do that, things like the `RouteLeg.segmentRangesByStep` property added in mapbox/mapbox-directions-swift#250 will be important for avoiding out-of-bounds issues.

But in the short term, we should keep pushing things like `tracepoints` down into the individual matches so that the matches are more self-contained.

This is a good example of how I think we want to be deliberate about where this library’s types resemble or differ from the API response. The API response format has a lot of warts, having bent over backwards to preserve backwards compatibility, so we only want to copy what make sense. We prioritized refactoring the waypoint/tracepoint types because they were too different from the corresponding API response objects, leading to some questionable patterns like treating `RouteOptions.waypoints` like an inout parameter. But I tried to avoid churning other classes were we differed in more minor or more deliberate ways.

Sources/MapboxDirections/MapMatching/Tracepoint.swift
1ec5: A waypoint (or tracepoint) name coming from the API isn’t actually that meaningful. Normally it’s just the name of the street that the waypoint got snapped to. If the application supplied a waypoint name as part of the request, that name is worth preserving. In the past, we’d override the API response’s waypoint names with the request’s waypoint names, where available. Now it’s up to the application to hang onto the request and associate the request’s waypoints with the response’s waypoints in some cases.

Sources/MapboxDirections/RouteResponse.swift
1ec5: I think you’re right, because #418 turns the three fields into an Error in the Result.

Sources/MapboxDirections/MapMatching/Tracepoint.swift
1ec5: These indices are parallel state that can get out of sync with the array of matches or waypoints. This starts to get into #386 territory. I think the filtering suggested in that issue could make these properties unnecessary.

Sources/MapboxDirections/MapMatching/MatchResponse.swift
1ec5: At the time this PR was originally written, MatchResponse interpreted Map Matching API responses as Map Matching API responses (MatchOptions→Match), whereas MapMatchingResponse interpreted Map Matching API responses as Directions API responses (MatchOptions→Route). It was confusing, so #406 eliminated MapMatchingResponse and renamed MatchResponse to MapMatchingResponse.

Sources/MapboxDirections/RouteResponse.swift
1ec5: There will be fewer leg separators if any of the waypoints in the request had `separatesLegs` equal to `false`.

Sources/MapboxDirections/MapMatching/Tracepoint.swift
Udumft: Sorry, it's a bit misleading: "that name is worth preserving" and "it’s up to the application to hang onto the request and associate the request’s waypoints with the response’s waypoints" seems to be the opposite statements.
If we are to preserve waypoint name returned by the API (and apparently `Route.Waypoint` is), `Tracepoint` should have a `name`. Otherwise, `Waypoints` won't be decoded correctly as they depend on `Tracepoint` data: https://github.com/mapbox/mapbox-directions-swift/pull/388/files#diff-a2198eb412da79c9d8351c4e811b7dc6R63-R66

Sources/MapboxDirections/MapMatching/MatchResponse.swift
Udumft: OK, so due to some reasonable matters like API structure and JSON format we may not follow the doc strictly but instead "adjust" info we provide to users to have some adequate form. 
At the same time, to correctly "keep pushing things like tracepoints down into the individual matches" we need to follow the doc and decode `Tracepoint`'s `matchings_index` and `waypoint_index`... I see that previous implementation just copied all tracepoints to all matches, which seem to work only while there is a single `Match` provided (unless it was removed [here](https://github.com/mapbox/mapbox-directions-swift/commit/48abd96069cbe131f076040b0e1bb3c04ef48bf8#diff-c28392baf0208871bf1ac5c51e283fe1L50-L54)).
We may just restore this behavior, but why not improve it to make it "more safe".

Udumft:
`distance` property added to Waypoint, `Waypoint` and `Tracepoint` classes replaced with structures. Did I miss anything else should be done here?

Issue #391, opened 2019-11-23
Encapsulate responses in coherent public types


1ec5:
Currently, `RouteCompletionHandler` and `MatchCompletionHandler` accept multiple arguments that separate the routes (or matches) from the waypoints they connect. Because these arrays are passed in as top-level arguments to the completion handlers, adding more detail about the response would break backwards compatibility. For this reason, we’ve taken to copying metadata about the response as a whole – namely, the response UUID and options object from the request – into each route or match.

As part of adopting Codable, #382 implements private MapMatchingResponse, MatchResponse, and RouteResponse classes that represent the response as a whole. These classes give us a convenient, central place to put response metadata and neatly parallel MapKit’s `MKDirections.Response` class. We should make these new classes public and pass them to the completion handler directly instead of dissassembling them for public consumption.

At the same time, we should unify the response classes into two classes: MapMatchingResponse would decode Map Matching API responses (as MapMatchingResponse and MatchResponse both currently do), while RouteResponse would decode Directions API responses. For the case that MapMatchingResponse is meant to handle, it would be better for MapMatchingResponse to produce Match objects; `calculateRoutes(matching:completionHandler:)` could call a new initializer on RouteResponse that takes a MapMatchingResponse as input.

This refactoring is probably going to be necessary once #388 separates Waypoint into three separate types.

* [x] Implement decodable response types: #382
* [x] Make RouteResponse and MatchResponse public: #393
* [ ] Take MapMatchingResponse private and internally convert it to RouteResponse
* [ ] Rename MatchResponse to MapMatchingResponse
* [ ] Document RouteResponse and MatchResponse and each of their members
* [ ] Fold an optional error property into RouteResponse and MatchResponse and decode the error as part of decoding these types
* [ ] Make RouteResponse and MatchResponse conform to Encodable?
* [ ] Replace the RouteCompletionHandler and MatchCompletionHandler arguments with RouteResponse and MatchResponse

/cc @mapbox/navigation-ios

1ec5:
Fixed in #406.

PR #393, opened 2019-12-16
Objective-C removal followups for navigation SDK compatibility

JThramer:
This PR aims to fix some loose ends found while working on https://github.com/mapbox/mapbox-navigation-ios/pull/2275. It is a work-in-progress.

/cc @mapbox/navigation-ios


Sources/MapboxDirections/Lane.swift
```diff
@@ -3,15 +3,15 @@ import Foundation
 /**
  A lane on the road approaching an intersection.
  */
-struct Lane: Equatable {
+public struct Lane: Equatable {
```
1ec5: `Lane` was intentionally removed from the public API. Use `VisualInstruction.Component.lane(indications:isUsable:)` instead.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -1,12 +1,12 @@
...
-    let uuid: String?
-    let routes: [Route]?
-    let waypoints: [Waypoint]?
+public struct RouteResponse {
+    public var code: String?
+    public var message: String?
+    public var error: String?
+    public let uuid: String?
+    public let routes: [Route]?
+    public let waypoints: [Waypoint]?
```
1ec5: This change requires us to address all of #391. Are you sure we should do so before integrating the navigation SDK with the first round of Swifitification in this library?

Sources/MapboxDirections/Directions.swift
```diff
@@ -141,6 +141,7 @@ open class Directions: NSObject {
     
     // MARK: Getting Directions
     
+    internal typealias RouteCompletionHandlerInteral = (_ waypoints: [Waypoint]?, _ routes: [Route]?, _ error: DirectionsServiceError?) -> Void
```
1ec5: Not sure what needs to be internal about this. Isn’t DirectionsServiceError already public?

Issue #395, opened 2019-12-21
testGeoJSON() fails intermittently (probably slow decoding)


1ec5:
`V5Tests.testGeoJSON()` has been failing occasionally for months, but ever since #382 landed, it has been [failing](https://circleci.com/gh/mapbox/MapboxDirections.swift/1231) much more frequently:

```
Test Case '-[MapboxDirectionsTests.V5Tests testGeoJSON]' started.
/Users/distiller/project/Tests/MapboxDirectionsTests/V5Tests.swift:64: error: -[MapboxDirectionsTests.V5Tests testGeoJSON] : Asynchronous wait failed: Exceeded timeout of 5 seconds, with unfulfilled expectations: "calculating directions should return results".
/Users/distiller/project/Tests/MapboxDirectionsTests/V5Tests.swift:65: error: -[MapboxDirectionsTests.V5Tests testGeoJSON] : XCTAssertNil failed: "Error Domain=com.apple.XCTestErrorDomain Code=0 "(null)"" - Error: Error Domain=com.apple.XCTestErrorDomain Code=0 "(null)"
/Users/distiller/project/Tests/MapboxDirectionsTests/V5Tests.swift:73: error: -[MapboxDirectionsTests.V5Tests testGeoJSON] : XCTAssertNotNil failed
Test Case '-[MapboxDirectionsTests.V5Tests testGeoJSON]' failed (5.306 seconds).
```

This is most likely a symptom of the Codable performance issue raised in https://github.com/mapbox/MapboxDirections.swift/pull/221#issuecomment-353057526. The issue lies upstream in Turf, which this library relies on to decode LineStrings, and technically [further upstream in the Swift standard library](https://bugs.swift.org/browse/SR-6252).

This isn’t necessarily a showstopper, since the navigation SDK uses Polyline encoding by default instead of GeoJSON. However, we should find a way to get performance back to within the same ballpark as it was before adopting Codable. If it’s already in the same ballpark, then we should increase the timeout and also add a performance unit test to make any runtime changes more apparent.

/cc @mapbox/navigation-ios @frederoni

1ec5:
I’m not sure if equality comparisons are contributing to the test’s flakiness, but we could improve equality performance by storing Polyline-encoded representations of `Route.shape` and `RouteStep.shape` and comparing those strings instead of the full LineString structs.

[...]

PR #404, opened 2020-01-06
Speed up CI preparations with Carthage binary

frederoni:
Instead of installing Carthage with brew, we can install Carthage binary now that
Swift’s ABI is stable.

Installing Carthage now takes 10 seconds compared to 10 minutes.

I also converted the step-library to commands. IIRC, commands are just a newer version of the step library with support for parameters.

Carthage finds the cache for Mapbox SDK but not the dependencies that are built from source. I think it's just a mismatch that will be fixed by bumping the cache key.

cc @1ec5 @JThramer

### DIFF
--- a/Tests/MapboxDirectionsTests/V5Tests.swift
+++ b/Tests/MapboxDirectionsTests/V5Tests.swift
@@ -61,7 +61,7 @@ class V5Tests: XCTestCase {
         }
         XCTAssertNotNil(task)
         
-        waitForExpectations(timeout: 5) { (error) in
+        waitForExpectations(timeout: 10) { (error) in
             XCTAssertNil(error, "Error: \(error!)")
             XCTAssertEqual(task.state, .completed)
         }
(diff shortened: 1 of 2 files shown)


1ec5:
> Installing Carthage now takes 10 seconds compared to 10 minutes.
> Carthage finds the cache for Mapbox SDK but not the dependencies that are built from source. I think it's just a mismatch that will be fixed by bumping the cache key.

I’ve tried rerunning the builds but the “Dependencies” step continues to take at least 8&nbsp;minutes each time. Still, it’s a lot better than the 15&nbsp;minutes or more I had been seeing on recent PRs.

Tests/MapboxDirectionsTests/V5Tests.swift
```diff
@@ -61,7 +61,7 @@ class V5Tests: XCTestCase {
         }
         XCTAssertNotNil(task)
         
-        waitForExpectations(timeout: 5) { (error) in
+        waitForExpectations(timeout: 10) { (error) in
```
1ec5: This more than addresses #395. 😅

PR #405, opened 2020-01-10
Draft PR for adding guidance view support to data model

avi-c:


Directions Example/ViewController.swift
```diff
@@ -67,8 +67,6 @@ class ViewController: UIViewController, MBDrawingViewDelegate {
         let wp2 = Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.8977, longitude: -77.0365), name: "White House")
         let options = RouteOptions(waypoints: [wp1, wp2])
         options.includesSteps = true
-        options.routeShapeResolution = .full
-        options.attributeOptions = [.congestionLevel, .maximumSpeedLimit]
```
1ec5: Remember to revert these changes before merging.

Sources/MapboxDirections/VisualInstructionBanner.swift
```diff
@@ -27,6 +28,7 @@ open class VisualInstructionBanner: Codable {
         primaryInstruction = primary
         secondaryInstruction = secondary
         tertiaryInstruction = tertiary
+        quaternaryInstruction = nil
```
1ec5: This instruction should also be passed in as a parameter.

Sources/MapboxDirections/VisualInstructionBanner.swift
```diff
@@ -45,6 +47,7 @@ open class VisualInstructionBanner: Codable {
         primaryInstruction = try container.decode(VisualInstruction.self, forKey: .primaryInstruction)
         secondaryInstruction = try container.decodeIfPresent(VisualInstruction.self, forKey: .secondaryInstruction)
         tertiaryInstruction = try container.decodeIfPresent(VisualInstruction.self, forKey: .tertiaryInstruction)
+        quaternaryInstruction = try container.decodeIfPresent(VisualInstruction.self, forKey: .quaternaryInstruction)
```
1ec5: This line should be balanced out by something in `encode(to:)`.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -147,6 +149,22 @@ public extension VisualInstruction.Component {
...
+         Initializes an image representation bearing the image at the given base URL.
+         */
+        public init(imageBaseURL: URL?) {
+            self.imageBaseURL = imageBaseURL
+        }
+
+        /**
+         Returns a remote URL to the image file that represents the component.
+         */
+        public let imageBaseURL: URL?
```
1ec5: Unlike the property in ImageRepresentation, this property represents the full URL (minus any authentication-related query parameters), so “base” should be removed from the name.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -147,6 +149,22 @@ public extension VisualInstruction.Component {
             return scale
         }
     }
+
+    struct GuidanceViewImageRepresentation: Equatable {
```
1ec5: ImageRepresentation isn’t meant to be limited to the `image` case; it just means that it contains the details for rendering an image. So does this struct, but the difference is in how the URL is interpreted. A separate struct is fine, but we should probably rename the two structs to better reflect the distinction. Perhaps one would be ScalableImageRepresentation or something to that effect. Needs more thought.

1ec5:
Incorporated into #425.

Issue #107, opened 2017-01-26
Support for multiple waypoints, and the Trip Optimizer API


danpat:
Our Directions API supports multiple waypoints, in the form:

/directions/v5/mapbox.whatever/1,1;2,2;3,3;4;4

and it'll return a route that passes through all those coordinates, in that order.  Each pair are returned as a RouteLeg object.

As a bonus, the Optimized Trip API (landing soon!) has exactly the same request/response format.  Same URL, same JSON response structure.  The only difference is that the route returned will have the waypoints re-arranged to be in the shortest-path order (https://github.com/mapbox/api-directions/pull/899) and it'll give you a round-trip.

There is stuff in the works to allow for optimized trips between specific source/destination coordinates with multiple waypoints in the middle.  It's expected that this will require adding `source=N&destination=Y` to the request URL, and the JSON response body will be identical.

This ticket is about supporting multiple waypoints, and exposing an option to "optimize the route" by switching to the Optimized Trip API instead of the normal Directions API.

1ec5:
> supporting multiple waypoints

MapboxDirections.swift does support multiple waypoints, as of #16.

Given the similarities between the Optimized Trip API and Directions API v5, we could simply overload RouteOptions with an additional `sortsWaypoints` flag (defaulting to false) or `arrivesAtWaypointsInOrder` (defaulting to true). Alternatively, if this API is expected to diverge from the Directions API, we could set up subclasses along the lines of the existing RouteOptionsV4.

/cc @bsudekum

danpat:
We don't expect any diverging any time soon.

1ec5:
This is about the [Optimized Trips API](https://www.mapbox.com/api-documentation/#optimized-trips), which is currently available as a preview. There’s quite a lot of overlap between that API and Directions API v5, which is why @danpat and I think MapboxDirections.swift would be a good home for it. We may be able to take advantage of the generic interface that allows the v4 and v5 API support to coexist.

samgehret:
This would be nice for https://github.com/mapbox/sales-engineering/issues/820

Ryde's use case is to input multiple waypoints/stops, create an optimal route to stop at each way point, and feed this into turn by turn directions.

abouroubi:
Hi !
Is there any ETA on this ? Been hanging for more than 3 years

zugaldia:
[...]

PR #406, opened 2020-01-17
Public Response Types

JThramer:
Implements #391. 

### To-Do: 
_Nothing!_ 🎉 

### Done: 
- [x] Documentation
- [x] Add error case that handles a no-connection situation.
- [x] PR Comments
- [x]  Fix Tests
- [x] `DirectionsError` needs to be `Codable`
- [x] `DirectionsCredentials` on `Directions` class that contains `accessToken`, `host`, and `skuToken`, copied to response on calculate
Convert a match response into a route response:
- [x] `Match`es need to be convertible to `Route`s
- [x] `MapMatchingResponse` needs to be convertible to `RouteResponse`
- [x] ~`MatchOptions` needs to be convertible to `RouteOptions`~

cc @mapbox/navigation-ios


Sources/MapboxDirections/Directions.swift
```diff
@@ -77,15 +77,15 @@ open class Directions: NSObject {
      If the request was canceled or there was an error obtaining the routes, this argument is `nil`. This is not to be confused with the situation in which no results were found, in which case the array is present but empty.
      - parameter error: The error that occurred, or `nil` if the placemarks were obtained successfully.
```
1ec5: Remember to update the completion handler types’ documentation. (×2)

Sources/MapboxDirections/Directions.swift
```diff
@@ -158,48 +158,55 @@ open class Directions: NSObject {
         let requestTask = URLSession.shared.dataTask(with: request) { (possibleData, possibleResponse, possibleError) in
             let responseEndDate = Date()
             guard let response = possibleResponse, ["application/json", "text/html"].contains(response.mimeType) else {
-                completionHandler(nil, nil, .invalidResponse)
+                let response = RouteResponse(code: nil, message: nil, error: .invalidResponse(possibleResponse), uuid: nil, routes: nil, waypoints: nil)
```
1ec5: Let’s provide default argument values or an alternative initializer for the error case to reduce repetition.

Sources/MapboxDirections/Directions.swift
```diff
@@ -158,48 +158,55 @@ open class Directions: NSObject {
...
+                        result.error = .unableToRoute
                         DispatchQueue.main.async {
-                            completionHandler(result.waypoints, nil, .unableToRoute)
+                            completionHandler(result)
                         }
                         return
                     }
                     
-                    self.postprocess(routes, fetchStartDate: fetchStart, responseEndDate: responseEndDate, uuid: result.uuid)
+                    result.postprocess(accessToken: self.accessToken, apiEndpoint: self.apiEndpoint, fetchStartDate: fetchStart, responseEndDate: responseEndDate)
```
1ec5: RouteResponse should read in most of these properties via decoder options. Even `responseEndDate` can probably be set automatically at the end of `RouteResponse(from:)` when there’s a `responseStartDate`.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -57,4 +65,32 @@ extension RouteResponse: Codable {
...
+    
+    /**
+     Adds request- or response-specific information to each result in a response.
+     */
+    func postprocess(accessToken: String, apiEndpoint: URL, fetchStartDate: Date, responseEndDate: Date) {
+        guard let routes = self.routes else {
+            return
+        }
+        
+        for result in routes {
```
1ec5: I neglected to call it out specifically in #391, but the major reason for this refactoring is so that properties like `accessToken`, `routeIdentifier` (a misnomer), and `fetchStartDate` can live on the response object instead of each individual result object.

Moving these properties to the response object will likely have some effect on the navigation SDK, but I think it’s healthy that the SDK will need to keep track of a whole response and perhaps a route index within that response, not just an individual result.

Sources/MapboxDirections/Directions.swift
```diff
@@ -441,29 +465,21 @@ open class Directions: NSObject {
...
 
 public extension CodingUserInfoKey {
     static let options = CodingUserInfoKey(rawValue: "com.mapbox.directions.coding.routeOptions")!
+    static let routesFromMatch = CodingUserInfoKey(rawValue: "com.mapbox.directions.coding.routesFromMatch")!
+    static let tracepoints = CodingUserInfoKey(rawValue: "com.mapbox.directions.coding.tracepoints")!
+
+
 }
 
 public class DirectionsDecoder: JSONDecoder {
```
1ec5: Why do we need to subclass JSONDecoder? Can we use JSONDecoder as is?

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -17,18 +17,26 @@ extension RouteResponse: Codable {
...
+        
+        // Matching-service specific keys. Used in decode only.
+        case matches = "matchings"
+        case tracepoints
     }
     
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
+        // Is this an edge-case, where we are creating a route response from the map matching service?
+        let fromMatchingService = (decoder as? DirectionsDecoder)?.decodingRouteResponseFromMatchService ?? false
```
1ec5: The former MapMatchingResponse and RouteResponse functionality should not be unified into a single class. Doing so results in special cases within the implementation, which could turn into spaghetti code in the future as we add support for the Optimization API (#107) and Matrix API.

https://github.com/mapbox/mapbox-directions-swift/issues/391#issue-527602027 proposed:

> For the case that MapMatchingResponse is meant to handle, it would be better for MapMatchingResponse to produce Match objects; `calculateRoutes(matching:completionHandler:)` could call a new initializer on RouteResponse that takes a MapMatchingResponse as input.

What would be the downside of first interpreting a Map Matching API response as a series of Matches in a MapMatchingResponse, then converting the Matches to Routes in a RouteResponse? Could this approach lead to better handling of submatches than #386?

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -1,49 +1,46 @@
 import Foundation
 
 public struct MapMatchingResponse {
-    public var code: String
-    public var routes : [Route]?
-    public var waypoints: [Waypoint]
+    public var code: String?
+    public var message: String?
```
1ec5: `code` and `message` can both be encapsulated in `error`, so we should keep `code` and `message` around as parallel state, especially since all these properties are writable.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 public struct RouteResponse {
     public var code: String?
     public var message: String?
-    public var error: String?
+    public var error: DirectionsError?
     public let uuid: String?
```
1ec5: This ID isn’t guaranteed to be a valid UUID. (The Swift standard library’s UUID type expects UUIDs to conform to RFC 4122 version 4.) So we should call it `identifier` instead.

Sources/MapboxDirections/Directions.swift
```diff
@@ -152,54 +113,71 @@ open class Directions: NSObject {
...
                 return
             }
             
             if let error = possibleError {
-                completionHandler(nil, nil, .unknown(response: possibleResponse, underlying: error, code: nil, message: nil))
+                let response = RouteResponse(httpResponse: httpResponse, options: .route(options), credentials: self.credentials)
+                let unknownError = DirectionsError.unknown(response: possibleResponse, underlying: error, code: nil, message: nil)
+                completionHandler(response, unknownError)
                 return
             }
```
SebastianOsinski: Current error handling and this one has big problem: when there is no internet, or request was cancelled, it always returns `invalidResponse` error instead of unknown or other one. This is caused by not checking `possibleError` first. It should be inspected as first one (as it contains error when request is canceled or failed for other reasons - see Foundation.URLError for reference). Only after assuring it is nil, we should inspect `possibleResponse` and `possibleData`.

JThramer: Great suggestion, @SebastianOsinski! Will resolve.

Sources/MapboxDirections/Directions.swift
```diff
@@ -153,59 +116,86 @@ open class Directions: NSObject {
...
      */
     @discardableResult open func calculate(_ options: RouteOptions, completionHandler: @escaping RouteCompletionHandler) -> URLSessionDataTask {
-        let fetchStart = Date()
+        options.fetchStartDate = Date()
         let request = urlRequest(forCalculating: options)
         let requestTask = URLSession.shared.dataTask(with: request) { (possibleData, possibleResponse, possibleError) in
-            let responseEndDate = Date()
-            guard let response = possibleResponse, ["application/json", "text/html"].contains(response.mimeType) else {
+            
+            let offlineErrors: [URLError.Code] = [.cannotConnectToHost, .dataNotAllowed, .notConnectedToInternet]
```
SebastianOsinski: I don't think we should select only offline errors from all errors. IMO if `possibleError` is not nil, then we should call completion handler with some `.network(error)` and call it a day. 

Current solution still suffers from handling some URLSession errors as invalid response errors. E.g. if we cancel `URLSessionDataTask` returned by this method, then `possibleError` will be URLError with code `canceled`. Yet we'll call completion with `invalidResponse` error. This prevents users of this SDK from correctly handling request cancellation.

I propose putting `if let error = possibleError` at the beginning of the method, without any additional checks about the nature of the error, and just calling completion with error wrapped in our own `network` error case. That way we'll enable users of the SDK to correctly react to various URLSession errors.

JThramer: 🤔 Agreed. Will change.

Sources/MapboxDirections/Directions.swift
```diff
@@ -153,59 +116,85 @@ open class Directions: NSObject {
...
             
             if let error = possibleError {
+                let response = RouteResponse(httpResponse: httpResponse, options: .route(options), credentials: self.credentials)
+                let unknownError = DirectionsError.unknown(response: possibleResponse, underlying: error, code: nil, message: nil)
                 DispatchQueue.main.async {
-                    completionHandler(nil, nil, .unknown(response: possibleResponse, underlying: error, code: nil, message: nil))
+                    completionHandler(response, unknownError)
                 }
                 return
             }
```
SebastianOsinski: This check is probably not needed anymore. I think URLSession can only return URLError instances, so first check at the top is enough. Also this is dead code path. According to docs, `possibleError` is not nil only when `possibleData` is nil (and vice versa), so previous guard will enter its `else` clause before this if will be evaluated.

JThramer: 👋 Hey @SebastianOsinski 

Great catch! yeah, that's definitely dead code. Will remove.

Sources/MapboxDirections/DirectionsError.swift
```diff
@@ -4,17 +4,54 @@ import Foundation
...
+            }
+        } else {
+            self = .unknown(response: response, underlying: error, code: code, message: message)            
+        }
+    }
+
+    /**
+     There is no network connection available to perform the network request.
+     */
+    case network(_: URLError?)
```
SebastianOsinski: We can probably drop `?` from here.

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -1,49 +1,55 @@
...
-    public var code: String
-    public var routes : [Route]?
-    public var waypoints: [Waypoint]
+    public let httpResponse: HTTPURLResponse?
+    
+    public var matches : [Match]?
+    public var tracepoints: [Tracepoint?]?
+    
+    public let options: MatchOptions
+    public let credentials: DirectionsCredentials
```
Udumft: I am not sure to understand why we need `credentials` in both `MapMatchingResponse` and `RouteResponse`.
Is it really important to know in a response which host or access token it used? Is it the case that there might be multiple hosts and/or tokens used within one app, so user needs to distinguish them?

Sources/MapboxDirections/MapMatching/Match.swift
```diff
@@ -3,9 +3,38 @@ import CoreLocation
 import Polyline
 import struct Turf.LineString
 
-extension CodingUserInfoKey {
-    static let tracepoints = CodingUserInfoKey(rawValue: "com.mapbox.directions.coding.tracepoints")!
+public enum Weight: Equatable {
```
Udumft: Where can I find some docs or something to learn more about what "Weight" is?

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
JThramer: Yes, generally the credentials are important for purposes of re-routing, which can happen anytime during a navigation session.

Sources/MapboxDirections/MapMatching/Match.swift
JThramer: Good catch, I wrote a documentation comment for the enum.

Udumft:
At this point these changes heavily conflict with #388 . I would be important to have #406 merged before continuing #388.

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -1,49 +1,55 @@
 import Foundation
 
 public struct MapMatchingResponse {
-    public var code: String
-    public var routes : [Route]?
-    public var waypoints: [Waypoint]
```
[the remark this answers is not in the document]
JThramer: Because a `MapMatchingRespone` should have `Matches` and `Tracepoints`, which are the proper models documented in the service API.

Sources/MapboxDirections/MapMatching/MatchResponse.swift
```diff
@@ -1,31 +0,0 @@
...
-        case matches = "matchings"
-        case tracepoints
-    }
-    
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        code = try container.decode(String.self, forKey: .code)
-        message = try container.decodeIfPresent(String.self, forKey: .message)
-        tracepoints = try container.decodeIfPresent([Tracepoint?].self, forKey: .tracepoints)
-        matches = try container.decodeIfPresent([Match].self, forKey: .matches)
```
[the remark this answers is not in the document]
JThramer: It was moved over to `MapMatchingResponse`. 

https://github.com/mapbox/mapbox-directions-swift/pull/406/files/a6607fbcd788a0575b752d3a0e9bee74e9d8ddcd#diff-c28392baf0208871bf1ac5c51e283fe1R51-R52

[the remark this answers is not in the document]
Udumft: I see, but what about this assignment? https://github.com/mapbox/mapbox-directions-swift/pull/406/files/a6607fbcd788a0575b752d3a0e9bee74e9d8ddcd#diff-74f5c94109c402f808fec0e79d7c655cL25-L29

PR #417, opened 2020-02-25
Fix decoding TransportType for 'pushing bike' case

### LINKED ISSUE #413: RouteStep parsing fails when directions for cycling require user to dismount bike
**Version: 1.0.0-alpha.1**

Response for `calculate(_: completionHandler:)` sometimes fails to decode, when `RouteOptions`  is `NavigationRouteOptions` with `profileIdentifier` set to `.cycling`. If directions require user to dismount bike, then `mode` field in one of the `RouteStep` parts of response is set to `pushing bike`:

```json
 {
                "intersections": [
                     ...
                ],
                "driving_side": "right",
                "geometry": "sonn`Bafmn_@nLoaA|@{G",
                "mode": "pushing bike", // <-- this
                "maneuver": {
                  "bearing_after": 106,
                  "bearing_before": 57,
                  "location": [
                    17.030257,
                    51.109642
                  ],
                  "modifier": "right",
                  "type": "continue",
                  "instruction": "Turn right to stay on Rynek"
                },
                "weight": 87.4,
                "duration": 87.4,
                "name": "Rynek",
                "distance": 88.6,
                "bannerInstructions": [
                     ...
                ]
              },
```

`RouteStep` does not support this type of mode. `mode` is represented by `TransportType` enum which does not have case for that.

Documentation comment: https://github.com/mapbox/mapbox-directions-swift/blob/cb704e694d7e199871e05f0e7bbff4195ebd1dd7/Sources/MapboxDirections/RouteStep.swift#L45 suggests that `walking` should be returned as mode instead of `pushing bike`

Documentation for Directions API doesn't mention this type of mode: https://docs.mapbox.com/api/navigation/#route-step-object

Screen Shot 2020-02-14 at 10 06 48

SebastianOsinski:
Fixes #413

### DIFF
--- a/Tests/MapboxDirectionsTests/TransportTypeTests.swift
+++ b/Tests/MapboxDirectionsTests/TransportTypeTests.swift
@@ -1,0 +1,59 @@
+//
+//  TransportTypeTests.swift
+//  MapboxDirections
+//
+//  Created by Sebastian Osiński on 25/02/2020.
+//  Copyright © 2020 Mapbox. All rights reserved.
+//
+
+import XCTest
+@testable import MapboxDirections
+
+class TransportTypeTests: XCTestCase {
+    func testDecoding() {
+        let examples: [Example<String, TransportType>] = [
+            Example("driving", .automobile),
+            Example("ferry", .ferry),
+            Example("movable bridge", .movableBridge),
+            Example("unaccessible", .inaccessible),
+            Example("walking", .walking),
+            Example("cycling", .cycling),
+            Example("train", .train)
+        ]
+
+        let decoder = JSONDecoder()
+
+        for example in examples {
+            let decoded = try? decoder.decode(
+                CodableContainer<TransportType>.self,
+                from: jsonData(type: example.input)
+            )
+
+            XCTAssertEqual(decoded?.wrapped, example.expected)
+        }
+    }
+
+    private func jsonData(type: String) -> Data {
+        """
+        {
+            "wrapped": "\(type)"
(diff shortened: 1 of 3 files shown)


Tests/MapboxDirectionsTests/TransportTypeTests.swift
```diff
@@ -0,0 +1,52 @@
...
+                CodableContainer<TransportType>.self,
+                from: jsonData(type: example.input)
+            )
+
+            XCTAssertEqual(decoded?.wrapped, example.expected)
+        }
+    }
+
+    private func jsonData(type: String) -> Data {
+        """
```
1ec5: This appears to be the cause of a build failure on CI:

```
/Users/distiller/project/Tests/MapboxDirectionsTests/TransportTypeTests.swift:34:31: warning: expression of type 'Data' is unused
        """.data(using: .utf8)!
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
/Users/distiller/project/Tests/MapboxDirectionsTests/TransportTypeTests.swift:35:5: error: missing return in a function expected to return 'Data'
    }
    ^
```

```suggestion
        return """
```

SebastianOsinski: I already got used to omitting return statements in Swift 5.1 😅

1ec5: We’ll get there before long! 😄

Issue #414, opened 2020-02-14
Suggestion: Consider using Result instead of optional response and error


SebastianOsinski:
This is merely a suggestion. We could use Result for completion handlers instead of all optional arguments:

```swift
public typealias RouteCompletionHandler = Result<([Waypoint], [Route]), DirectionsError> -> Void

public typealias MatchCompletionHandler = Result<[Match], DirectionsError> -> Void
```

This still can be done before 1.0.0 release. The only thing I am not sure about is `[Waypoint]` in `RouteCompletionHandler`. Is it always returned when routes are returned? Or is it possible to get waypoints but not routes or vice-versa?

1ec5:
Thanks for the suggestion! This overlaps with some of the proposal in #391 to encapsulate Route/Match in a response object. That refactoring is underway in #406.

We’ve been going back and forth about whether the options object and error should be members of the response object or live alongside it in the completion handler. The optionality is unfortunate, given that a nontrivial response is mutually exclusive with an error. That might be a case for using an enumeration with associated values. 🤔

A generic response type would be an interesting idea. So far we’ve been thinking about a response protocol, which would enforce the presence of a few members we think would be common to any navigation-related Mapbox API. But a generic type might be a cleaner way to go about that.

Please have a look at #406. We’d welcome suggestions, though we also expect these types to require additional rounds of refactoring, such as for waypoints in #388.

> The only thing I am not sure about is `[Waypoint]` in `RouteCompletionHandler`. Is it always returned when routes are returned? Or is it possible to get waypoints but not routes or vice-versa?

As things stand, the response objects would fall back to the waypoints in the options object:

https://github.com/mapbox/mapbox-directions-swift/blob/fe77075932d469473237bb5346dfcc5909124886/Sources/MapboxDirections/RouteResponse.swift#L42-L44

However, certain errors will pass no waypoints to the completion handler, while others will pass in the response object’s waypoints:

https://github.com/mapbox/mapbox-directions-swift/blob/fe77075932d469473237bb5346dfcc5909124886/Sources/MapboxDirections/Directions.swift#L187-L189 https://github.com/mapbox/mapbox-directions-swift/blob/fe77075932d469473237bb5346dfcc5909124886/Sources/MapboxDirections/Directions.swift#L194-L196

1ec5:
Oh whoops, I didn’t realize you were referring to the Result type in the standard library – that’s a good idea.

JThramer:
[...]

PR #418, opened 2020-02-26
Use Result for directions response

JThramer:
Using `Foundation.Result` as the base type returned in `Directions.calculate`. Implements #414. Thanks @SebastianOsinski!

To-Do: 
- [x] Finish fixing tests
- [x] Update docs

/cc @mapbox/navigation-ios

### DIFF
--- a/Sources/MapboxDirections/Directions.swift
+++ b/Sources/MapboxDirections/Directions.swift
@@ -60,7 +60,7 @@ open class Directions: NSObject {
     
     public typealias MapMatchingError = (response: MapMatchingResponse, error: DirectionsError)
     
-    public typealias DirectionsSession = (options: DirectionsOptions, credentials: DirectionsCredentials)
+    public typealias Session = (options: DirectionsOptions, credentials: DirectionsCredentials)
     
     /**
      A closure (block) to be called when a directions request is complete.
@@ -69,7 +69,7 @@ open class Directions: NSObject {
      
      - parameter error: The error that occurred, or `nil` if the solution was obtained successfully.
      */
-    public typealias RouteCompletionHandler = (_ session: DirectionsSession, _ result: Result<RouteResponse, DirectionsError>) -> Void
+    public typealias RouteCompletionHandler = (_ session: Session, _ result: Result<RouteResponse, DirectionsError>) -> Void
     
     /**
      A closure (block) to be called when a map matching request is complete.
@@ -78,7 +78,7 @@ open class Directions: NSObject {
 
      - parameter error: The error that occurred, or `nil` if the solution was obtained successfully.
      */
-    public typealias MatchCompletionHandler = (_ session: DirectionsSession, _ result: Result<MapMatchingResponse, DirectionsError>) -> Void
+    public typealias MatchCompletionHandler = (_ session: Session, _ result: Result<MapMatchingResponse, DirectionsError>) -> Void
     
     // MARK: Creating a Directions Object
     
--- a/Sources/MapboxDirections/Directions.swift
+++ b/Sources/MapboxDirections/Directions.swift
@@ -60,9 +60,6 @@ open class Directions: NSObject {
     
     public typealias MapMatchingError = (response: MapMatchingResponse, error: DirectionsError)
     
-//    public typealias RouteError: Error = (result: RouteResponse, error: DirectionsError)
-//    public typealias MapMatchingError: Error = (result: MapMatchingResponse, error: DirectionsError)
-    
     public typealias DirectionsSession = (options: DirectionsOptions, credentials: DirectionsCredentials)
     
     /**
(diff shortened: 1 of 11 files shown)


Sources/MapboxDirections/Directions.swift
```diff
@@ -57,14 +57,22 @@ let userAgent: String = {
  Each result produced by the directions object is stored in a `Route` object. Depending on the `RouteOptions` object you provide, each route may include detailed information suitable for turn-by-turn directions, or it may include only high-level information such as the distance, estimated travel time, and name of each leg of the trip. The waypoints that form the request may be conflated with nearby locations, as appropriate; the resulting waypoints are provided to the closure.
  */
 open class Directions: NSObject {
+    
+    public typealias MapMatchingError = (response: MapMatchingResponse, error: DirectionsError)
+    
+//    public typealias RouteError: Error = (result: RouteResponse, error: DirectionsError)
+//    public typealias MapMatchingError: Error = (result: MapMatchingResponse, error: DirectionsError)
```
JThramer: remove comment

Sources/MapboxDirections/Directions.swift
```diff
@@ -57,14 +57,19 @@ let userAgent: String = {
...
+    public typealias MapMatchingError = (response: MapMatchingResponse, error: DirectionsError)
+    
+    public typealias DirectionsSession = (options: DirectionsOptions, credentials: DirectionsCredentials)
+    
     /**
      A closure (block) to be called when a directions request is complete.
      
      - parameter response: A `RouteResponse` object that contains the entire payload of the Directions API solution. See `RouteResponse.swift` for more information.
      
      - parameter error: The error that occurred, or `nil` if the solution was obtained successfully.
```
Udumft: Please do not forget to update docs in result

Sources/MapboxDirections/Directions.swift
```diff
@@ -152,40 +155,36 @@ open class Directions: NSObject {
...
+                            completionHandler(session, .failure(apiError))
                         }
                         return
                     }
                     
                     let result = try decoder.decode(RouteResponse.self, from: data)
                     guard result.routes != nil else {
                         DispatchQueue.main.async {
-                            completionHandler(result, .unableToRoute)
+                            completionHandler(session, .failure(.unableToRoute))
```
Udumft: I am not sure about original reasons, but as @1ec5 mentioned in [original proposal](https://github.com/mapbox/mapbox-directions-swift/issues/414#issuecomment-586404352),  this was the case which returned an error and a waypoints at the same time. I don't know if it was done for a good reason or not, but apparently such behavior will be lost.

Sources/MapboxDirections/Directions.swift
```diff
@@ -57,14 +57,19 @@ let userAgent: String = {
  Each result produced by the directions object is stored in a `Route` object. Depending on the `RouteOptions` object you provide, each route may include detailed information suitable for turn-by-turn directions, or it may include only high-level information such as the distance, estimated travel time, and name of each leg of the trip. The waypoints that form the request may be conflated with nearby locations, as appropriate; the resulting waypoints are provided to the closure.
  */
 open class Directions: NSObject {
+    
+    public typealias MapMatchingError = (response: MapMatchingResponse, error: DirectionsError)
```
Udumft: Unused synonym. Did you intend to use it as a result type at some point?

JThramer:
Blocked from merge by https://github.com/mapbox/mapbox-directions-swift/pull/418, need to fix cartfile

Sources/MapboxDirections/Directions.swift
JThramer: as it turns out, `session` contains an options object with the waypoints.

Tests/MapboxDirectionsTests/DirectionsTests.swift
```diff
@@ -113,16 +117,19 @@ class DirectionsTests: XCTestCase {
         
         let directions = Directions(credentials: BogusCredentials)
         let opts = RouteOptions(locations: [one, two])
-        directions.calculate(opts, completionHandler: { (response, error) in
+        directions.calculate(opts, completionHandler: { (session, result) in
             expectation.fulfill()
```
Udumft, line 121: fulfilling expectation before other checks seems incorrect...

Tests/MapboxDirectionsTests/DirectionsTests.swift
```diff
@@ -157,20 +164,20 @@ class DirectionsTests: XCTestCase {
...
```
Udumft, line 168: Same fullfilment not as above

Tests/MapboxDirectionsTests/DirectionsTests.swift
JThramer: Why do you think that? Keep in mind that the fulfillment is not there to say "okay test is over", It's merely there to ensure that the (faked) directions request takes no longer than two seconds to call it's `completionHandler.`

JThramer: I was all ready to explain why you're wrong, but as it turns out, you're not. 

Very good catch. 🙇‍♂️

PR #420, opened 2020-04-24
Route Refresh

Udumft:
mapbox/mapbox-navigation-ios#2284

Work-In-Progress.
Added Public method to call a Route Refresh.
Refactored Route Coding a bit to re-use some parts in Refresh Response.
Looking for some better way for updating Route contents...


Sources/MapboxDirections/RouteRefreshResponse.swift
```diff
@@ -0,0 +1,83 @@
...
+            if legEnum.offset >= legIndex {
+                leg.updateAnnotationData(from: annotation)
+            }
+            return leg
+        }
+        
+        let route = Route(legs: updatedLegs,
+                           shape: route.shape,
+                           distance: route.distance,
+                           expectedTravelTime: route.expectedTravelTime)
```
1ec5, line 45: Instead of creating a brand-new Route object based on the old one, can we mutate the existing route object? I’m just concerned that some SDK or application code may have stored a reference to the old Route that’s now outdated. It would be fine to make additional properties read-write in order to reuse the existing data structures. If we’re concerned about changing data out from behind the client’s back, then I suppose we could create a parallel MutableRoute class instead, but that could be a lot to maintain. In the future, once we find a way to convert these types to structs, we’ll be able to make mutability more explicit anyways.

Udumft: Making `Route` mutable should still be done explicitly outside `Directions.refresh` because (1) users should be aware of the update and have control over it , (2) refreshing route also involves updating `RouteProgress` (and potentially anything else) and doing some calculations, so it is safer be done by user's completion.

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -50,6 +50,8 @@ extension MapMatchingResponse: Codable {
         
         tracepoints = try container.decodeIfPresent([Tracepoint?].self, forKey: .tracepoints)
         matches = try container.decodeIfPresent([Match].self, forKey: .matches)
-
+        matches?.enumerated().forEach {
+            $0.element.routeIndex = $0.offset
+        }
```
1ec5, line 55: This is fine, but generally it’s clearer to use a for loop unless we’re doing something particularly functional with the array:

```suggestion
        for (index, match) in matches?.enumerated() ?? [] {
            match.routeIndex = index
        }
```

Sources/MapboxDirections/DirectionsError.swift
```diff
@@ -102,6 +102,12 @@ public enum DirectionsError: LocalizedError {
      */
     case rateLimited(rateLimitInterval: TimeInterval?, rateLimit: UInt?, resetTime: Date?)
     
+    /**
+     Could not refresh the route with given parameters.
+     
+     Try validating Route UID, Route Index and Leg Index.
```
1ec5, line 108:
```suggestion
     Make sure `DirectionsResult.routeIdentifier`, `CodingUserInfoKey.routeIndex`, and `CodingUserInfoKey.legIndex` are set to the right values.
```

Sources/MapboxDirections/DirectionsError.swift
```diff
@@ -169,6 +177,8 @@ public enum DirectionsError: LocalizedError {
             }
             let formattedDate: String = DateFormatter.localizedString(from: rolloverTime, dateStyle: .long, timeStyle: .long)
             return "Wait until \(formattedDate) before retrying."
+        case .unableToRefresh:
+            return "Try checking Route UID, Index and Leg index passed for refreshing."
```
1ec5, line 181:
```suggestion
            return "Make sure DirectionsResult.routeIdentifier, CodingUserInfoKey.routeIndex, and CodingUserInfoKey.legIndex are set to the right values when refreshing."
```

Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -157,6 +157,13 @@ open class DirectionsResult: Codable {
      */
     open var routeIdentifier: String?
     
+    /**
+     Route index position as was provided by corresponding directions response.
+     
+     `Directions.calculate` methods may produce multiple alternative routes. `routeIndex` displays order of appearance in the response. This parameter might be useful for route refreshing.
+     */
+    open var routeIndex: Int?
```
1ec5, line 165: We’ve long tried to avoid introducing explicit index properties, instead relying on the containing structure (in this case an array in RouteResponse) to enforce order. If we need to introduce this property, I think we should make it constant to avoid confusion and possible scenarios where it gets out of sync with the actual order in the list of results.

Sources/MapboxDirections/Directions.swift
```diff
@@ -375,6 +384,101 @@ open class Directions: NSObject {
         return requestTask
     }
     
+    @discardableResult open func refresh(route: Route, currentLegIndex: Int, completionHandler: @escaping RouteRefreshCompletionHandler) -> URLSessionDataTask? {
+        
+        guard let routeIdentifier = route.routeIdentifier, let routeIndex = route.routeIndex else {
+            completionHandler(credentials, .failure(.invalidInput(message: "Route must have an Identifier and Index.")))
```
1ec5, line 390:
```suggestion
            completionHandler(credentials, .failure(.invalidInput(message: "Route must have a routeIdentifier and routeIndex.")))
```

Sources/MapboxDirections/Directions.swift
```diff
@@ -375,6 +384,101 @@ open class Directions: NSObject {
         return requestTask
     }
     
+    @discardableResult open func refresh(route: Route, currentLegIndex: Int, completionHandler: @escaping RouteRefreshCompletionHandler) -> URLSessionDataTask? {
+        
+        guard let routeIdentifier = route.routeIdentifier, let routeIndex = route.routeIndex else {
```
1ec5, line 389: Instead of persisting the original index as part of the route, we can have this method take the whole RouteResponse and a route index. The original goal of #391 was that the navigation SDK would keep track of the entire RouteResponse and a route index, or the route and route index. That would require some straightforward but fairly invasive refactoring of the public API. I think we need to make that change ahead of v1.1 anyhow, because this is only the first of several features that will require context about more than the individual route that the user selected.

1ec5:
For context, the Route Refresh API only refreshes a single leg at a time, and it doesn’t accept the route object itself as input, so instead it requires us to pass in the “route” UUID (which is actually the same across all the routes in a single response), a route index (to figure out which of those routes we mean), and a leg index. The Directions API doesn’t return route indices; that was something the PR originally calculated based on each route’s index in the array.

The motivation behind https://github.com/mapbox/mapbox-directions-swift/pull/420#discussion_r480203743 is that we’re trying to move away from stashing ad-hoc stuff inside Route and Match objects, in part because MapboxNavigationNative will soon handle postprocessing the route response for MapboxCoreNavigation and may lose or barf on any nonstandard properties we add.

Sources/MapboxDirections/RouteRefreshResponse.swift
```diff
@@ -0,0 +1,82 @@
...
+        self.credentials = credentials
+        
+        self.identifier = try container.decodeIfPresent(String.self, forKey: .identifier)
+        
+        let refreshedRoute = try container.decode(RefreshedRoute.self, forKey: .route)
+        self.legAnnotations = refreshedRoute.legAnnotations
+        
+        self.routeIndex = decoder.userInfo[.routeIndex] as! Int
+        self.legIndex = decoder.userInfo[.legIndex] as! Int
+        let refreshingRoute = decoder.userInfo[.refreshingRoute] as! Route
```
1ec5, line 69: Instead of force-casting these dictionary values, throw a `DirectionsCodingError.missingOptions` if they aren’t present and of the expected type.

/ref #453

Sources/MapboxDirections/RouteRefreshResponse.swift
```diff
@@ -0,0 +1,82 @@
...
+                leg.updateAnnotationData(from: annotation)
+            }
+            return leg
+        }
+        
+        let route = Route(legs: updatedLegs,
+                           shape: route.shape,
+                           distance: route.distance,
+                           expectedTravelTime: route.expectedTravelTime)
+        route.routeIdentifier = route.routeIdentifier
```
1ec5, line 46: I think the intention was to copy the original route’s UUID into the new route, but this statement ends up being a no-op because of the shadowed variable.

Sources/MapboxDirections/Directions.swift
```diff
@@ -375,6 +384,103 @@ open class Directions: NSObject {
...
+        requestTask.priority = 1
+        requestTask.resume()
+        return requestTask
+    }
+    
+    open func urlRequest(forRefreshing routeId: String, routeIndex: Int, currentLegIndex: Int) -> URLRequest {
+        var params: [URLQueryItem] = []
+        params += [URLQueryItem(name: "access_token", value: credentials.accessToken)]
+        
+        var unparameterizedURL = URL(string: "directions-refresh/v1/\(DirectionsProfileIdentifier.automobileAvoidingTraffic.rawValue)", relativeTo: credentials.host)!
```
1ec5, line 472: The `automobileAvoidingTraffic` profile identifier is hard-coded here because the directions refresh API endpoint only supports refreshing routes calculated using this profile. (The other profiles have no traffic congestion data to refresh.) There is code in RouteOptions to avoid enabling route refreshing for the other profiles, but should we also check here that the `RouteLeg.profileIdentifier` is `.automobileAvoidingTraffic`, in case the developer enabled route refreshing on a route of the route profile?

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -98,18 +82,24 @@ open class RouteLeg: Codable {
...
+            try container.encode(annotation, forKey: .annotation)
         }
     }
 
+    func updateAnnotationData(from annotation: RouteLegAnnotation?) {
+        segmentDistances = annotation?.segmentDistances
+        expectedSegmentTravelTimes = annotation?.expectedSegmentTravelTimes
+        segmentSpeeds = annotation?.segmentSpeeds
+        segmentCongestionLevels = annotation?.segmentCongestionLevels
+        segmentMaximumSpeedLimits = annotation?.segmentMaximumSpeedLimits
```
1ec5, line 100: I think we can live with this repetitiveness for now, but long term I’m concerned about potentially adding more attribute options and forgetting to copy one of them over here. To avoid that problem, and also to clean up the RouteLeg interface a bit, we could make RouteLegAnnotation public and replace the various segment* properties on RouteLeg with a single `attributes` property that we’d simply replace when refreshing.

We can take care of this refactoring as tail work in a future release, preserving backwards compatibility by leaving the segment* properties as computed properties.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -283,6 +273,25 @@ extension RouteLeg: CustomQuickLookConvertible {
     }
 }
 
+extension RouteLeg: NSCopying {
+    public func copy(with zone: NSZone? = nil) -> Any {
```
1ec5, line 277: Since #382, we’ve avoided implementing NSCopying, because in the past, we’ve frequently forgotten to add to copy methods when adding new properties, resulting in silent data loss. The hope expressed in #381 was that we’d convert the classes to structs, which have built-in copying functionality. In the meantime, the navigation SDK worked around the lack of NSCopying conformance in RouteOptions by round-tripping through JSON:

https://github.com/mapbox/mapbox-navigation-ios/blob/5f0642fa1574450302baa7fa0f8683df39e5bfdd/MapboxCoreNavigation/RouteOptions.swift#L31-L35

But I can see how that approach would be wildly unperformant for copying an entire leg. Do we really need to copy each leg and create a new route when refreshing it? By making the segment* properties mutable, we’ve already set the expectation that they can be changed out from under client code. I think it would be simpler to just set those properties on the original object.

Sources/MapboxDirections/RouteLegAnnotation.swift
```diff
@@ -0,0 +1,49 @@
+
+import Foundation
+import CoreLocation
+
+struct RouteLegAnnotation {
```
1ec5, line 5: Note that this library avoids the term “annotation” in public API, since that term is so overloaded on Apple platforms. This struct isn’t public, but the suggestion above is to eventually make it public, in which case we’d rename it to something like RouteLegAttributes and probably nest it within RouteLeg.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -159,6 +163,11 @@ open class RouteOptions: DirectionsOptions {
      */
     open var includesExitRoundaboutManeuver = false
     
+    /**
+     A Boolean value indicating wether route can be refreshed during turn-by-turn navigation. This includes ETA and route congestion data.
```
1ec5, line 167:
```suggestion
     A Boolean value indicating whether `Directions` can refresh time-dependent properties of the `RouteLeg`s of the resulting `Route`s.
     
     To refresh the `RouteLeg.expectedSegmentTravelTimes`, `RouteLeg.segmentSpeeds`, and `RouteLeg.segmentCongestionLevels` properties, use the `Directions.refresh(routeResponse:routeIndex:currentLegIndex:completionHandler:)` method. This property is ignored unless `profileIdentifier` is `DirectionsProfileIdentifier.automobileAvoidingTraffic`.
```

Sources/MapboxDirections/RouteRefreshResponse.swift
```diff
@@ -0,0 +1,82 @@
+import Foundation
+
+public struct RouteRefreshResponse {
```
1ec5, line 3: I’m not sure it’s meaningful for RouteRefreshResponse to be public while its main data is locked away in a private `legAnnotations`. Until client code needs to use RouteRefreshResponse directly, let’s make this struct private.

Sources/MapboxDirections/RouteLeg.swift
1ec5: > Do we really need to copy each leg and create a new route when refreshing it?

Looks like we previously discussed this point in https://github.com/mapbox/mapbox-directions-swift/pull/420#discussion_r433385717.

Sources/MapboxDirections/RouteRefreshResponse.swift
1ec5: > users should be aware of the update and have control over it

Developers can observe changes to the Route and RouteLeg properties using key-value observation. The refreshable properties are already individually mutable, so there’s already an expectation that they could change behind the developer’s back. What the interface doesn’t explicitly call out is that refreshing the route silently makes any existing route references become stale; client code might not be in a good position to obtain the refreshed route in that case.

Sources/MapboxDirections/RouteRefreshResponse.swift
1ec5: Never mind, this struct is used in RouteRefreshCompletionHandler, which is public. Not sure how I missed that earlier.

Sources/MapboxDirections/Directions.swift
1ec5: To keep this refactor from ballooning in scope at the tail end of v1.0, we’re going to avoid pulling in the whole RouteResponse for now. In fact, I’ve refactored this method to take the response identifier itself, to make it easier to remove the `Route.routeIdentifier` property in #452.

Sources/MapboxDirections/RouteLeg.swift
1ec5: I moved this copying to a more obvious location in the setter of the `attributes` property, taking advantage of copy-by-value for structs.

Sources/MapboxDirections/RouteRefreshResponse.swift
1ec5: I ended up making the whole shebang public so that client code can take responsibility for merging.

chezzdev (approved):

Sources/MapboxDirections/Directions.swift
```diff
@@ -375,6 +385,103 @@ open class Directions: NSObject {
...
+            }
+            
+            guard let data = possibleData else {
+                DispatchQueue.main.async {
+                    completionHandler(self.credentials, .failure(.noData))
+                }
+                return
+            }
+            
+            DispatchQueue.global(qos: .userInitiated).async {
```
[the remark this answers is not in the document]
1ec5: This is consistent with `Directions.calculate(_:completionHandler:)`. If I’m not mistaken, the idea is to perform deserialization on a dedicated thread other than the CFNetwork thread.

1ec5 (approved):

Tests/MapboxDirectionsTests/RouteRefreshTests.swift
```diff
@@ -175,7 +175,7 @@ class RouteRefreshTests: XCTestCase {
     
     func testEncoding() {
         let leg = RefreshedRouteLeg(attributes:
-            .init(segmentDistances: [0], expectedSegmentTravelTimes: [0], segmentSpeeds: [0], segmentCongestionLevels: [.severe], segmentMaximumSpeedLimits: [Measurement(value: 1, unit: .milesPerHour)]))
+            .init(segmentDistances: [0], expectedSegmentTravelTimes: [0], segmentSpeeds: [0], segmentCongestionLevels: [CongestionLevel.severe], segmentMaximumSpeedLimits: [Measurement(value: 1, unit: .milesPerHour)]))
```
1ec5, line 178: A new test is failing to build in Xcode 10.2 due to what appears to be a compiler bug:

```
/Users/distiller/project/Tests/MapboxDirectionsTests/RouteRefreshTests.swift:178:14: error: type 'RouteLeg.Attributes' has no member 'severe'
            .init(segmentDistances: [0], expectedSegmentTravelTimes: [0], segmentSpeeds: [0], segmentCongestionLevels: [.severe], segmentMaximumSpeedLimits: [Measurement(value: 1, unit: .milesPerHour)]))
             ^~~~                                                                                                       ~~~~~~~
```

I’ve worked around the issue by adding an explicit type that shouldn’t need to be there.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -159,6 +163,13 @@ open class RouteOptions: DirectionsOptions {
      */
     open var includesExitRoundaboutManeuver = false
     
+    /**
+     A Boolean value indicating whether `Directions` can refresh time-dependent properties of the `RouteLeg`s of the resulting `Route`s.
+     
+     To refresh the `RouteLeg.expectedSegmentTravelTimes`, `RouteLeg.segmentSpeeds`, and `RouteLeg.segmentCongestionLevels` properties, use the `Directions.refreshRoute(responseIdentifier:routeIndex:fromLegAtIndex:completionHandler:)` method. This property is ignored unless `profileIdentifier` is `DirectionsProfileIdentifier.automobileAvoidingTraffic`.
+     */
+    open var refreshingEnabled = true
```
1ec5, line 171: The default values of these properties should match Directions API default values as much as possible. `refreshingEnabled` may not be relevant to applications that use the MapboxDirections library directly without the navigation SDK. So this property’s default value should be `false`, and NavigationRouteOptions in the navigation SDK would set the value to `true`.

PR #424, opened 2020-06-03
Roundtrip null voiceLocale

chezzdev:
This PR adds an internal property `speechLocaleIsSupported` to `DirectionsResult` which indicates whether the "voiceLocale" key was present in the original response.
This helps to encode the `DirectionsResult` object saving the original "voiceLocale" key-value or its absence.

The question we have to resolve to move this PR from the draft state is whether there's an external need for the `speechLocaleIsSupported` knowledge so we should either leave it `internal` or make it `public` and utilize it wherever needed.

cc: @mapbox/navigation-ios

Resolves https://github.com/mapbox/mapbox-directions-swift/issues/415

### DIFF
--- a/Sources/MapboxDirections/DirectionsResult.swift
+++ b/Sources/MapboxDirections/DirectionsResult.swift
@@ -77,7 +77,7 @@ open class DirectionsResult: Codable {
         try container.encode(distance, forKey: .distance)
         try container.encode(expectedTravelTime, forKey: .expectedTravelTime)
         try container.encodeIfPresent(routeIdentifier, forKey: .routeIdentifier)
-        try container.encodeIfPresent(speechLocale?.identifier, forKey: .speechLocale)
+        try container.encode(speechLocale?.identifier, forKey: .speechLocale)
     }
     
     // MARK: Getting the Shape of the Route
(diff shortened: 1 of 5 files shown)


Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -62,6 +63,8 @@ open class DirectionsResult: Codable {
         } else {
             speechLocale = nil
         }
+
+        speechLocaleIsSupported = container.contains(.speechLocale)
```
1ec5, line 67: Based on this logic, I think `speechLocaleIsSupported` should be renamed. It’s true if and only if `voiceLocale` is present in the JSON. In other words, it’s true as long as voice instructions were requested, whether or not the language is supported. If the language is unsupported, this property would be `false` but `speechLocale` would be `nil`.

Basically, we need to distinguish between “no speech locale” and “speech locale not applicable”, at least when it comes to the Mapbox Directions API. I can’t think of a great name for the property at the moment. One option would be to remove this property in favor of looking at whether the route options has `includesSpokenInstructions` set to `true`, but we should make sure there aren’t other cases where `voiceInstructions=true` in the request but `voiceLocale` is undefined in the response.

/cc @mandeepsandhu @danpaz

PR #425, opened 2020-06-03
Fabian junction views

fabian-guerra:
wip


Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -34,6 +34,8 @@ public extension VisualInstruction {
          */
         case image(image: ImageRepresentation, alternativeText: TextRepresentation)
         
+        case guidanceView(image: GuidanceViewImageRepresentation, alternativeText: TextRepresentation)
```
1ec5, line 37: Add documentation about this case; otherwise it won’t show up in documentation.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -149,13 +151,29 @@ public extension VisualInstruction.Component {
     }
 }
 
+public struct GuidanceViewImageRepresentation: Equatable {
```
1ec5, line 154: ImageRepresentation isn’t meant to be limited to the `image` case; it just means that it contains the details for rendering an image. So does this struct, but the difference is in how the URL is interpreted. I wonder if we should try to use ImageRepresentation for both cases but genericize it to make sense in both. If we must have a separate struct, we should probably rename the two structs to better reflect the distinction. Perhaps one would be ScalableImageRepresentation or something to that effect. Needs more thought.

/ref https://github.com/mapbox/mapbox-directions-swift/pull/405#discussion_r365487164

Tests/MapboxDirectionsTests/VisualInstructionTests.swift
```diff
@@ -24,6 +24,18 @@ class VisualInstructionsTests: XCTestCase {
                 "modifier": "right",
             ],
             "secondary": nil,
+            "view": [
+              "text": "CA01610_1_E",
+              "components": [
+                [
+                  "text": "CA01610_1_E",
+                  "type": "guidance-view",
+                  "url": "https://api-turn-here-staging-451578336.us-east-1.elb.amazonaws.com/guidance-views/v1/z/jct/CA01610_1_E"
```
1ec5, line 33: Remove this staging URL and replace it with a fake URL.

Tests/MapboxDirectionsTests/VisualInstructionTests.swift
```diff
@@ -36,12 +48,17 @@ class VisualInstructionsTests: XCTestCase {
...
             XCTAssertNil(banner.secondaryInstruction)
+            XCTAssertNotNil(banner.quaternaryInstruction)
             XCTAssertEqual(banner.drivingSide, .default)
         }
         
         let component = VisualInstruction.Component.text(text: .init(text: "Weinstock Strasse", abbreviation: nil, abbreviationPriority: nil))
         let primaryInstruction = VisualInstruction(text: "Weinstock Strasse", maneuverType: .turn, maneuverDirection: .right, components: [component])
-        banner = VisualInstructionBanner(distanceAlongStep: 393.3, primary: primaryInstruction, secondary: nil, tertiary: nil, drivingSide: .right)
+        
+        let guideViewComponent = VisualInstruction.Component.guidanceView(image: GuidanceViewImageRepresentation(imageURL: URL(string: "https://api-turn-here-staging-451578336.us-east-1.elb.amazonaws.com/guidance-views/v1/z/jct/CA01610_1_E")), alternativeText: VisualInstruction.Component.TextRepresentation(text: "CA01610_1_E", abbreviation: nil, abbreviationPriority: nil))
```
1ec5, line 58: Remove this staging URL and replace it with a fake URL.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -195,14 +235,22 @@ extension VisualInstruction.Component: Codable {
             if let imageBaseURLString = try container.decodeIfPresent(String.self, forKey: .imageBaseURL) {
                 imageBaseURL = URL(string: imageBaseURLString)
             }
-            let imageRepresentation = ImageRepresentation(imageBaseURL: imageBaseURL)
+            let imageRepresentation = ImageRepresentation(imageBaseURL: imageBaseURL, availableFormats: [.png, .svg])
```
1ec5, line 238: This representation is available at 1×, 2×, and 3× scales.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -124,11 +159,14 @@ public extension VisualInstruction.Component {
          - returns: A remote URL to the image.
          */
         public func imageURL(scale: CGFloat? = nil, format: Format = .png) -> URL? {
```
1ec5, line 161: Update the `scale` documentation to replace:

> Only the values 1, 2, and 3 are currently supported.

with:

> Only the values in the `availableScales` property are currently supported.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -124,11 +159,14 @@ public extension VisualInstruction.Component {
...
+            let imageScale = scale ?? ImageRepresentation.currentScale
             guard let imageBaseURL = imageBaseURL,
+                self.availableScales.contains(imageScale),
+                self.availableFormats.contains(format),
```
1ec5, line 165: Nit: it’s OK to omit `self.` here, but not a big deal.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -104,18 +109,48 @@ public extension VisualInstruction.Component {
...
+        /// - Parameters:
+        ///   - imageURL: The image URL.
+        ///   - availableScales: The image’s available scale factors. If this argument is unspecified, the current screen’s native scale factor is used.
+        ///   - availableFormats: The available file formats of the image.
+        public init(imageURL: URL?, availableScales: Set<CGFloat>? = nil, availableFormats: Set<Format>) {
+            self.imageURL = imageURL
+            self.imageBaseURL = nil
+            
+            let defaultScales: Set<CGFloat> = [ImageRepresentation.currentScale]
+            self.availableScales = availableScales ?? defaultScales
```
1ec5, line 136: Think of `availableScales` as the set of scales that can be _explicitly_ part of the URL, rather than the scale that the image happens to be in. So the default here should be an empty set.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -104,18 +109,48 @@ public extension VisualInstruction.Component {
...
+            self.availableFormats = availableFormats
+        }
+        
+        /// Initializes an image representation bearing the image at the given URL.
+        /// - Parameters:
+        ///   - imageURL: The image URL.
+        ///   - availableScales: The image’s available scale factors. If this argument is unspecified, the current screen’s native scale factor is used.
```
1ec5, line 129: > If this argument is unspecified, the current screen’s native scale factor is used.

This is true of `imageURL(scale:format:)`, but I think it would be better for a `nil` `availableScales` to mean “don’t append a scale to the URL”.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -124,11 +159,14 @@ public extension VisualInstruction.Component {
...
         public func imageURL(scale: CGFloat? = nil, format: Format = .png) -> URL? {
+            let imageScale = scale ?? ImageRepresentation.currentScale
             guard let imageBaseURL = imageBaseURL,
+                self.availableScales.contains(imageScale),
+                self.availableFormats.contains(format),
                 var imageURLComponents = URLComponents(url: imageBaseURL, resolvingAgainstBaseURL: false) else {
                 return nil
             }
-            imageURLComponents.path += "@\(Int(scale ?? ImageRepresentation.currentScale))x.\(format)"
+            imageURLComponents.path += "@\(Int(imageScale))x.\(format)"
```
1ec5, line 169: Does the Guidance Views API put `@1x` etc. and a file extension in the URL? If not, let’s use an empty `availableScales` or `availableFormats` set to detect that these things should not be added to the path.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -104,18 +109,48 @@ public extension VisualInstruction.Component {
...
         public let imageBaseURL: URL?
         
+        /// The image URL.
+        public let imageURL: URL?
+        
+        /// The image's available scales.
+        public let availableScales: Set<CGFloat>
+        
+        /// The image's available formats.
+        public let availableFormats: Set<Format>
```
1ec5, line 152: These properties aren’t being encoded, decoded, and compared explicitly because they’re derived from the component type, which is already being encoded, decoded, and compared.

1ec5:
https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/585/workflows/b57babdc-132a-4c5e-ac62-69117412d8b9/jobs/1725/steps

> Test Case '-[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent]' started.
> 
> /Users/distiller/project/Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift:39: error: -[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent] : XCTAssertEqual failed: ("nil") is not equal to ("Optional("https://s3.amazonaws.com/mapbox/shields/v3/us-42@1x.svg")")
> 
> Test Case '-[MapboxDirectionsTests.VisualInstructionComponentTests testImageComponent]' failed (0.008 seconds).

Sources/MapboxDirections/VisualInstructionComponent.swift
fabian-guerra: It doesn't its a direct URL that's the reason that I consider if you initialized an image with an imageURL you can use the property instead

1ec5: This line is unconditionally adding e.g. `@1x.png` to the URL. We should either avoid adding it to the URL or remove it from the URL before initializing the component. Let’s add a test to verify that the caller of this method will get the same URL that the Directions API returned for a guidance view image.

Issue #348, opened 2019-02-15
CoordinateBounds mixes up its corners, breaking offline tile downloads


1ec5:
`Directions.downloadTiles(in:version:session:completionHandler:)` always fails with the following error if the coordinate bounds is created using the `CoordinateBounds(northWest:southEast:)` initializer:

```
"BBox coordinates must be in 'minLon,minLat;maxLon,maxLat' format"
```

CoordinateBounds properties store and `CoordinateBounds.path` returns the northwest and southeast coordinates instead of the southwest and northeast coordinates that the routing tile API endpoint expects. Ironically, this error doesn’t occur when using the `CoordinateBounds(coordinates:)` initializer, because that convenience initializer stores the southwest and northeast coordinates in the `northWest` and `southEast` properties, respectively.

`northWest` and `southEast` should be replaced by `southWest` and `northEast`, but the older properties can be turned into computed variables for backwards compatibility and completeness.

/ref #303
/cc @frederoni @akitchen

PR #426, opened 2020-06-05
Turf update

### LINKED ISSUE #423: Upgrade to Turf v0.4.0
Once Turf v0.4.0 is released, we’ll need to upgrade this library to it. v0.4.0 comes with a number of backwards-incompatible changes, but hopefully most of those changes will be transparent to MapboxDirections. One bright spot is that Turf no longer defines a `Polyline` type alias for `LineString` (mapbox/turf-swift#93), so we no longer have to do gymnastics to use both Polyline.framework and Turf.framework in the same file:

https://github.com/mapbox/mapbox-directions-swift/blob/f2841b6235eceacc6e733406be714fb024e64725/Sources/MapboxDirections/Extensions/GeoJSON.swift#L3-L4

Another thing to watch out for is that Turf no longer makes `CLLocationCoordinate2D` publicly conform to `Codable` (mapbox/turf-swift#96), so it’s possible that MapboxDirections may need to reimplement that conformance privately.

/cc @mapbox/navigation-ios

Udumft:
Resolves #423 

Bumbed Turf version to `0.4.0`. Implemented now hidden `Codable` conformances, refined importing conflicts between `Turf` and `Polyline`.

### DIFF
--- a/Sources/MapboxDirections/CoordinateBounds.swift
+++ b/Sources/MapboxDirections/CoordinateBounds.swift
@@ -59,13 +59,51 @@ extension CoordinateBounds: Codable {
     
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-        southWest = try container.decode(CLLocationCoordinate2D.self, forKey: .southWest)
-        northEast = try container.decode(CLLocationCoordinate2D.self, forKey: .northEast)
+        southWest = try container.decode(CLLocationCoordinate2DCodable.self, forKey: .southWest).decodedCoordinates
+        northEast = try container.decode(CLLocationCoordinate2DCodable.self, forKey: .northEast).decodedCoordinates
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        try container.encode(southWest.codableCoordinates)
+        try container.encode(northEast.codableCoordinates)
     }
 }
 
 extension CoordinateBounds: CustomStringConvertible {
     public var description: String {
         return "\(southWest.longitude),\(southWest.latitude);\(northEast.longitude),\(northEast.latitude)"
+    }
+}
+
+struct CLLocationCoordinate2DCodable: Codable {
+    var latitude: CLLocationDegrees
+    var longitude: CLLocationDegrees
+    var decodedCoordinates: CLLocationCoordinate2D {
+        return CLLocationCoordinate2D(latitude: latitude,
+                                      longitude: longitude)
+    }
+    
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.unkeyedContainer()
+        try container.encode(longitude)
+        try container.encode(latitude)
+    }
+    
+    init(from decoder: Decoder) throws {
+        var container = try decoder.unkeyedContainer()
+        longitude = try container.decode(CLLocationDegrees.self)
+        latitude = try container.decode(CLLocationDegrees.self)
+    }
+    
+    init(_ coordinate: CLLocationCoordinate2D) {
+        latitude = coordinate.latitude
+        longitude = coordinate.longitude
+    }
+}
+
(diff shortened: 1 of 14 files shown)


Sources/MapboxDirections/CoordinateBounds.swift
```diff
@@ -59,8 +59,14 @@ extension CoordinateBounds: Codable {
     
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
-        southWest = try container.decode(CLLocationCoordinate2D.self, forKey: .southWest)
-        northEast = try container.decode(CLLocationCoordinate2D.self, forKey: .northEast)
+        southWest = try container.decode(CLLocationCoordinate2DCodable.self, forKey: .southWest).decodedCoordinates
+        northEast = try container.decode(CLLocationCoordinate2DCodable.self, forKey: .northEast).decodedCoordinates
```
1ec5, line 63: It would be tempting to remove `CoordinateBounds` in favor of the `BoundingBox` struct in Turf. Unfortunately, it looks like `BoundingBox` was copied from `CoordinateBounds` back before we fixed #348, so we’ll have to fix mapbox/turf-swift#102 before we can rely on `BoundingBox`.

Sources/MapboxDirections/Extensions/Codable.swift
```diff
@@ -28,12 +28,18 @@ enum PolyLineString {
...
+    }
+    
     init(from decoder: Decoder) throws {
         let container = try decoder.singleValueContainer()
         let options = decoder.userInfo[.options] as? DirectionsOptions
         switch options?.shapeFormat ?? .default {
         case .geoJSON:
-            self = .lineString(try container.decode(LineString.self))
+            let lineStringContainer = try decoder.container(keyedBy: LineStringCodingKeys.self)
+            let coordinates = try lineStringContainer.decode([CLLocationCoordinate2DCodable].self, forKey: .coordinates).map { $0.decodedCoordinates }
```
1ec5, line 41: Is it necessary to manually decode the LineString here? Can we continue to rely on Turf’s implementation of LineString’s Codable conformance?

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -519,7 +519,7 @@ open class RouteStep: Codable {
         try maneuver.encode(instructions, forKey: .instruction)
         try maneuver.encode(maneuverType, forKey: .type)
         try maneuver.encodeIfPresent(maneuverDirection, forKey: .direction)
-        try maneuver.encodeIfPresent(maneuverLocation, forKey: .location)
+        try maneuver.encode(CLLocationCoordinate2DCodable(maneuverLocation), forKey: .location)
```
1ec5, line 522: Are there any maneuver types for which there’s no `maneuverLocation` present?

Sources/MapboxDirections/Extensions/Codable.swift
Udumft: `Turf` does not provide `LineString` `Codable` conformance. So I decided we should not add it in `Directions` either.

Sources/MapboxDirections/RouteStep.swift
Udumft: As I understand what `Maneuver` is, it always has a location where it "occurs" or "vocalized". Also, in the code, `maneuverLocation` is non-nil.

Sources/MapboxDirections/Extensions/Codable.swift
1ec5: Ah, that’s a little bit awkward. But Geometry is codable, so we can continue to wrap the LineString in a Geometry and decode that, right?

Udumft: Yes, I think it could work for decoding, but encoding `Geometry` will also serialize it's `type`, which is undesired.

PR #430, opened 2020-06-14
Add ability to initialize route with visual and audio instructions.

### LINKED ISSUE #429: Support Route initialization
DirectionsResult `init:` is not public. This prevents its usage outside of framework with error:
`'Route' initializer is inaccessible due to 'internal' protection level`.

There is also problem with `RouteStep` initializer, properties like `intersections`, `instructionsSpokenAlongStep` and `instructionsDisplayedAlongStep` are assigned to `nil`, this leads to inability to create `Route` object and later use it in Navigation SDK.

MaximAlien:
Closes #429.

### DIFF
--- a/Sources/MapboxDirections/DirectionsResult.swift
+++ b/Sources/MapboxDirections/DirectionsResult.swift
@@ -20,8 +20,7 @@ open class DirectionsResult: Codable {
     }
     
     // MARK: Creating a Directions Result
-    
-    init(legs: [RouteLeg], shape: LineString?, distance: CLLocationDistance, expectedTravelTime: TimeInterval) {
+    public init(legs: [RouteLeg], shape: LineString?, distance: CLLocationDistance, expectedTravelTime: TimeInterval) {
         self.legs = legs
         self.shape = shape
         self.distance = distance
--- a/Sources/MapboxDirections/DirectionsResult.swift
+++ b/Sources/MapboxDirections/DirectionsResult.swift
@@ -20,6 +20,7 @@ open class DirectionsResult: Codable {
     }
     
     // MARK: Creating a Directions Result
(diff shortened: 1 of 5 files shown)


Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -20,8 +20,7 @@ open class DirectionsResult: Codable {
     }
     
     // MARK: Creating a Directions Result
-    
-    init(legs: [RouteLeg], shape: LineString?, distance: CLLocationDistance, expectedTravelTime: TimeInterval) {
+    public init(legs: [RouteLeg], shape: LineString?, distance: CLLocationDistance, expectedTravelTime: TimeInterval) {
```
1ec5, line 23: Per the class documentation, it is intentional that clients can’t create DirectionsResult instances directly, because DirectionsResult is an abstract superclass of Route and Match. #406 incorrectly removed all the public initializers from both classes. Can you restore them to those classes instead? Thanks!

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -475,11 +475,11 @@ open class RouteStep: Codable {
         self.codes = codes
         self.destinationCodes = destinationCodes
         self.destinations = destinations
-        self.intersections = nil
+        self.intersections = intersections
```
1ec5, line 478: Good catch, thanks!

Sources/MapboxDirections/DirectionsResult.swift
MaximAlien: Thanks for pointing that out. Added public init for `Route`, inits were available for `Match`, but public API docs had to be updated.

PR #432, opened 2020-06-17
Fix guidance views json parsing.

fabian-guerra:
Fixes an issue that caused the image url in guidance views to be nil.

### DIFF
--- a/Sources/MapboxDirections/VisualInstructionComponent.swift
+++ b/Sources/MapboxDirections/VisualInstructionComponent.swift
@@ -177,7 +177,7 @@ extension VisualInstruction.Component: Codable {
         case abbreviatedText = "abbr"
         case abbreviatedTextPriority = "abbr_priority"
         case imageBaseURL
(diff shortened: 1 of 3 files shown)


Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -177,7 +177,7 @@ extension VisualInstruction.Component: Codable {
         case abbreviatedText = "abbr"
         case abbreviatedTextPriority = "abbr_priority"
         case imageBaseURL
-        case imageURL = "url"
+        case imageURL
```
1ec5, line 180: The raw value of this coding key still needs to be `url` for shields, right? Let’s create a separate coding key for that case then.

1ec5: Never mind, the `image` case uses the `imageBaseURL` coding key one line above. This is fine.

PR #433, opened 2020-06-17
Update version to 0.33.0

fabian-guerra:
Bumps directions version to 0.33.0


Directions Example/Info.plist
```diff
@@ -19,7 +19,7 @@
 	<key>CFBundleSignature</key>
```
1ec5, line 19: Looks like `agvtool new-marketing-version 0.33.0` also needs to be run on the repository. The line above here still says `0.31.0`.

PR #434, opened 2020-06-18
Fix crash on Example app startup.

### LINKED ISSUE #431: MapboxDirections example application crashes on startup
Example target provided in MapboxDirections repo crashes on startup with error:
```
dyld: Library not loaded: @rpath/MapboxMobileEvents.framework/MapboxMobileEvents
  Referenced from: /Users/maksimmakhun/Library/Developer/CoreSimulator/Devices/ABC90875-5F0D-44B9-A1A4-C3F4F60C6DE5/data/Containers/Bundle/Application/03CE6589-7FC4-4FD3-81E9-6E970B8CA8F8/Example.app/Frameworks/Mapbox.framework/Mapbox
  Reason: image not found
```
Before running Example all dependencies were installed correctly by running: `$ carthage update`.
To fix crash `Cartfile` should be updated to contain stable version of `MapboxMobileEvents.framework` used by Navigation SDK for iOS.

MaximAlien:
Fixes #431.


Cartfile
```diff
@@ -1,2 +1,3 @@
 github "raphaelmor/Polyline" ~> 4.2
 github "mapbox/turf-swift" ~> 0.5
+github "mapbox/mapbox-events-ios" ~> 0.10
```
1ec5, line 3: This should be a private dependency, not a public one. Only the example application target depends on the map SDK, which is what depends on MapboxMobileEvents.

MaximAlien: Thanks, moved `mapbox-events-ios` to `Cartfile.private`.

Cartfile
```diff
@@ -1,2 +1,2 @@
 github "raphaelmor/Polyline" ~> 4.2
-github "mapbox/turf-swift" ~> 0.5
+github "mapbox/turf-swift" ~> 0.5
```
dersim-davaod, line 2: Could you please add newline characters at the end of touched files?
Screen Shot 2020-07-09 at 16 54 04

MaximAlien: Fixed.

PR #437, opened 2020-06-20
Add Options to JSON Encoder in RouteResponse

hactar:
When initialising a RouteResponse, the decoder receives the `.options` passed in. The temporary encoder to encode the routes however does not. This can create a mismatched RouteResponse object, for example if the options state that `polyline6` was used.

This PR also passes the `.options` to the temporary encoder and so fixes an issue for us where the routes are off by 1 decimal position when using `polyline6`.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,2 +1,3 @@
 binary "https://www.mapbox.com/ios-sdk/Mapbox-iOS-SDK.json" ~> 5.5
 github "linksmt/OHHTTPStubs" "563f48d3fab84ef04639649c770b00f4fa502cca"
+github "mapbox/mapbox-events-ios" ~> 0.10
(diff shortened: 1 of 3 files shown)


Udumft:
Hello, @hactar! Thank you for contribution!
I was trying to reproduce the issue you described, but in my attempts `RouteResponse` object was always decoded correctly, whether I used `.polyline6` or usual `.polyline` format. Could you please provide some more details or a snipped to demonstrate it?

1ec5:
Sorry @hactar, I screwed up and force-pushed this repository’s master to your master branch. I had intended to push an additional commit to your branch, but it was named master, which meant I had to stage it on a branch with a different name on my end. I’ll open a different PR with your contribution. Please feel free to force-push your local clone back to your fork. Also, it’s a good idea to PR branches other than master in the future, to avoid mishaps like this. Thanks, and sorry for the inconvenience!

Issue #445, opened 2020-08-18
Decoded RouteResponse doesn't contain targetCoordinate if it was passed via ResponseOptions.


MaximAlien:
In case if `RouteResponse` is encoded with `targetCoordinate`, `heading`, `headingAccuracy`, `separatesLegs` and `allowsArrivingOnOppositeSide` being set and then decoded all stated properties will not be saved in result object. It is expected that encoded and decoded `RouteResponse` objects have the same properties except `coordinate`.

PR #446, opened 2020-08-18
Improve route response decoding.

MaximAlien:
Closing #445.

### DIFF
--- a/Sources/MapboxDirections/RouteResponse.swift
+++ b/Sources/MapboxDirections/RouteResponse.swift
@@ -107,8 +107,16 @@ extension RouteResponse: Codable {
             // The response lists the same number of tracepoints as the waypoints in the request, whether or not a given waypoint is leg-separating.
             waypoints = zip(decodedWaypoints, optionsWaypoints).map { (pair) -> Waypoint in
                 let (decodedWaypoint, waypointInOptions) = pair
-                let waypoint = Waypoint(coordinate: decodedWaypoint.coordinate, coordinateAccuracy: waypointInOptions.coordinateAccuracy, name: waypointInOptions.name?.nonEmptyString ?? decodedWaypoint.name)
+                let waypoint = Waypoint(coordinate: decodedWaypoint.coordinate,
+                                        coordinateAccuracy: waypointInOptions.coordinateAccuracy,
+                                        name: waypointInOptions.name?.nonEmptyString ?? decodedWaypoint.name)
+
+                waypoint.targetCoordinate = waypointInOptions.targetCoordinate
+                waypoint.heading = waypointInOptions.heading
+                waypoint.headingAccuracy = waypointInOptions.headingAccuracy
                 waypoint.separatesLegs = waypointInOptions.separatesLegs
+                waypoint.allowsArrivingOnOppositeSide = waypointInOptions.allowsArrivingOnOppositeSide
+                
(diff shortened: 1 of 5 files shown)


Sources/MapboxDirections/Waypoint.swift
```diff
@@ -215,7 +215,13 @@ extension Waypoint: Equatable {
...
 extension Waypoint: CustomStringConvertible {
     public var description: String {
-        return name ?? "<latitude: \(coordinate.latitude); longitude: \(coordinate.longitude)>"
+        return Mirror(reflecting: self).children.compactMap({
+            if let label = $0.label {
+                return "\(label): \($0.value)"
+            }
+            
+            return ""
+        }).joined(separator: "\n")
```
1ec5, line 224: Not bad. What do you think about further customizing the Quick Look implementation of this type to show the target coordinate when present instead of the main coordinate?

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -107,8 +107,16 @@ extension RouteResponse: Codable {
...
-                let waypoint = Waypoint(coordinate: decodedWaypoint.coordinate, coordinateAccuracy: waypointInOptions.coordinateAccuracy, name: waypointInOptions.name?.nonEmptyString ?? decodedWaypoint.name)
+                let waypoint = Waypoint(coordinate: decodedWaypoint.coordinate,
+                                        coordinateAccuracy: waypointInOptions.coordinateAccuracy,
+                                        name: waypointInOptions.name?.nonEmptyString ?? decodedWaypoint.name)
+
+                waypoint.targetCoordinate = waypointInOptions.targetCoordinate
+                waypoint.heading = waypointInOptions.heading
+                waypoint.headingAccuracy = waypointInOptions.headingAccuracy
                 waypoint.separatesLegs = waypointInOptions.separatesLegs
+                waypoint.allowsArrivingOnOppositeSide = waypointInOptions.allowsArrivingOnOppositeSide
```
1ec5, line 118: Aside from the tests below, what would be a good way to prevent a future option from getting overlooked this way? How about copying the waypointInOptions and overwriting the fewer properties that are available in the decodedWaypoint. This could be good fodder for tail work; not essential for this PR.

1ec5: Then again, #388 would make a finer distinction between a requested waypoint and a received waypoint, such that the developer would be responsible for keeping track of the requested waypoints’ details separately, mirroring the API response format. But until that change lands someday, this is a good change to have for consistency.

MaximAlien: I was thinking about using a copy of `waypointInOptions` and assigning only minimum number of essential properties, but `public let coordinate: CLLocationCoordinate2D` prevents me from assigning it from `decodedWaypoint`. I'm not sure why coordinate is declared as `let` whereas others use `var`, but if that's okay to make coordinate var as well it'd make it possible.

Sources/MapboxDirections/Waypoint.swift
MaximAlien: Interestingly `debugQuickLookObject` doesn't work for me in Swift, when using Objective-C it works as expected. Will add `targetCoordinate ?? coordinate` though.

PR #461, opened 2020-09-15
Republicize DirectionsOptions.urlQueryItems

### LINKED ISSUE #460: Subclasses of DirectionsOptions cannot override urlQueryItems
#343 added an overridable `DirectionsOptions.urlQueryItems` property for the purpose of enabling applications to take advantage of not-yet-ready-for-primetime Directions API parameters. Unfortunately, 593efd3d16810cb8f1a4f28c487e5340d0642dff for #382 made the property internal. It was probably an oversight, and I think we can safely reexpose the property. MatchOptions is unaffected; its `urlQueryItems` property remains `open`.

Until we reexpose `DirectionsOptions.urlQueryItems` publicly, the workaround would be to take the result of `Directions.url(forCalculating:)` or similar, use `URLComponents` to parse the URL and add more `urlQueryItems`, get the URL back out, request that URL using `URLSession`, and use `JSONDecoder` to decode the response as a `RouteResponse` – basically, everything that `Directions.calculate(_:completionHandler:)` normally does for you.

/cc @mapbox/navigation-ios

1ec5:
Made `DirectionsOptions.urlQueryItems` open once again.

Fixes #460.

/cc @mapbox/navigation-ios @jmkiley @captainbarbosa @geografa


1ec5:
Once this PR lands, you’ll be able to try out the Mapbox Directions API’s `depart_at` parameter (currently in beta) by subclassing RouteOptions (or NavigationRouteOptions if you’re using the navigation SDK). For example, this RouteOptions subclass requests directions for a trip that would begin an hour from now:

```swift
// Subclass NavigationRouteOptions instead if using the navigation SDK
class DeferredRouteOptions: RouteOptions {
    var startDate = Date()
    
    override open var urlQueryItems: [URLQueryItem] {
        let isoHourFromNow = ISO8601DateFormatter().string(from: startDate, timeZone: .autoupdatingCurrent)
        return super.urlQueryItems + [
            URLQueryItem(name: "depart_at", value: isoHourFromNow),
        ]
    }
}

let options = DeferredRouteOptions(coordinates: coordinates, profileIdentifier: .automobile)
options.startDate = Date() + 60 * 60
…
```

MaximAlien (approved):

Issue #459, opened 2020-09-14
Publicly expose duration_typical value.


MaximAlien:
To be able to get delay information `duration_typical` should be publicly exposed so that it can be used by Navigation SDK.

1ec5:
Fixed in #462.

PR #462, opened 2020-09-15
Expose duration_typical via public API.

MaximAlien:
Closing #459.

### DIFF
--- a/Sources/MapboxDirections/DirectionsResult.swift
+++ b/Sources/MapboxDirections/DirectionsResult.swift
@@ -14,6 +14,7 @@ open class DirectionsResult: Codable {
         case legs
         case distance
         case expectedTravelTime = "duration"
+        case typicalTravelTime = "duration_typical"
         case directionsOptions
         case routeIdentifier
         case speechLocale = "voiceLocale"
@@ -21,11 +22,12 @@ open class DirectionsResult: Codable {
     
     // MARK: Creating a Directions Result
     
-    init(legs: [RouteLeg], shape: LineString?, distance: CLLocationDistance, expectedTravelTime: TimeInterval) {
+    init(legs: [RouteLeg], shape: LineString?, distance: CLLocationDistance, expectedTravelTime: TimeInterval, typicalTravelTime: TimeInterval? = nil) {
         self.legs = legs
         self.shape = shape
         self.distance = distance
         self.expectedTravelTime = expectedTravelTime
+        self.typicalTravelTime = typicalTravelTime
         self.responseContainsSpeechLocale = false
     }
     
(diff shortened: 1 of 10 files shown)


Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -139,6 +143,11 @@ open class DirectionsResult: Codable {
      */
     open var expectedTravelTime: TimeInterval
     
+    /**
+     The route’s typical travel time, measured in seconds.
```
1ec5, line 147: Should we mention any caveats to this property, such as the ones in the `expectedTravelTime` documentation above?

* Does it work in every profile or just driving profiles?
* Presumably these times are still subject to the vagaries of ferry schedules.
* The developer shouldn’t assume the user would travel along the route at a fixed speed. Does the Directions API expose an attribute option (attribute) for typical travel times as it does for expected travel times?

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -712,6 +717,11 @@ open class RouteStep: Codable {
      */
     open var expectedTravelTime: TimeInterval
     
+    /**
+     The step’s typical travel time, measured in seconds.
```
1ec5, line 721: Also mention any caveats here.

Sources/MapboxDirections/DirectionsResult.swift
```diff
@@ -139,6 +143,11 @@ open class DirectionsResult: Codable {
...
+     */
+    open var typicalTravelTime: TimeInterval?
```
1ec5, line 149: This property is on DirectionsResult rather than Route, so it’s also available on Match. Does the Map Matching API expose this value as well?

1ec5 (approved):
Looks good code-wise, with some opportunities for improved documentation.

Suggested changelog entry:

* Added the `DirectionsResult.typicalTravelTime` and `RouteStep.typicalTravelTime` properties that indicate the typical travel time, as opposed to the current expected travel time. ([#462](https://github.com/mapbox/mapbox-directions-swift/pull/462))

Sources/MapboxDirections/DirectionsResult.swift
MaximAlien: Just did a few tests related to Map Matching API and don't see that `duration_typical` is exposed.

Sources/MapboxDirections/DirectionsResult.swift
MaximAlien: I've added more details about `duration_typical`:
- `duration_typical` is only available when using `driving-traffic` profile.
- Tested with route which contains ferry and both `duration` and `duration_typical` are affected, mentioned this in docs.
- I'll mention that user shouldn't make an assumption that travel along the route will be at a fixed speed. I also don't see that `duration_typical` is supported by `AttributeOptions`.

@danpaz, not sure if you're the right person to ask, but would it be possible to update documentation in https://docs.mapbox.com/api/navigation/#retrieve-directions to contain `duration_typical` related info?

PR #466, opened 2020-09-30
Route alerts and urban area attribute

avi-c:
Add Intersection attributes for passive maneuvers and urban areas. Add admin regions to RouteLeg.

### DIFF
--- a/Sources/MapboxDirections/AdministrationRegion.swift
+++ b/Sources/MapboxDirections/AdministrationRegion.swift
@@ -1,0 +1,24 @@
+import Foundation
+
+public struct AdministrationRegion: Codable, Equatable {
+
+    private enum CodingKeys: String, CodingKey {
+        case countryCodeAlpha3 = "iso_3166_1_alpha3"
+        case countryCode = "iso_3166_1"
+    }
+
+    public var countryCodeAlpha3: String
+    public var countryCode: String
+
+    public init(from decoder: Decoder) throws {
+        let values = try decoder.container(keyedBy: CodingKeys.self)
+        countryCodeAlpha3 = try values.decode(String.self, forKey: .countryCodeAlpha3)
+        countryCode = try values.decode(String.self, forKey: .countryCode)
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(countryCodeAlpha3, forKey: .countryCodeAlpha3)
+        try container.encode(countryCode, forKey: .countryCode)
+    }
+}
--- a/Sources/MapboxDirections/AdministrationRegion.swift
+++ b/Sources/MapboxDirections/AdministrationRegion.swift
@@ -1,5 +1,8 @@
 import Foundation
 
+/**
+ `AdministrationRegion` describes corresponding object on the route.
+ */
 public struct AdministrationRegion: Codable, Equatable {
 
     private enum CodingKeys: String, CodingKey {
@@ -10,15 +13,8 @@ public struct AdministrationRegion: Codable, Equatable {
     public var countryCodeAlpha3: String
     public var countryCode: String
 
-    public init(from decoder: Decoder) throws {
-        let values = try decoder.container(keyedBy: CodingKeys.self)
-        countryCodeAlpha3 = try values.decode(String.self, forKey: .countryCodeAlpha3)
-        countryCode = try values.decode(String.self, forKey: .countryCode)
-    }
-
-    public func encode(to encoder: Encoder) throws {
-        var container = encoder.container(keyedBy: CodingKeys.self)
-        try container.encode(countryCodeAlpha3, forKey: .countryCodeAlpha3)
-        try container.encode(countryCode, forKey: .countryCode)
+    public init(countryCode: String, countryCodeAlpha3: String) {
+        self.countryCode = countryCode
+        self.countryCodeAlpha3 = countryCodeAlpha3
     }
 }
--- a/Sources/MapboxDirections/Incident.swift
+++ b/Sources/MapboxDirections/Incident.swift
@@ -1,0 +1,67 @@
+import Foundation
+
+public struct Incident: Codable, Equatable {
+
+    private enum CodingKeys: String, CodingKey {
+        case identifier = "id"
+        case type
+        case description = "description"
+        case creationTime = "creation_time"
+        case startTime = "start_time"
+        case endTime = "end_time"
+        case impact = "impact"
+        case subtype = "sub_type"
+        case subtypeDescription = "sub_type_description"
+        case alertCodes = "alertc_codes"
+        case lanesBlocked = "lanes_blocked"
+        case geometryIndexStart = "geometry_index_start"
+        case geometryIndexEnd = "geometry_index_end"
+    }
+
+    var identifier: UInt
+    var type: String
+    var description: String
+    var creationTime: String
+    var startTime: String
+    var endTime: String
+    var impact: String
+    var subtype: String?
+    var subtypeDescription: String?
+    var alertCodes: [Int]
+    var lanesBlocked: [Int]
+    var geometryIndexStart: Int
+    var geometryIndexEnd: Int
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.type = try container.decode(String.self, forKey: .type)
+        self.identifier = try container.decode(UInt.self, forKey: .identifier)
+        self.description = try container.decode(String.self, forKey: .description)
+        self.creationTime = try container.decode(String.self, forKey: .creationTime)
+        self.startTime = try container.decode(String.self, forKey: .startTime)
+        self.endTime = try container.decode(String.self, forKey: .endTime)
+        self.impact = try container.decode(String.self, forKey: .impact)
+        self.subtype = try container.decodeIfPresent(String.self, forKey: .subtype)
+        self.subtypeDescription = try container.decodeIfPresent(String.self, forKey: .subtypeDescription)
+        self.alertCodes = try container.decode([Int].self, forKey: .alertCodes)
+        self.lanesBlocked = try container.decode([Int].self, forKey: .lanesBlocked)
+        self.geometryIndexStart = try container.decode(Int.self, forKey: .geometryIndexStart)
+        self.geometryIndexEnd = try container.decode(Int.self, forKey: .geometryIndexEnd)
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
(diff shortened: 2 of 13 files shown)


avi-c:
Intended to implement the data model support for https://github.com/mapbox/navigation-sdks/issues/552

Udumft:
Did you consider wrapping route alerts into `enum`, similar to what [we have](https://github.com/mapbox/turf-swift/blob/main/Sources/Turf/Geometry.swift) in `Turf.swift` for geometries? 
Thus there are still separate `struct`s for each type fo alert existing + possibility to make base `RouteAlert` implementation that will encapsulate common parts and act as a "placeholder" for possible new types of alerts coming in the future?  Also, having such `enum` will allow `Intersection` and `RouteLeg` classes to have just an array of `alerts` instead of named properties for each existing alert type.

Udumft:
CI fail requires [cathage workaround](https://github.com/Carthage/Carthage/issues/3019#issuecomment-665136323) to be applied. Don't think it is a scope of current PR.

avi-c:
@MaximAlien, @Udumft  - Are we ok to merge in a change that requires the Carthage workaround? Is a separate PR adding the workaround and fixing the CI a prerequisite to merging this one in?

1ec5:
> CI fail requires [cathage workaround](https://github.com/Carthage/Carthage/issues/3019#issuecomment-665136323) to be applied. Don't think it is a scope of current PR.

mapbox/mapbox-directions-swift#470

avi-c:
@1ec5 Is this ready now then?

Sources/MapboxDirections/Intersection.swift
```diff
@@ -72,6 +84,18 @@ public struct Intersection {
      If road class information is unavailable, this property is set to `nil`.
      */
     public let outletRoadClasses: RoadClasses?
+
+    public let tunnelName: String?
```
MaximAlien, line 88: Shall we add more descriptive comments to newly added properties?

Sources/MapboxDirections/RouteAlert.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  RouteAlert.swift
+//  MapboxDirections
+//
+//  Created by Avi Cieplinski on 10/6/20.
+//  Copyright © 2020 Mapbox. All rights reserved.
+//
+
+import Foundation
```
MaximAlien, line 9: I think we do not use Copyright header sections across this repo, so probably better to remove this one as well.

MaximAlien (approved):

avi-c:
@1ec5 - I'd love a final 👍 if you have a chance

Sources/MapboxDirections/RouteAlert.swift
avi-c: This file isn't needed and I'll remove it.

Sources/MapboxDirections/Intersection.swift
avi-c: Added some comments across the new properties.

1ec5 (commented):
Sorry for the late review. Most of this PR is OK, but the administrative region portion needs to be redesigned based on #477. To avoid breaking backwards compatibility, I’d recommend one of the following approaches:

* Remove the administrative region properties if they aren’t essential to the route alerts feature
* Hold off on a final v1.1.0 release until we make the changes suggested in #477 (releasing RCs in the meantime)
* Leave undocumented any new parameters, properties, and types related to administrative regions so we have the flexibility to change them in v1.2.0.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -72,6 +84,48 @@ public struct Intersection {
...
+     If the information is unavailable, this property is set to `nil`.
+     */
+    public let administrationRegionIndex: Int?
+
+    /**
+     The index of the RouteStep within a RouteLeg that contains this Intersection.
+
+     This property is set to `nil` if unavailable.
+     */
+    let geometryIndex: Int?
```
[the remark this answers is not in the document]
1ec5: @danpaz also pointed out that the documentation comment for `geometryIndex` is misleading: it’s more like the index of the coordinate in `RouteStep.shape` that corresponds to `Intersection.location`. But only if the developer never reuses this Intersection object in the context of any other step, leg, or route or creates it independently. #478 proposes replacing `geometryIndex` with a property on RouteStep to avoid the semantic problem.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -72,6 +84,48 @@ public struct Intersection {
...
+
+    /**
+     The type of toll collection mechanism.
+
+     If this Intersection is not a toll collection intersection, or if this information is unavailable then this property is set to `nil`.
+     */
+    public let tollCollection: TollCollection?
+
+    /**
+     The type of rest stop.
```
1ec5, line 103: RestStop itself contains a `type` property, so it seems like this `restStop` property is potentially a place to store more information about a rest stop?

Sources/MapboxDirections/AdministrationRegion.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/**
+ `AdministrationRegion` describes corresponding object on the route.
+ */
+public struct AdministrationRegion: Codable, Equatable {
```
1ec5, line 6: Administrat**ive**Region would be more grammatical.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -236,6 +267,10 @@ open class RouteLeg: Codable {
      Do not assume that the user would travel along the leg at a fixed speed. For the expected travel time on each individual segment along the leg, use the `RouteStep.expectedTravelTimes` property. For more granularity, specify the `AttributeOptions.expectedTravelTime` option and use the `expectedSegmentTravelTimes` property.
      */
     open var expectedTravelTime: TimeInterval
+
+    open var administrationRegions: [AdministrationRegion]?
+
+    open var incidents: [Incident]?
```
1ec5, line 273: This property is undocumented. We shouldn’t use it in the navigation SDK until we document it.

Sources/MapboxDirections/Incident.swift
```diff
@@ -0,0 +1,65 @@
...
+    var creationTime: String
+    var startTime: String
+    var endTime: String
+    var impact: String?
+    var subtype: String?
+    var subtypeDescription: String?
+    var alertCodes: [Int]
+    var lanesBlocked: [Int]
+    var geometryIndexStart: Int
+    var geometryIndexEnd: Int
```
1ec5, line 36: These properties are all internal, so they aren’t available to clients such as the navigation SDK.

Sources/MapboxDirections/TollCollection.swift
```diff
@@ -0,0 +1,22 @@
...
+ `TollCollection` describes corresponding object on the route.
+ */
+public struct TollCollection: Codable, Equatable {
+
+    public enum CollectionType: String, Codable {
+        case booth = "toll_booth"
+        case gantry = "toll_gantry"
+    }
+
+    let type: CollectionType
```
1ec5, line 13: This property is internal, so clients such as the navigation SDK don’t have access to it.

Sources/MapboxDirections/RestStop.swift
```diff
@@ -0,0 +1,22 @@
...
+ `RestStop` describes corresponding object on the route.
+ */
+public struct RestStop: Codable, Equatable {
+
+    public enum StopType: String, Codable {
+        case serviceArea = "service_area"
+        case restArea = "rest_area"
+    }
+
+    let type: StopType
```
1ec5, line 13: This property is internal, so clients such as the navigation SDK don’t have access to it.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -141,6 +159,19 @@ open class RouteLeg: Codable {
         return segmentRangesByStep
     }()
     
+    /**
+     Segments for each Intersection along the route.
+     
+     Ordered by `steps`, inside one `step` - ordered by `Intersection`.  `nil` value means no index was provided. Index values correspond to `route`'s `shape` elements.
+     */
+    public private(set) lazy var intersectionsIndexesByStep: [[Int?]?] = {
```
1ec5, line 167: Consider renaming to `intersectionIndicesByStep`, though #478 proposes replacing this property with `RouteStep.segmentRangesByIntersection`, since it’s trivial to aggregate those step-based arrays at the leg level.

Sources/MapboxDirections/Incident.swift
```diff
@@ -0,0 +1,65 @@
...
+        case geometryIndexStart = "geometry_index_start"
+        case geometryIndexEnd = "geometry_index_end"
+    }
+
+    var identifier: String
+    var type: String
+    var description: String
+    var creationTime: String
+    var startTime: String
+    var endTime: String
```
1ec5, line 29: In Swift, any timestamp should be represented as a Date and called e.g. `createdDate` or `dateCreated`.

Sources/MapboxDirections/Incident.swift
```diff
@@ -0,0 +1,65 @@
...
+        case subtype = "sub_type"
+        case subtypeDescription = "sub_type_description"
+        case alertCodes = "alertc_codes"
+        case lanesBlocked = "lanes_blocked"
+        case geometryIndexStart = "geometry_index_start"
+        case geometryIndexEnd = "geometry_index_end"
+    }
+
+    var identifier: String
+    var type: String
```
1ec5, line 25: What are the valid values of `type` and `subtype`, or is they freeform? If there are sets of valid values, we should make them into enumerations ideally.

Sources/MapboxDirections/Incident.swift
1ec5, line 36: Combine these two properties into a Range.

1ec5 (commented):
Some additional feedback about the route alert feature that was at the center of this PR, to be addressed as tail work.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -14,7 +14,13 @@ public struct Intersection {
                 outletIndexes: IndexSet,
                 approachLanes: [LaneIndication]?,
                 usableApproachLanes: IndexSet?,
-                outletRoadClasses: RoadClasses?) {
+                outletRoadClasses: RoadClasses?,
+                tollCollection: TollCollection?,
+                tunnelName: String?,
+                restStop: RestStop?,
+                isUrban: Bool?,
+                administrationRegionIndex: Int?,
```
1ec5, line 22: Adding arguments to an initializer is a backwards-incompatible change. To mitigate that change, give each argument a default value.

avi-c:
@1ec5 - Your comments left here are very valuable and the issues should be addressed. Is this going to take place in one of the existing tickets for finishing up Route Alerts or do we need a separate ticket to make sure that the items are addressed? cc @Udumft

Issue #471, opened 2020-10-23
Replace Xcode 10 with Xcode 12 on CI


1ec5:
The CircleCI configuration currently builds this library in Xcode 10, which is practically unused and no longer supported by the navigation SDK. Meanwhile, we have no build coverage on Xcode 12, which is the focus of development. We should replace the Xcode 10 build bot with an Xcode 12 one.

Carthage is currently incompatible with Xcode 12: Carthage/Carthage#3019. As a temporary workaround, we’ll need to copy the [wcarthage.sh](https://github.com/mapbox/mapbox-navigation-ios/blob/89f6cf503b1064e80da220c13ab5962c81630d5c/scripts/wcarthage.sh) script originally added in mapbox/mapbox-navigation-ios#2649.

/cc @mapbox/navigation-ios

1ec5:
Dropping Xcode 10 from CI also implies that Xcode 10 will no longer be supported as a build environment.

avi-c:
CircleCI is updated but we are still getting unrelated errors building MME.

avi-c:
Fixed now with #472. MME errors taken care of by #474

PR #472, opened 2020-10-23
Update CircleCI to include an Xcode 12 build and drop Xcode 10 CI job.

avi-c:
Update CircleCI config with newer Xcode versions and workaround for Carthage incompatibility with Xcode 12

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,10 +87,10 @@ jobs:
       - run:
           name: Dependencies
           command: |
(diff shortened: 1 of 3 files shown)


avi-c:
Fix for #471

1ec5 (approved):
Looks like the readme doesn’t explicitly say which Xcode versions are supported. Mind adding a line about that under “Getting started”?

avi-c:
Added comment line in README as @1ec5 suggests.

PR #480, opened 2020-11-02
Fix error requesting route with silent waypoints and waypoint targets

### LINKED ISSUE #479: Error requesting route with silent waypoints and waypoint targets
When requesting a route, where one of the waypoints has `.separatesLegs = false`, route request returns an error:
"The operation couldn’t be completed. Number of waypoint_targets must match number of coordinates or waypoints provided."

I think this comes from incorrect ["waypoint_targets" parameter composing](https://github.com/mapbox/mapbox-directions-swift/blob/main/Sources/MapboxDirections/RouteOptions.swift#L205-L208) as it doesn't respect silent waypoints and includes data from all of them.

https://github.com/mapbox/mapbox-directions-swift/blob/6739b8bb021e074355e51eeddc33017a20d8ea9e/Sources/MapboxDirections/RouteOptions.swift#L205-L208

Udumft:
Resolves #479 
Added filtering silent waypoints when composing "waypoint_targets". Otherwise it added excessive entity which ruins endpoint response.

Note that [`waypoint_targets` doc](https://docs.mapbox.com/api/navigation/#directions) states that number of "targets" must match `coordinates` but in fact it produces an error if it doesn't match `waypoints` count (which in it's turn lists non-silent waypoints only). I've [reached out](https://mapbox.slack.com/archives/C03KW1JQD/p1604320592123400) to `Directions` team for clarification.

### DIFF
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -203,7 +203,7 @@ open class RouteOptions: DirectionsOptions {
         }
         
         if waypoints.first(where: { $0.targetCoordinate != nil }) != nil {
-            let targetCoordinates = waypoints.map { $0.targetCoordinate?.requestDescription ?? "" }.joined(separator: ";")
+            let targetCoordinates = waypoints.filter { $0.separatesLegs == true }.map { $0.targetCoordinate?.requestDescription ?? "" }.joined(separator: ";")
             params.append(URLQueryItem(name: "waypoint_targets", value: targetCoordinates))
         }
 
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -203,7 +203,7 @@ open class RouteOptions: DirectionsOptions {
         }
         
         if waypoints.first(where: { $0.targetCoordinate != nil }) != nil {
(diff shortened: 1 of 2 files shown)


Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -203,7 +203,7 @@ open class RouteOptions: DirectionsOptions {
         }
         
         if waypoints.first(where: { $0.targetCoordinate != nil }) != nil {
-            let targetCoordinates = waypoints.map { $0.targetCoordinate?.requestDescription ?? "" }.joined(separator: ";")
+            let targetCoordinates = waypoints.filter { $0.separatesLegs == true }.map { $0.targetCoordinate?.requestDescription ?? "" }.joined(separator: ";")
```
1ec5, line 206: Nit: this property isn’t optional:

```suggestion
            let targetCoordinates = waypoints.filter { $0.separatesLegs }.map { $0.targetCoordinate?.requestDescription ?? "" }.joined(separator: ";")
```

1ec5 (approved):
Thanks! Suggested changelog entry:

* Fixed an error that occurred when setting the `Waypoint.separatesLegs` property to `true` and setting the `Waypoint.targetCoordinate` property. ([#480](https://github.com/mapbox/mapbox-directions-swift/pull/480))

MaximAlien (approved):

PR #485, opened 2020-11-03
Administrative regions by leg and intersection

### LINKED ISSUE #477: Decode administrative regions by leg and intersection
Valhalla-based Directions API endpoints now return a new `admins` object on each [route leg object](https://docs.mapbox.com/api/navigation/#route-leg-object) and a new `admin_index` object on each [route step’s intersection object](https://docs.mapbox.com/api/navigation/#route-step-object).

`admins` should be decoded as `RouteLeg.administrativeRegions`, typed as an optional array of `AdministrativeRegion` structs. `admin_index` should be decoded as `RouteStep.administrativeRegionIndicesByIntersection`, typed as `[Int]?`.

We shouldn’t decode `admin_index` as part of the Intersection object, because the index of an object in an array is not an intrinsic aspect of that object, and an Intersection object can be reused outside the context of a particular step or leg. There are other examples like waypoint indices where we also have to diverge from the JSON response format because of this object-orientedness principle.

`RouteStep.administrativeRegionIndicesByIntersection` will be very awkward to use, and this level of indirection is completely unnecessary on Apple platforms, which standardize on ISO&nbsp;3166-1 alpha-2 codes for regions. For convenience, we can resolve these administrative regions at decoding time, adding an `Intersection.regionCode` property, typed as a two-character-long string, which developers can easily use in conjunction with `Locale.isoRegionCodes` and `Locale.localizedString(forRegionCode:)`. We’d still need to keep `RouteLeg.administrativeRegions` and `RouteStep.administrativeRegionIndicesByIntersection` around for round-tripping purposes, since the standard library doesn’t provide a way to map alpha-2 codes to alpha-3 codes, but documentation comments can steer developers toward the more convenient `Intersection.regionCode` property.

/cc @mapbox/navigation-ios @danpaz

Udumft:
Resolves #477 

Added properties `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection`. Made `Intersection.administrativeRegionIndex` internal.

I was trying to remove `Intersection.administrativeRegionIndex` completely and this is the solution I ended up with. Would love to see a feedback on that approach.
This solution introduces some internal manipulations in order to encode and decode `Routes` correctly, while preserving individual `Intersection` codability. The issue I am seeing there is that such behavior might be implicit for users, and mu confuse why same `Incident` gets encoded and decoded differently when it's a part of a `RouteStep` and when it's not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fixed an error that occurred when setting the `Waypoint.separatesLegs` property to `true` and setting the `Waypoint.targetCoordinate` property. ([#480](https://github.com/mapbox/mapbox-directions-swift/pull/480))
 * `Directions.fetchAvailableOfflineVersions(completionHandler:)` now calls its completion handler on the main queue consistently. ([#475](https://github.com/mapbox/mapbox-directions-swift/pull/475))
 * Upgraded to Polyline v5.0.0. ([#487](https://github.com/mapbox/mapbox-directions-swift/pull/487))
+* Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is now internal. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
 
 ## v1.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@
 * Fixed an error that occurred when setting the `Waypoint.separatesLegs` property to `true` and setting the `Waypoint.targetCoordinate` property. ([#480](https://github.com/mapbox/mapbox-directions-swift/pull/480))
 * `Directions.fetchAvailableOfflineVersions(completionHandler:)` now calls its completion handler on the main queue consistently. ([#475](https://github.com/mapbox/mapbox-directions-swift/pull/475))
 * Upgraded to Polyline v5.0.0. ([#487](https://github.com/mapbox/mapbox-directions-swift/pull/487))
-* Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is now internal. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
-* Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is removed in favor of the above, depending on needs in a string representation or index value. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
 
 ## v1.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 * Added support for building and running on any Linux distribution supported by Swift. ([#488](https://github.com/mapbox/mapbox-directions-swift/pull/488))
 * Added the `MapboxDirectionsCLI` command line tool that round-trips Mapbox Directions API responses between JSON format and Swift model objects. ([#469](https://github.com/mapbox/mapbox-directions-swift/pull/469))
 * The `CongestionLevel` enumeration now conforms to the `CaseIterable` protocol. ([#500](https://github.com/mapbox/mapbox-directions-swift/pull/500))
+* Refined encoding/decoding logic for `AdministrativeRegions` by `Leg` and `Intersection` ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485)). Added few properties for convenience access:
+    * `Intersection.regionCode` - A 2-letter region code to identify corresponding country that this intersection lies in.
+    * `RouteLeg.regionCode(atStepIndex:, intersectionIndex:)` - Returns the ISO 3166-1 alpha-2 region code for the administrative region through which the given intersection passes.
 
 ## v1.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * `Directions.fetchAvailableOfflineVersions(completionHandler:)` now calls its completion handler on the main queue consistently. ([#475](https://github.com/mapbox/mapbox-directions-swift/pull/475))
 * Upgraded to Polyline v5.0.0. ([#487](https://github.com/mapbox/mapbox-directions-swift/pull/487))
 * Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is now internal. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
+* Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is removed in favor of the above, depending on needs in a string representation or index value. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
 
 ## v1.0.0
 
--- a/Sources/MapboxDirections/AdministrativeRegion.swift
+++ b/Sources/MapboxDirections/AdministrativeRegion.swift
@@ -18,4 +18,18 @@ public struct AdministrativeRegion: Codable, Equatable {
         self.countryCode = countryCode
         self.countryCodeAlpha3 = countryCodeAlpha3
     }
+    
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        
+        countryCode = try container.decode(String.self, forKey: .countryCode)
+        countryCodeAlpha3 = try container.decode(String.self, forKey: .countryCodeAlpha3)
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        
+        try container.encode(countryCode, forKey: .countryCode)
+        try container.encode(countryCodeAlpha3, forKey: .countryCodeAlpha3)
+    }
 }
--- a/Sources/MapboxDirections/AdministrativeRegion.swift
+++ b/Sources/MapboxDirections/AdministrativeRegion.swift
@@ -23,13 +23,13 @@ public struct AdministrativeRegion: Codable, Equatable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         
         countryCode = try container.decode(String.self, forKey: .countryCode)
-        countryCodeAlpha3 = try container.decode(String.self, forKey: .countryCodeAlpha3)
+        countryCodeAlpha3 = try container.decodeIfPresent(String.self, forKey: .countryCodeAlpha3)
     }
     
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         
         try container.encode(countryCode, forKey: .countryCode)
-        try container.encode(countryCodeAlpha3, forKey: .countryCodeAlpha3)
+        try container.encodeIfPresent(countryCodeAlpha3, forKey: .countryCodeAlpha3)
     }
 }
--- a/Sources/MapboxDirections/AdministrativeRegion.swift
+++ b/Sources/MapboxDirections/AdministrativeRegion.swift
@@ -2,9 +2,9 @@ import Foundation
 
 /**
  :nodoc:
- `AdministrationRegion` describes corresponding object on the route.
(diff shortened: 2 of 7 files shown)


Sources/MapboxDirections/Intersection.swift
```diff
@@ -162,11 +165,33 @@ extension Intersection: Codable {
         case tunnelName = "tunnelName"
         case isUrban = "is_urban"
         case restStop = "rest_stop"
-        case administrationRegionIndex = "admin_index"
+        case administrativeRegionIndex = "admin_index"
         case geometryIndex = "geometry_index"
     }
     
+    static func encode(intersections: [Intersection], to parentContainer: inout UnkeyedEncodingContainer, administrativeRegionIndices: [Int?]?) throws {
```
Udumft, line 172: I am not particularly happy with this `inout` parameter, but this is how encoding array of objects is done. Passing `Encoder` here is not an option, as we need to call `container.superEncoder()` for each item we encode. Otherwise it just overwrites the JSON node.

Sources/MapboxDirections/AdministrativeRegion.swift
```diff
@@ -2,9 +2,9 @@ import Foundation
 
 /**
  :nodoc:
- `AdministrationRegion` describes corresponding object on the route.
+ `AdministrativeRegion` describes corresponding object on the route.
  */
-public struct AdministrationRegion: Codable, Equatable {
+public struct AdministrativeRegion: Codable, Equatable {
 
     private enum CodingKeys: String, CodingKey {
```
1ec5, line 9: While you’re at it, can you take a look at #491? I think the fix lies somewhere in here too.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 * Fixed an error that occurred when setting the `Waypoint.separatesLegs` property to `true` and setting the `Waypoint.targetCoordinate` property. ([#480](https://github.com/mapbox/mapbox-directions-swift/pull/480))
 * `Directions.fetchAvailableOfflineVersions(completionHandler:)` now calls its completion handler on the main queue consistently. ([#475](https://github.com/mapbox/mapbox-directions-swift/pull/475))
 * Upgraded to Polyline v5.0.0. ([#487](https://github.com/mapbox/mapbox-directions-swift/pull/487))
+* Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is now internal. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
+* Added `Intersection.regionCode` and `RouteStep.administrativeRegionIndicesByIntersection` for more convenient region data usage. `Intersection.administrativeRegionIndex` is removed in favor of the above, depending on needs in a string representation or index value. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))
```
1ec5, line 10: #483 effectively removed these properties from the public API (as far as documentation is concerned), and we unreleased v1.1.0, so there’s no need to talk about the APIs we briefly exposed prior to release.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -534,7 +541,35 @@ open class RouteStep: Codable {
         }
     }
     
-    public required init(from decoder: Decoder) throws {
+    static func decode(from decoder: Decoder, administrativeRegions: [AdministrativeRegion]) throws -> [RouteStep] {
```
1ec5, line 544: `init(from:)` is required for Decodable conformance. The administrative regions need to be passed in via the decoder’s `userInfo`, just like `routeOptions`.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -776,6 +828,13 @@ open class RouteStep: Codable {
     */
     public let intersections: [Intersection]?
     
+    /**
+     An array of `Administrative Regions` indicies for each `intersection` along current step.
+     
+     Array may be `nil` in case `intersections` data is not available. Array element may be `nil` if corresponding `intersection` has no `Administrative Region` assigned.
+     */
+    public let administrativeRegionIndicesByIntersection: [Int?]?
```
1ec5, line 836: Per https://github.com/mapbox/mapbox-directions-swift/issues/477#issue-732531128, even though it’s desirable to round-trip these indices, the RouteStep API should publicly expose a `regionCodesByIntersection` property of type `[String]`. There’s no need to explicitly give developers an option between alpha-2 and alpha-3 codes as the API does, though I suppose this method can remain in place as a more indirect option.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -274,7 +276,7 @@ open class RouteLeg: Codable {
      */
     open var expectedTravelTime: TimeInterval
 
-    open var administrationRegions: [AdministrationRegion]?
+    open var administrativeRegions: [AdministrativeRegion]?
```
1ec5, line 279: If we keep this property and `RouteStep.administrativeRegionIndicesByIntersection` around, then we need to document this property and also address #491.

Sources/MapboxDirections/RouteStep.swift
Udumft: `userInfo` is not mutable during decoding process, so admin regions should be configured by user, outside.
About updating entities after the actual decoding - I didn't want to create series of `mutating` methods, as `Intersection` is a struct.

1ec5: If we need to define this custom method, then we still need an explicit `init(from:)`; otherwise, Codable will synthesize one for us that may not match the expected structure.

Sources/MapboxDirections/AdministrativeRegion.swift
1ec5: #492 takes care of #491.

Sources/MapboxDirections/RouteStep.swift
Udumft: `RouteLeg.administrativeRegionIndicesByIntersection`, `RouteLeg.Intersections` and `Intersection.regionCode` are all constants. Is it possible to get them out of sync? Having `administrativeRegionIndicesByIntersection` as a computed property seem to become expensive getter.

1ec5: I guess I had been thinking these properties would need to be non-constant in order for the populate-after-decoding idea in https://github.com/mapbox/mapbox-directions-swift/pull/485#discussion_r519562906 to work. It’s still suboptimal for derived values to be stored explicitly, but you’re right that it’s even more suboptimal to make it a computed property over the whole step, since client code may only ever need one index at a time.

Mainly I was hoping to keep client code from having to subscript any array to find out the administrative region that contains a given intersection, but I guess it’s inevitable that indices will be part of this API no matter what. We could keep what you have so far, but add a `RouteLeg.regionCode(atStepIndex:intersectionIndex:)` method. At least that method will be performant by only looking for a single intersection, and developers could get the answer they’re looking for in one method call.

Sources/MapboxDirections/RouteStep.swift
Udumft: explicit `init(from:)` is still [available](https://github.com/mapbox/mapbox-directions-swift/pull/485/files#diff-6f755dc6c64528357f03e87f359b0f8c20d9c827b3cedf8ff27eec0c472542d0R568-R570) and both methods share decoding logic.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -575,7 +610,24 @@ open class RouteStep: Codable {
         typicalTravelTime = try container.decodeIfPresent(TimeInterval.self, forKey: .typicalTravelTime)
         
         transportType = try container.decode(TransportType.self, forKey: .transportType)
-        intersections = try container.decodeIfPresent([Intersection].self, forKey: .intersections)
+        
+        administrativeRegionIndicesByIntersection = try container.decodeIfPresent([AdministrativeAreaIndex].self,
+                                                                                  forKey: .intersections)?.map { $0.administrativeRegionIndex }
```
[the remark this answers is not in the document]
Udumft: Renamed.

Sources/MapboxDirections/RouteLeg.swift
Udumft: Done

Sources/MapboxDirections/RouteStep.swift
Udumft: Added the proposed convenience method

CHANGELOG.md
Udumft: Removed these lines

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -778,6 +829,15 @@ open class RouteStep: Codable {
     */
     public let intersections: [Intersection]?
     
+    /**
+     An array of `Administrative Regions` indicies for each `intersection` along current step.
+     
+     Array may be `nil` in case `intersections` data is not available. Array element may be `nil` if corresponding `intersection` has no `Administrative Region` assigned.
```
1ec5, line 835:
```suggestion
     Each intersection’s administrative region index.
     
     This property is set to `nil` if the `intersections` property is `nil`. An individual array element may be `nil` if the corresponding `Intersection` instance has no administrative region assigned.
```

Sources/MapboxDirections/Intersection.swift
```diff
@@ -120,14 +120,19 @@ public struct Intersection {
...
-
-     If the information is unavailable, this property is set to `nil`.
+     A 2-letter region code to identify corresponding country that this intersection lies in.
+     
+     Automatically populated during decoding a `RouteLeg` object, since this is the source of all `AdministrativeRegion`s. Value is `nil` if such information is unavailable.
+     
+     - seealso: `RouteStep.regionCode(atStepIndex:, intersectionIndex:)`
      */
-    public let administrationRegionIndex: Int?
+    public private(set) var regionCode: String?
```
1ec5, line 131: This getter is in the public API, so we can mention it in the changelog.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -249,6 +251,22 @@ open class RouteLeg: Codable {
         }
     }
     
+    /**
+     Return corresponding `iso_3166_1` encoded region code, if such information is available.
+     
+     - seealso: `Intersection.regionCode`
+     */
+    public func regionCode(atStepIndex stepIndex: Int, intersectionIndex: Int) -> String? {
```
1ec5, line 259: This method is in the public API, so we can mention it in the changelog too.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -249,6 +251,22 @@ open class RouteLeg: Codable {
         }
     }
     
+    /**
+     Return corresponding `iso_3166_1` encoded region code, if such information is available.
```
1ec5, line 255:
```suggestion
     Returns the ISO 3166-1 alpha-2 region code for the administrative region through which the given intersection passes. The intersection is identified by its step index and intersection index.
```

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -276,7 +294,10 @@ open class RouteLeg: Codable {
      */
     open var expectedTravelTime: TimeInterval
 
-    open var administrationRegions: [AdministrationRegion]?
+    /**
+     Containts a list of `AdministrativeRegion`'s on the current leg. `nil` value means that such data is not available.
```
1ec5, line 298: AdministrativeRegion is `:nodoc:`’d, so this property should also be hidden:

```suggestion
     :nodoc:
     The administrative regions through which the leg passes.
     
     This property is set to `nil` if no administrative region data is available.
```

Sources/MapboxDirections/Intersection.swift
```diff
@@ -120,14 +120,19 @@ public struct Intersection {
...
      */
     public let isUrban: Bool?
-
+    
     /**
-     :nodoc:
-     The index of the item in the `administrationRegions` array that corresponds to the country code of the country that this intersection lies in.
-
-     If the information is unavailable, this property is set to `nil`.
+     A 2-letter region code to identify corresponding country that this intersection lies in.
```
1ec5, line 125:
```suggestion
     The ISO 3166-1 alpha-2 code of the region where this intersection is located, if known.
```

1ec5 (approved):
Thanks for addressing the feedback above. This turned out a lot better than what we had in v1.1.0. The only remaining feedback is to copyedit the documentation. Suggested changelog entry:

* Added the `RouteLeg.regionCode(atStepIndex:intersectionIndex:)` method and `Intersection.regionCode` property to determine the administrative region where a specific intersection is located. ([#485](https://github.com/mapbox/mapbox-directions-swift/pull/485))

PR #486, opened 2020-11-04
Guard against new/unrecognized platforms and architectures when building a user agent string

gwynne:
This PR is an exact copy of the same change made to identical code found in `mapbox-speech-swift`. Please refer to mapbox/mapbox-speech-swift#41 for full details of the change and the reasoning behind it.


1ec5 (approved):
Thanks! I opened mapbox/mapbox-navigation-ios#2726, mapbox/MapboxGeocoder.swift#208, and mapbox/MapboxStatic.swift#108 to track making the same change in other Mapbox libraries.

PR #489, opened 2020-11-05
v1.1.0

1ec5:
Updated the Xcode project, CocoaPods podspec, and readme for v1.1.0.

/cc @mapbox/navigation-ios


MaximAlien (approved):
Overall looks good. Probably good idea would be to mention that `Polyline` was updated to `v5.0.0` in changelog.

1ec5:
#491 has the potential to break Directions API usage in multiple profiles, and we aren’t in a good position to fix the issue over the weekend, so I’ve deleted v1.1.0: CocoaPods/Specs@460fe24b9b402aa1980f838d844b649e8632e145. We’ll rerelease next week.

PR #490, opened 2020-11-06
Removing Intersection.geometryIndex

### LINKED ISSUE #478: Move geometry indices from intersections to route steps
#466 added an `Intersection.geometryIndex` property, which is problematic because the geometry is stored independently of the intersection, in `RouteStep.shape`. It isn’t currently possible to reuse an Intersection object on a different step, such as when creating a new route, because of this tight coupling.

`geometry_index` should be decoded as a `RouteStep.segmentRangesByIntersection` property, similar to `RouteLeg.segmentRangesByStep`.

/ref https://github.com/mapbox/mapbox-directions-swift/pull/466#pullrequestreview-519935443 #477
/cc @mapbox/navigation-ios @avi-c @danpaz

Udumft:
Resolves #478 

This PR utilizes same approach as in #477 (and is based on that branch).


Sources/MapboxDirections/RouteStep.swift
```diff
@@ -776,6 +840,20 @@ open class RouteStep: Codable {
...
+     An array of `Administrative Regions` indicies for each `intersection` along current step.
+     
+     Array may be `nil` in case `intersections` data is not available. Array element may be `nil` if corresponding `intersection` has no `Administrative Region` assigned.
+     */
+    public let administrativeRegionIndicesByIntersection: [Int?]?
+
+    /**
+     Segments indicies for each Intersection along the step.
+     
+     Ordered by the appearance. Array may be `nil` in case `intersections` data is not available. Array element may be `nil` if corresponding `intersection` has no `geometry index` data assigned.
```
1ec5, line 853:
```suggestion
     Segments indices for each `Intersection` along the step.
     
     The indices are arranged in the same order as the items of `intersections`. This property is `nil` if `intersections` is `nil`. An individual item may be `nil` if the corresponding JSON-formatted intersection object has no `geometry_index` property.
```

1ec5 (approved):

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -575,7 +619,27 @@ open class RouteStep: Codable {
...
+        
+        administrativeRegionIndicesByIntersection = try container.decodeIfPresent([AdministrativeAreaIndex].self,
+                                                                                  forKey: .intersections)?.map { $0.administrativeRegionIndex }
+        var rawIntersections = try container.decodeIfPresent([Intersection].self, forKey: .intersections)
+        
+        // Updating `Intersection.regionCode` since we removed it's `admin_index` for convenience
+        if let administrativeRegions = administrativeRegions,
+           rawIntersections != nil,
+           let rawAdminIndicies = administrativeRegionIndicesByIntersection {
+            for index in 0..<rawIntersections!.count {
```
MaximAlien, line 631: Can we prevent force-unwrapping usage here?

MaximAlien (approved):
I think `vk/477-admin-decode` should be rebased onto `main` to get the latest changes from there. There are also conflicts in current branch with base branch, which should be addressed before merge.

`segmentIndicesByIntersection` usage looks good to me.

Sources/MapboxDirections/RouteStep.swift
Udumft: We cannot use `let rawIntersections = rawIntersections` construct here because we need to mutate contents of this array later in the code.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -534,7 +543,42 @@ open class RouteStep: Codable {
...
+    private struct AdministrativeAreaIndex: Codable {
+        private enum CodingKeys: String, CodingKey {
+            case administrativeRegionIndex = "admin_index"
+        }
+        
+        var administrativeRegionIndex: Int?
+    }
+
+    /// Used to Decode `Intersection.geometry_index`
+    private struct IntersectionIndex: Codable {
```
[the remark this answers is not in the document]
Udumft: Since this struct is used only inside `RouteStep` for specific decoding scenario, I would prefer to have it `private`ly nested into `RouteStep`. But indeed some `shape` suffix could make it more descriptable.

MaximAlien (approved):

MaximAlien:
@Udumft, do you think it's ready to be merged into `main`?

Udumft:
This PR is based on #485 and could be merged right after that one is approved and merged.

PR #501, opened 2020-11-19
v1.2.0-alpha.1

chezzdev:
Version update and changelog fixture.


CHANGELOG.md
```diff
@@ -3,11 +3,11 @@
 ## v1.2.0
 
 * Added support for building and running on any Linux distribution supported by Swift. ([#488](https://github.com/mapbox/mapbox-directions-swift/pull/488))
+* Added the `MapboxDirectionsCLI` command line tool that round-trips Mapbox Directions API responses between JSON format and Swift model objects. ([#469](https://github.com/mapbox/mapbox-directions-swift/pull/469))
 
 ## v1.1.0
 
 * Added the `DirectionsResult.typicalTravelTime`, `RouteLeg.typicalTravelTime` and `RouteStep.typicalTravelTime` properties that indicate the typical travel time, as opposed to the current expected travel time. ([#462](https://github.com/mapbox/mapbox-directions-swift/pull/462))
-* Added the `MapboxDirectionsCLI` command line tool that round-trips Mapbox Directions API responses between JSON format and Swift model objects. ([#469](https://github.com/mapbox/mapbox-directions-swift/pull/469))
```
1ec5, line 10: Thanks, good catch!

README.md
```diff
@@ -27,7 +27,7 @@ Or in your [CocoaPods](http://cocoapods.org/) Podfile:
 # Latest stable release
 pod 'MapboxDirections', '~> 1.1'
 # Latest prerelease
-pod 'MapboxDirections', :git => 'https://github.com/mapbox/mapbox-directions-swift.git', :tag => 'v1.1.0'
+pod 'MapboxDirections', :git => 'https://github.com/mapbox/mapbox-directions-swift.git', :tag => 'v1.2.0-alpha.1'
```
1ec5, line 30: Looks like the SPM prerelease line below didn’t get updated. This find-and-replace only looks for the `.exact(…)` syntax:

https://github.com/mapbox/mapbox-directions-swift/blob/8d421c760d3e9bc7bfb6a3a54409b3ca5c98fbe1/scripts/update-version.sh#L43

But we’re using `from:` with SPM because it can respect tadpole-operator semantics even with prerelease tags. So we’d want to replace the second occurrence of `from:` to have the prerelease tag.

1ec5 (approved):
Looks good, but we’ll have to manually fix the readme and/or the release script for SPM installation instructions.

1ec5 (approved):
You’re a better `sed` wizard than I, thanks!

PR #503, opened 2020-11-23
Linux SPM Build error

### LINKED ISSUE #498: Linux builds intermittently fail without explanation
As seen in #497, [Linux builds intermittently fail without any logs](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/847/workflows/77e1584a-0ff2-4e98-8efc-18ad678c4b02/jobs/2535), sometimes even after repeated retries. But SSH’ing into the build machine and manually running `swift build && swift test` works without a hitch.

/cc @mapbox/navigation-ios @frederoni

Udumft:
Resolves #498 
I tried to find the core reason of the issue, but did not find any information. Apparently there is some threading bug in compiling, but I have no means to prove it.

What I've found during the investigation:
* Resetting build cache (via `swift package reset`) causes next `swift build` to fail)
* The usual build runs correctly if started for the second time (via `ssh`)
* Console output and error outputs are clean
* Neither `--verbose` nor `--sanitize` options provide more information for debugging

All the above points towards compiler bug, and current solution is just a workaround for the issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
       - image: swift:5.3
     steps:
       - checkout
-      - run: swift build
+      - run: swift build --jobs=1
       - run: swift test
 
   spm-job:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,7 @@ jobs:
       - image: swift:5.3
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


chezzdev (approved):
Maybe the fix deserves comment for future generations :)

PR #506, opened 2020-12-02
Route Alerts Refinement

### LINKED ISSUE #482: Refine route alert types and properties for public consumption
This issue tracks tail work regarding the API for route alerts added in #466, based on https://github.com/mapbox/mapbox-directions-swift/pull/466#pullrequestreview-521913167:

* [x] Rename `AdministrationRegion` to `AdministrativeRegion`
* [x] Clarify whether RestStop represents a rest stop or its type
* [x] Document `RouteLeg.incidents`
* [x] Make Incident properties public and document them
* [x] Make `TollCollection.type` public and document it
* [x] Make `RestStop.type` public and document it
* [x] Change Incident timestamp properties to `Date`s and rename them
* [x] Change `Incident.type` and `subtype` to enumerations (?)
* [x] Replace `Incident.geometryIndexStart` and `geometryIndexEnd` with `shapeIndexRange`

Other tail work is tracked in #477 and #478.

Udumft:
Prepares Route Alerts related types for public usage.

This PR refines Docs for number of entities and properties, modifies `Incident` struct. For the latter I used info provided by Directions team in [slack convo](https://mapbox.slack.com/archives/C03KW1JQD/p1606915892095200) and existing [Valhalla schema](https://github.com/mapbox/valhalla-internal/blob/master/proto/incidents.proto#L59).

Fixes #482.


Sources/MapboxDirections/AdministrativeRegion.swift
```diff
@@ -1,7 +1,6 @@
 import Foundation
 
 /**
- :nodoc:
  `AdministrativeRegion` describes corresponding object on the route.
```
1ec5, line 4: If we publicly document this type, then the documentation should probably mention the `Intersection.regionCode` property that is of the simpler String type and that this type is primarily used for encoding and decoding purposes.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,144 @@ public struct Incident: Codable, Equatable {
...
-    var subtypeDescription: String?
-    var alertCodes: [Int]
-    var lanesBlocked: [Int]
-    var geometryIndexStart: Int
-    var geometryIndexEnd: Int
+    
+    /// Defines known types of incidents.
+    ///
+    /// Each incident may or may not have specific set of data, depending on it's `type`
+    public enum IncidentType: String {
```
1ec5, line 27: The full name of this type would be `Incident.IncidentType`, which is redundant. Swift already takes up the `Type` name, so we could call it `Kind`, for consistency with `VisualInstruction.Component.Kind`.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,144 @@ public struct Incident: Codable, Equatable {
...
+    public var type: IncidentType
+    var rawType: String
+    /// Short description of an incident. May be used as an additional info.
+    public var description: String
+    /// Date when incident item was created. Uses ISO8601 format
+    public var creationDate: Date
+    /// Date when incident happened. Uses ISO8601 format
+    public var startDate: Date
+    /// Date when incident shall end. Uses ISO8601 format
+    public var endDate: Date
```
1ec5, line 56: Now that these properties are of type Date, they don’t follow any string format.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,144 @@ public struct Incident: Codable, Equatable {
...
+    /// Provides additional classification of an incident. May be not available for all incident types.
+    public var subtype: String?
+    /// Breif description of the subtype. May be not available for all incident types and is not available if `subtype` is `nil`
+    public var subtypeDescription: String?
+    /// Contains list of ISO 14819-2:2013 codes
+    ///
+    /// See https://www.iso.org/standard/59231.html for details
+    public var alertCodes: [Int]
+    /// A list of lane indices, affected by the incident
+    public var lanesBlocked: [Int]
```
1ec5, line 68: Are these indices consistent with the indices in `Intersection.approachLanes`? If so, let’s mention it in this documentation comment.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,144 @@ public struct Incident: Codable, Equatable {
...
+        case LaneRestriction = "lane_restriction"
+        case MassTransit = "mass_transit"
+        case Miscellaneous = "miscellaneous"
+        case OtherNews = "other_news"
+        case PlannedEvent = "planned_event"
+        case RoadClosure = "road_closure"
+        case RoadHazard = "road_hazard"
+        case Weather = "weather"
+        
+        case Unknown = "<<unknown type>>"
```
1ec5, line 41: Perhaps it would be more correct to represent an unknown kind as a `nil` value for the `kind` property. It would be up to the encoding and decoding code to fall back to this `<<unknown type>>` string, however.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,144 @@ public struct Incident: Codable, Equatable {
...
+    /// Shows severity of an incident. May be not available for all incident types.
+    public var impact: String?
+    /// Provides additional classification of an incident. May be not available for all incident types.
+    public var subtype: String?
+    /// Breif description of the subtype. May be not available for all incident types and is not available if `subtype` is `nil`
+    public var subtypeDescription: String?
+    /// Contains list of ISO 14819-2:2013 codes
+    ///
+    /// See https://www.iso.org/standard/59231.html for details
+    public var alertCodes: [Int]
```
1ec5, line 66: Does the order matter? If not, maybe it could be represented by a Set. (JSON doesn’t have a native set type, hence the use of an array.)

Sources/MapboxDirections/RestStop.swift
```diff
@@ -11,7 +10,10 @@ public struct RestStop: Codable, Equatable {
         case restArea = "rest_area"
     }
 
-    let type: StopType
+    /**
+     The kind of the rest stop.
+     */
+    public let type: StopType
```
1ec5, line 16: Similarly, `RestStop.StopType` is redundant. Let’s use `Kind` whenever possible.

1ec5 (commented):
If you have time, it would be great to include a basic encoding/decoding unit test to ensure that the Incident type and its dependent types all round-trip in the usual case. I realize the tests were missing from #466, but any (justified) departure from the JSON format increases the risk of a roundtripping issue.

Sources/MapboxDirections/Incident.swift
Udumft: According to the [schema](https://github.com/mapbox/live-traffic-service-interface/blob/master/service.proto#L126-L127), these are not consistent with approach [lanes](https://docs.mapbox.com/api/navigation/directions/#lane-object).

Sources/MapboxDirections/AdministrativeRegion.swift
Udumft: Added mention to the doc

Sources/MapboxDirections/Incident.swift
Udumft: Removed

Sources/MapboxDirections/Incident.swift
Udumft: 'Kind' sound a bit odd to my ear, but if there are already such cases I assume that is reasonable to use.

Sources/MapboxDirections/Incident.swift
Udumft: In the end I decided to introduce an `enum` to ease handling these. It is unlikely that humanity invents new types of lanes, but I still left a place to avoid crashing once naming is updated. Even NN [does not trust](https://github.com/mapbox/mapbox-navigation-native/blob/9ae67c67660688d833601cbe59bbb7ece9bc15a4/bindings/navigator.mbg#L201) this list to be static.
Unfortunately this is not compatible with `LaneIndication` we use in `Intersection`s.

Sources/MapboxDirections/Incident.swift
Udumft: In fact, that even simplified the Coding part! Updated code doc to explain `nil` value.

Sources/MapboxDirections/Incident.swift
Udumft: Done

Sources/MapboxDirections/Incident.swift
Udumft: Or am I overthinking it introducing `rawLanesBlocked`?

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,174 @@ public struct Incident: Codable, Equatable {
...
-    var geometryIndexEnd: Int
+    
+    /// Defines known types of incidents.
+    ///
+    /// Each incident may or may not have specific set of data, depending on it's `kind`
+    public enum Kind: String {
+        case Accident = "accident"
+        case Congestion = "congestion"
+        case Construction = "construction"
+        case DisabledVehicle = "disabled_vehicle"
```
1ec5, line 31: jazzy is currently configured not to show any undocumented symbol, so although these kind enumerations will show up, none of their cases will. It’s a lot to document, so if you need to just document each one pro forma (such as “Disabled vehicle” here), that’s fine for now.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,174 @@ public struct Incident: Codable, Equatable {
...
+    /// Each `Incident` may have arbitrary of affected lanes
+    public enum BlockedLane: String, Codable {
+        case Left = "LEFT"
+        case LeftCenter = "LEFT CENTER"
+        case LeftTurnLane = "LEFT TURN LANE"
+        case Center = "CENTER"
+        case Right = "RIGHT"
+        case RightCenter = "RIGHT CENTER"
+        case RightTurnLane = "RIGHT TURN LANE"
+        case Hov = "HOV"
```
1ec5, line 53: “HOV” is an initialism for “high-occupancy vehicle”. I think we should write it out because it’s a somewhat specialized term that especially non-native English speakers may be unfamiliar with.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,174 @@ public struct Incident: Codable, Equatable {
...
-    var alertCodes: [Int]
-    var lanesBlocked: [Int]
-    var geometryIndexStart: Int
-    var geometryIndexEnd: Int
+    
+    /// Defines known types of incidents.
+    ///
+    /// Each incident may or may not have specific set of data, depending on it's `kind`
+    public enum Kind: String {
+        case Accident = "accident"
```
1ec5, line 28: An enumeration case should begin with a lowercase letter.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,174 @@ public struct Incident: Codable, Equatable {
...
+        case Lane1 = "1"
+        case Lane2 = "2"
+        case Lane3 = "3"
+        case Lane4 = "4"
+        case Lane5 = "5"
+        case Lane6 = "6"
+        case Lane7 = "7"
+        case Lane8 = "8"
+        case Lane9 = "9"
+        case Lane10 = "10"
```
1ec5, line 66: For the documentation comments for these cases, let’s say e.g. “Second lane from the left”, assuming that’s what `lane2` means even in a region that drives on the left.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,174 @@ public struct Incident: Codable, Equatable {
...
+        let geometryIndexEnd = try container.decode(Int.self, forKey: .geometryIndexEnd)
+        shapeIndexRange = geometryIndexStart..<geometryIndexEnd
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        let formatter = ISO8601DateFormatter()
+        
+        try container.encode(identifier, forKey: .identifier)
+        try container.encode(rawKind, forKey: .type)
```
1ec5, line 180: Do we need a separate `rawKind` property? I think we can make `Kind` conform to Codable and it’ll automatically be possible to replace `rawKind` with `kind` here.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,174 @@ public struct Incident: Codable, Equatable {
...
+        lanesBlocked = rawLanesBlocked.reduce(into: Set<BlockedLane?>()) { $0.insert(BlockedLane(rawValue: $1)) }
+        
+        let geometryIndexStart = try container.decode(Int.self, forKey: .geometryIndexStart)
+        let geometryIndexEnd = try container.decode(Int.self, forKey: .geometryIndexEnd)
+        shapeIndexRange = geometryIndexStart..<geometryIndexEnd
+    }
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        let formatter = ISO8601DateFormatter()
```
1ec5, line 177: ISO8601DateFormatter outputs a date in RFC&nbsp;3339 format. Would MapboxNavigationNative be OK with that format, or does it need to be a different ISO&nbsp;8601 format?

/ref mapbox/mapbox-navigation-ios#2327

Sources/MapboxDirections/Incident.swift
Udumft: Doesn't "Lane1" means the leftmost lane for Left Driving Side regions and the rightmost lane for Right Driving Side regions?

Sources/MapboxDirections/Incident.swift
Udumft: Oh, I wanted to avoid introducing custom extension, but didn't know about such issue.

Sources/MapboxDirections/Incident.swift
Udumft: `rawKind` is introduced to preserve JSON encoding/decoding roundtrip for cases when current SDK version does not support new types of `Incident`s. Thus, we just store the raw value for coding purposes, and at the same time user is enabled to work only with known defined `Kind`s.

Same approach applied to `BlockedLane`, but IMHO it is less likely that new types of lanes could be introduced then new types of incidents. So I would rather agree that we should get rid of `rawLanesBlocked` instead.

Sources/MapboxDirections/Incident.swift
Udumft: `Incident` dates are originally supplied in `yyyy-MM-dd'T'HH:mm:ssZ` format and NN takes that while setting a Route, so I don't think there's a reason to change it.

Sources/MapboxDirections/Incident.swift
1ec5: I was assuming the numbering follows the California standard, but other regions may number the lanes differently, so maybe we need to be ambiguous for now.

Sources/MapboxDirections/Incident.swift
1ec5: To avoid parallel properties, we should store only the raw value and make the enumeration-typed property a computed property.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,215 @@ public struct Incident: Codable, Equatable {
...
+    /// Breif description of the subtype. May be not available for all incident types and is not available if `subtype` is `nil`
+    public var subtypeDescription: String?
+    /// Contains list of ISO 14819-2:2013 codes
+    ///
+    /// See https://www.iso.org/standard/59231.html for details
+    public var alertCodes: Set<Int>
+    /// A list of lanes indices, affected by the incident
+    ///
+    /// `nil` value indicates that such lane identifier is not supported
+    public var lanesBlocked: Set<BlockedLane?>
```
1ec5, line 138: As tail work, maybe we should consider turning BlockedLane into an OptionSet to more seamlessly accommodate multiple values.

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,215 @@ public struct Incident: Codable, Equatable {
...
+    }
+    
+    /// Defines a lane affected by the `Incident`
+    ///
+    /// Each `Incident` may have arbitrary of affected lanes
+    public enum BlockedLane: String, Codable {
+        /// Left lane
+        case left = "LEFT"
+        /// Left and center lanes
+        case leftCenter = "LEFT CENTER"
```
1ec5, line 49: According to [this presentation](https://slideplayer.com/slide/9372700/) (slide 11), traffic information management (TIM) systems in the U.S. use “left center lane” to refer to the second lane from left on a four-lane highway:



Note that this slide isn’t globally applicable: for example, inside and outside would be reversed in everyday speech in Taiwan. But I don’t know if that would affect the incident data we get.

Sources/MapboxDirections/Incident.swift
1ec5: The National Traffic Information Management Coalition, which is affiliated with AASHTO in the U.S., has been [recommending](https://www.michigan.gov/documents/mdot/13-NTIMC_Lane_Designation_Terminology_517666_7.pdf) the California lane numbering scheme for use nationwide.

Sources/MapboxDirections/BlockedLanes.swift
```diff
@@ -0,0 +1,116 @@
+
+import Foundation
+
+/// Defines a lane affected by the `Incident`
+public struct BlockedLanes: OptionSet, CustomStringConvertible {
```
1ec5, line 5: Oh cool, thanks for taking care of turning this type into an option set.

Sources/MapboxDirections/Incident.swift
```diff
@@ -25,94 +25,28 @@ public struct Incident: Codable, Equatable {
...
+        case construction = "construction"
+        case disabledVehicle = "disabled_vehicle"
+        case laneRestriction = "lane_restriction"
+        case massTransit = "mass_transit"
+        case miscellaneous = "miscellaneous"
+        case otherNews = "other_news"
+        case plannedEvent = "planned_event"
+        case roadClosure = "road_closure"
+        case roadHazard = "road_hazard"
+        case weather = "weather"
```
1ec5, line 39: These enumeration values should have at least pro-forma documentation comments so they appear in the docset.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -281,13 +281,19 @@ open class RouteLeg: Codable {
     open var expectedTravelTime: TimeInterval
 
     /**
-     :nodoc:
      The administrative regions through which the leg passes.
-          
+     
+     Items are ordered by appearance, most recent one is at the beginning.
      This property is set to `nil` if no administrative region data is available.
      */
```
1ec5, line 288: Since we’re making this type public, its documentation should refer to the `Intersection.regionCode` property as an alternative.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
     * `Intersection.regionCode` - A 2-letter region code to identify corresponding country that this intersection lies in.
     * `RouteLeg.regionCode(atStepIndex:, intersectionIndex:)` - Returns the ISO 3166-1 alpha-2 region code for the administrative region through which the given intersection passes.
 * Added `RouteStep.segmentIndicesByIntersection` for navigating `Intersection`s segments along the step;  ([#490](https://github.com/mapbox/mapbox-directions-swift/pull/490))
+* Added support for encoding and decoding `Route Alerts` data coming from `Directions API`. Includes `TollCollection`, `RestStop` and `AdministrationRegion`. Also added `Incident` struct for reflecting incoming incidents along the route ([#466](https://github.com/mapbox/mapbox-directions-swift/pull/466), [#506](https://github.com/mapbox/mapbox-directions-swift/pull/506))
```
1ec5, line 12: Let’s focus on the new getters so developers can find out how to make use of the new types:

```suggestion
* Added the `RouteLeg.incidents` property that indicates known traffic incidents, toll collection points, rest areas, and border crossings along the route leg. ([#466](https://github.com/mapbox/mapbox-directions-swift/pull/466), [#506](https://github.com/mapbox/mapbox-directions-swift/pull/506))
* Added the `RouteLeg.administrativeRegions` property that indicates the administrative regions traversed by the route leg. ([#466](https://github.com/mapbox/mapbox-directions-swift/pull/466), [#506](https://github.com/mapbox/mapbox-directions-swift/pull/506))
* Added the `Intersection.tunnelName`, `Intersection.tollCollection, `Intersection.restStop`, and `Intersection.isUrban` properties. ([#466](https://github.com/mapbox/mapbox-directions-swift/pull/466), [#506](https://github.com/mapbox/mapbox-directions-swift/pull/506))
```

1ec5 (approved):

Sources/MapboxDirections/Incident.swift
```diff
@@ -21,46 +20,156 @@ public struct Incident: Codable, Equatable {
...
+    
+    /// Incident identifier
+    public var identifier: String
+    /// The kind of an incident
+    ///
+    /// This value is set to `nil` if `kind` value is not supported.
+    public var kind: Kind? {
+        return Kind(rawValue: rawKind)
+    }
+    var rawKind: String
```
1ec5, line 62: Something for consideration before the final release: now that we’re writing in pure Swift, we could consolidate these two properties into a single `kind` property of type `Kind`, but one of the `Kind` cases could have an associated value with the raw type. This would in fact be an option for any enumeration that the Directions API reserves the right to extend, other than option sets.

1ec5 (approved):

PR #507, opened 2020-12-07
Detailed Road Classification

Udumft:
Adds support for Mapbox Streets V8 road classes.


Sources/MapboxDirections/MapboxStreetClass.swift
```diff
@@ -0,0 +1,65 @@
+
+import Foundation
+
+public enum MapboxStreetClass: String, Codable {
```
Udumft, line 4: Unlike `RoadClasses`, Mapbox Streets V8 assigns a single class to an intersection. Also, possible list of Mapbox road classes is not compatible with exiting `RoadClasses`, so I decided to put it as a separate `enum`.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -160,12 +168,21 @@ extension Intersection: Codable {
...
     }
     
+    /// Used to code `Intersection.streetClass`
+    private struct MapboxStreetClassCodable: Codable {
+        private enum CodingKeys: String, CodingKey {
+            case streetClass = "class"
+        }
+        
+        let streetClass: MapboxStreetClass?
+    }
```
Udumft, line 185: Introduced this container struct for coding purposes. Since there is not other data available under `mapbox_streets_v8` object, I see this as most effective solution. If in the future more data will be stored under this key - it will be trivial to extract this into separate entity

Tests/MapboxDirectionsTests/IntersectionTests.swift
```diff
@@ -16,6 +16,9 @@ class IntersectionTests: XCTestCase {
                 "bearings": [80],
                 "location": [13.426579, 52.508068],
                 "classes": ["toll", "restricted"],
+                "mapbox_streets_v8": [
```
MaximAlien, line 19: Is it possible that there will be multiple street classes? Would it be possible to add test for such case?

Sources/MapboxDirections/MapboxStreetClass.swift
MaximAlien, line 4: Can you also add conformance to `CaseIterable` protocol? Sometimes it might be useful when generating random street class.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
     * `Intersection.regionCode` - A 2-letter region code to identify corresponding country that this intersection lies in.
     * `RouteLeg.regionCode(atStepIndex:, intersectionIndex:)` - Returns the ISO 3166-1 alpha-2 region code for the administrative region through which the given intersection passes.
 * Added `RouteStep.segmentIndicesByIntersection` for navigating `Intersection`s segments along the step;  ([#490](https://github.com/mapbox/mapbox-directions-swift/pull/490))
+* Added Mapbox Streets V8 Road Classification info to `Intersection.roadClass` property. See [Mapbox Streets V8 Road classification doc](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#road) for more info on the classes ([#507](https://github.com/mapbox/mapbox-directions-swift/pull/507)).
```
MaximAlien, line 12: Should it be `Intersection.streetClass` instead of `Intersection.roadClass`?

Sources/MapboxDirections/MapboxStreetClass.swift
1ec5: Yes, that also makes sense because you can’t request a route that only traverses secondary roads, for example.

Tests/MapboxDirectionsTests/IntersectionTests.swift
1ec5: No, these are classes in the sense of road classification, so they’re mutually exclusive. If Mapbox Streets were to assign multiple classes to a single roadway, it would be a breaking change for Streets anyways.

Sources/MapboxDirections/MapboxStreetClass.swift
1ec5: I can see how that would be helpful in some cases, but there isn’t an inherent ordering to all these classifications. For example, a motorway is larger than a trunk road, but is an aerialway larger than a service road?

Sources/MapboxDirections/Intersection.swift
1ec5: I’m pretty sure a new road class value would wait until a Mapbox Streets source v9, because it would be pretty disruptive to existing styles.

Sources/MapboxDirections/MapboxStreetClass.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// (point Widened section at the end of a cul-de-sac for turning around a vehicle
+    case TurningCircle = "turning_circle"
+    /// (point Similar to a turning circle but with an island or other obstruction at the centerpoint
+    case TurningLoop = "turning_loop"
+    /// (point Lights or other signal controlling traffic flow at an intersection
+    case TrafficSignals = "traffic_signals"
+    /// (point A point indication for road junctions
+    case Junction = "junction"
+    /// (point Indicating the class and type of roads meeting at an intersection. Intersections are only available in Japan
+    case Intersection = "intersection"
```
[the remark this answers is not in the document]
Udumft: Directions team [suggests](https://mapbox.slack.com/archives/C03KW1JQD/p1607432389199300) to still have these values even though not all of them can really be present... So maybe we should leave it just to be safe?

CHANGELOG.md
Udumft: My bad

Sources/MapboxDirections/MapboxStreetClass.swift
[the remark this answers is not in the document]
Udumft: Done

Sources/MapboxDirections/MapboxStreetClass.swift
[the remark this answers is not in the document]
1ec5: I’m inclined to say [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it), since we could always add more cases to this non-frozen enumeration without breaking backwards compatibility, and it’s difficult to imagine a situation where the Directions API would start adding these values to edges without breaking backwards compatibility in other ways.

If we really want to be on the safe side, I guess we could leave it in but keep it undocumented. But even then, the downside is that developers still have to deal with these dead code paths whenever they deal with Streets road classes. Swift will complain about these cases being missing from any switch statement unless the developer adds a `default @unknown` clause.

Sources/MapboxDirections/MapboxStreetClass.swift
```diff
@@ -1,65 +1,68 @@
 
 import Foundation
 
+/// 'Mapbox Streets V8' compliant street class definition
+///
+/// For detailed info on road classes, see [the documentation](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#road)
```
1ec5, line 6: Nit: for concision:

```suggestion
/// A road classification according to the [Mapbox Streets source](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#road), version 8.
```

Sources/MapboxDirections/MapboxStreetClass.swift
```diff
@@ -1,65 +1,68 @@
...
 public enum MapboxStreetClass: String, Codable {
```
1ec5, line 7: Sorry I didn’t catch this earlier, but it should be MapboxStreetsRoadClass, as wordy as that is. Even the existing RoadClass is Mapbox-specific; what’s new about this enumeration is that it’s based on the Mapbox Streets source. Alternatively, if that sounds too wordy, we can call it DetailedRoadClass. Just make sure to be consistent throughout the PR.

Sources/MapboxDirections/MapboxStreetClass.swift
```diff
@@ -1,65 +1,68 @@
...
     /// Ski lifts, gondolas, and other types of aerialway.
-    case Aerialway = "aerialway"
+    case aerialway = "aerialway"
     /// The approximate centerline of a golf course hole
-    case Golf = "golf"
+    case golf = "golf"
     /// Circular continuous-flow intersection
-    case Roundabout = "roundabout"
+    case roundabout = "roundabout"
     /// Smaller variation of a roundabout with no center island or obstacle
```
1ec5, line 56: A mini roundabout is also a point feature; the Streets source documentation doesn’t say so but that’s an omission.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -111,6 +113,12 @@ public struct Intersection {
      */
     public let restStop: RestStop?
 
+    /**
+     Street class according to Mapbox Streets V8 classification.
+     
+     This value is set to `nil` of such info is not available.
+     */
+    public let streetClass: MapboxStreetClass?
```
1ec5, line 121: For consistency, move this property up next to `outletRoadClasses` and rename it to `outletDetailedRoadClass` or `outletMapboxStreetsRoadClass`, depending on the final name for the type.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
     * `Intersection.regionCode` - A 2-letter region code to identify corresponding country that this intersection lies in.
     * `RouteLeg.regionCode(atStepIndex:, intersectionIndex:)` - Returns the ISO 3166-1 alpha-2 region code for the administrative region through which the given intersection passes.
 * Added `RouteStep.segmentIndicesByIntersection` for navigating `Intersection`s segments along the step;  ([#490](https://github.com/mapbox/mapbox-directions-swift/pull/490))
+* Added Mapbox Streets V8 Road Classification info to `Intersection.streetClass` property. See [Mapbox Streets V8 Road classification doc](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#road) for more info on the classes ([#507](https://github.com/mapbox/mapbox-directions-swift/pull/507)).
```
1ec5, line 12: > * Added the `Intersection.outletMapboxStreetsRoadClass` property that indicates a more detailed road classification than the existing `Intersection.outletRoadClasses` property. ([#507](https://github.com/mapbox/mapbox-directions-swift/pull/507)).

Sources/MapboxDirections/Intersection.swift
```diff
@@ -111,6 +113,12 @@ public struct Intersection {
      */
     public let restStop: RestStop?
 
+    /**
+     Street class according to Mapbox Streets V8 classification.
+     
+     This value is set to `nil` of such info is not available.
+     */
```
1ec5, line 120:
```suggestion
    /**
     The road classes of the road that the containing step uses to leave the intersection, according to the [Mapbox Streets source](https://docs.mapbox.com/vector-tiles/reference/mapbox-streets-v8/#road), version 8.
     
     If detailed road class information is unavailable, this property is set to `nil`. This property only indicates the road classification; for other aspects of the road, use the `outletRoadClasses` property.
     */
```

And in the `outletRoadClasses` documentation, add a cross-reference like this:

> For more detailed road classifications, use the `outletMapboxStreetsRoadClass` property.

1ec5 (commented):
Some additional documentation nits:

Sources/MapboxDirections/MapboxStreetClass.swift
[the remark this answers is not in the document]
Udumft: According to the [source code](https://github.com/mapbox/valhalla-internal/blob/865440ba9ac47bb96a8e24790a1aa3109bb2594c/src/tyr/mapbox_streets.cc), there are more cases which will never make it to the real world navigation: `pedestrian` (this one is surprising, why not?) , `construction`, `track`, `path`, `"major_rail"`, `minor_rail`, `service_rail`, `aerialway`, `golf`.
Maybe we should also not add them for now?

Sources/MapboxDirections/MapboxStreetsRoadClass.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// Railways, including mainline, commuter rail, and rapid transit.
+    case majorRail = "major_rail"
+    /// Includes light rail & tram lines.
+    case minorRail = "minor_rail"
+    /// Yard and service railways.
+    case serviceRail = "service_rail"
+    /// Ski lifts, gondolas, and other types of aerialway.
+    case aerialway = "aerialway"
+    /// The approximate centerline of a golf course hole
+    case golf = "golf"
```
1ec5, line 51: In case you’re wondering how these classes would ever come up: railroads are sometimes used as driveways in rural areas, and theoretically a ski lift could be part of a pedestrian route.

1ec5 (approved):
Thanks for iterating on this PR a few times; it looks great.

Udumft:
I've added a small change that will allow to safely ignore unrecognized road classes. This will allow backwards compatibility with potential future additions to the road classes list without crashing the app.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -156,12 +165,36 @@ extension Intersection: Codable {
...
+            self.streetClass = streetClass
+        }
+        
+        init(from decoder: Decoder) throws {
+            let container = try decoder.container(keyedBy: CodingKeys.self)
+            
+            if let classString = try container.decodeIfPresent(String.self, forKey: .streetClass) {
+                streetClass = MapboxStreetsRoadClass(rawValue: classString)
+            } else {
+                streetClass = nil
```
1ec5, line 193: I think this is fine because adding new road classes is a rare event and the new value wouldn’t be immediately useful in the navigation SDK anyways. However, if we ever learn of a plan to add new road classes in Streets source v8._x_, we should preemptively implement the `other` case with an associated value, similar to what’s proposed in https://github.com/mapbox/mapbox-directions-swift/pull/506#discussion_r540281957. Not necessary just yet though.

1ec5 (approved):

PR #517, opened 2021-02-09
Turf v2.0.0-alpha.1

1ec5:
This PR requires Turf v2.0.0-alpha.1 (soon to be v2.0.0-alpha.2) as a stopgap to unblock mapbox/mapbox-navigation-ios#2629 and mapbox/mapbox-navigation-ios#2794. This PR will be merged to the release-v2.0 branch, which will not be merged into the main branch until we eliminate the need for a binary Turf distribution (mapbox/turf-swift#133) when using SPM.

/cc @mapbox/navigation-ios @macdrevx


1ec5 (commented):
v2.0.0-alpha.2 has been tagged. We need to upgrade to that tag, which has the correct build flags. We should also update the Cartfile and podspecs for consistency across dependency managers, especially since v2.0.0-alpha.2 has some unrelated type changes that haven’t been released as part of v1.3.0 yet.

1ec5:
[Linux](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/937/workflows/26d89cca-fa32-422f-8b98-e8418fb5da83/jobs/2845) builds are failing:

```
error: artifact of binary target 'Turf' failed extraction: The operation could not be completed. (TSCBasic.Process.Error error 0.)
```

Per https://github.com/mapbox/turf-swift/pull/133#discussion_r573134112, we’ll need to either exclude tvOS, watchOS, and Linux from this branch of MapboxDirections or conditionally have those platforms depend on a different tag of MapboxDirections.

zugaldia:
> we’ll need to either exclude tvOS, watchOS, and Linux

Sounds like a reasonable trade-off for the Public Preview, and one that we should reconsider before GA.

MaximAlien (approved):

zugaldia (approved):

PR #519, opened 2021-02-18
Turn Incident.Impact type into an enumeration

dersim-davaod:
Resolves: https://github.com/mapbox/mapbox-directions-swift/issues/518


1ec5 (approved):
Suggested changelog entry:

* The `Incident.impact` property is now an `Incident.Impact` value instead of a string. ([#519](https://github.com/mapbox/mapbox-directions-swift/pull/519))

PR #522, opened 2021-02-26
Add fallback access token.

MaximAlien:
Add fallback access token and use `MBXAccessToken` as default access token.

### DIFF
--- a/Sources/MapboxDirections/DirectionsCredentials.swift
+++ b/Sources/MapboxDirections/DirectionsCredentials.swift
@@ -1,7 +1,8 @@
 import Foundation
 
 /// The Mapbox access token specified in the main application bundle’s Info.plist.
-let defaultAccessToken = Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String
+let defaultAccessToken = Bundle.main.object(forInfoDictionaryKey: "MBXAccessToken") as? String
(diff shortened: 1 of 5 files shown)


Sources/MapboxDirections/DirectionsCredentials.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
 
 /// The Mapbox access token specified in the main application bundle’s Info.plist.
-let defaultAccessToken = Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String
+let defaultAccessToken = Bundle.main.object(forInfoDictionaryKey: "MBXAccessToken") as? String
+let fallbackAccessToken = Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String
```
1ec5, line 5: Consolidate this into a single variable with a `??` chain, so we can be sure no future code will check `MBXAccessToken` without also checking `MGLMapboxAccessToken`.

1ec5 (approved):
Suggested changelog entry:

* To gain access to the Mapbox Directions and Map Matching APIs, set `MBXAccessToken` in your Info.plist. `MGLMapboxAccessToken` is still supported but is now deprecated. ([#522](https://github.com/mapbox/mapbox-directions-swift/pull/522))

Sources/MapboxDirections/DirectionsCredentials.swift
MaximAlien: Done.

PR #524, opened 2021-03-03
 Prepare Mapbox Directions v2.0.0-alpha.2.

MaximAlien:
Bump Mapbox Directions version to v2.0.0-alpha.2.


1ec5 (approved):
This is good to go once #525 lands. FYI, for alpha 1, I copied MapboxDirections.podspec to MapboxDirections-pre.podspec, changed the pod name to `MapboxDirections-pre`, and pushed that ad-hoc pod to trunk.

PR #530, opened 2021-03-25
Support tests that use OHHTTPStubs in Swift Package Manager

### LINKED ISSUE #494: Mocked tests don’t run under Swift Package Manager
Swift Package Manager skips over several test suites that account for most of the unit tests in this repository, on all platforms, not just Linux. These tests have all been conditionally uncompiled with `!SWIFT_PACKAGE` because linksmt/OHHTTPStubs@563f48d3fab84ef04639649c770b00f4fa502cca doesn’t support Linux or SPM. We’re using this commit off a fork of OHHTTPStubs for AliSoftware/OHHTTPStubs#286, which introduces watchOS support, even though we don’t have a watchOS test target. Meanwhile, OHHTTPStubs v9._x_ does support both Linux and SPM.

This issue affects only `swift test` on the command line, not testing within Xcode, so the primary impact is to continuous integration.

mapbox/MapboxStatic.swift#109 shows that would be straightforward to upgrade to OHHTTPStubs v9._x_, but that’ll require the fix in apple/swift-package-manager#2817, which is in Xcode 12.2 (which is still a release candidate) or an equivalent Swift 5.3 image for Linux.

Here are the affected test suites:

* AnnotationTests
* DirectionsCredentialsTests
* DirectionsTests
* Most of MatchTests
* OfflineDirectionsTests
* RoutableMatchTest
* Most of RouteRefreshTests
* V5Tests
* VisualInstructionsTests

/cc @mapbox/navigation-ios

azarovalex:
Fixes #494.

Looks like OHHTTPStubs don't have Linux support because OHHTTPStubs uses objc/runtime to swizzle URLSesson methods [here](https://github.com/AliSoftware/OHHTTPStubs/blob/e37b05e0ea4916611c08babe7c68211c4de52a38/Sources/OHHTTPStubs/HTTPStubs%2BNSURLSessionConfiguration.m).

OHHTTPStubs on macOS through SMP works fine, there is a minor problem with module names – when we import OHHTTPStubs through Carthage, the name of the module is `OHHTTPStubs`, and when through SMP, the name is `OHHTTPStubsSwift`.  Even though the content of the modules is the same. Currently, I'm solving this issue with a conditional compilation like this:
```swift
import OHHTTPStubs
#if SWIFT_PACKAGE
import OHHTTPStubsSwift
#endif
```
In theory, we can create our own empty package that will use `@_exported` statements and then include this everywhere, but this API is private.

Also, we need to replace every bundle creating with `Bundle.module`.
If you have any suggestions on how to run these tests on Linux or how to simplify import statements, feel free to add a comment.


Package.swift
```diff
@@ -33,7 +42,7 @@ let package = Package(
             exclude: ["Info.plist"]),
         .testTarget(
             name: "MapboxDirectionsTests",
-            dependencies: ["MapboxDirections"],
+            dependencies: ["MapboxDirections"] + optionalTestTargetDependecies,
```
1ec5, line 45: If you specify a [target](https://developer.apple.com/documentation/swift_packages/target/dependency/3583309-target) explicitly, then its condition can be [dependent on the platform](https://developer.apple.com/documentation/swift_packages/targetdependencycondition/3554520-when) without resorting to conditional compilation.

1ec5 (approved):
The changes look pretty reasonable to me, including the conditional compilation. The `Bundle.module` changes are fine because, elsewhere in the test target, we declare that constant conditionally for Carthage. I just have one comment about a more straightforward way to declare the conditional OHHTTPStubs dependency, but it isn’t critical.

Package.swift
azarovalex: thx, fixed this one

PR #532, opened 2021-04-05
Fix VisualInstruction.component.lane type mismatch

jill-cardamon:
This PR fixes a type mismatch in the encoding/decoding of the `VisualInstruction.Component.lane.preferredDirection` enumeration case as well as the relevant tests.

### DIFF
--- a/Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
+++ b/Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
@@ -69,7 +69,7 @@ class VisualInstructionComponentTests: XCTestCase {
             "type": "lane",
             "active": true,
             "directions": ["right", "straight"],
-            "active_direction": ["right"],
+            "active_direction": "right",
         ]
         let componentData = try! JSONSerialization.data(withJSONObject: componentJSON, options: [])
         var component: VisualInstruction.Component?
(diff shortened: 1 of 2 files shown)


Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
```diff
@@ -69,7 +69,7 @@ class VisualInstructionComponentTests: XCTestCase {
             "type": "lane",
             "active": true,
             "directions": ["right", "straight"],
-            "active_direction": ["right"],
+            "active_direction": "right",
```
1ec5, line 72: This is the kind of error that we often catch through the integration tests that begin with raw JSON fixtures. It’s unfortunate that those fixtures are difficult to maintain though. For now, it’s a reminder to us to double-check these Swift-formatted fixtures against the actual responses coming out of the Directions API.

1ec5 (approved):

Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
jill-cardamon: Yeah. I wonder if there's a reliable way to use an actual response and just parse for what needs to be tested.

1ec5: It is possible, but we’ve chosen not to do that so far, because a dependency on a live network connection can cause a CI job to be unreliable. It would be more appropriate for a CI job that runs on a nightly schedule instead of on every commit or push.

PR #529, opened 2021-03-24
Intersections indicate preferred lane usage

jill-cardamon:
### Description
This PR adds new Directions API response properties that indicate preferred lane usage #516.
- [x] Add new Lane object properties
- [x] Add new Instruction component property
- [x] Update tests for lane object propeties
- [x] Update tests for instruction component property
- [x] Update CHANGELOG.md

### Implementation
For the banner instruction component object property `active_direction`: added `preferredDirection` argument (of type `LaneIndication?`) associated value of the `VisualInstruction.Component.lane(indications:isUsable:)` enumeration case.

For the lane object properties `active` and `valid_indication`: Added `Intersection.preferredApproachLanes` property (of type `IndexSet?`) modeled after `Intersection.usableApproachLanes`, added `Intersection.usableLaneIndication` property (of type `LaneIndication?`).

PR #535, opened 2021-04-12
Preferred lane direction is now a ManeuverDirection

jill-cardamon:
### Description
This PR is to change the type of preferred lane direction from `LaneIndication` to `ManeuverDirection` to remove the hacks introduced in #529 and #532 and avoid converting within the Nav SDK.
- [x] Update property type for preferred lane direction propeties
- [x] Update tests

### DIFF
--- a/Sources/MapboxDirections/Intersection.swift
+++ b/Sources/MapboxDirections/Intersection.swift
@@ -267,11 +267,7 @@ extension Intersection: Codable {
             lanes = approachLanes.map { Lane(indications: $0) }
             for i in usableApproachLanes {
                 lanes![i].isValid = true
-//                if lanes![i].indications.contains(usableLaneIndication){
-//                    lanes![i].validIndication = usableLaneIndication
-//                }
-//                let usable = LaneIndication(descriptions: [usableLaneIndication.rawValue])
-                if lanes![i].indications.contains(LaneIndication(descriptions: [usableLaneIndication.rawValue])!) {
+                if lanes![i].indications.descriptions.contains(usableLaneIndication.rawValue) {
                     lanes![i].validIndication = usableLaneIndication
                 }
             }
@@ -323,9 +319,11 @@ extension Intersection: Codable {
             approachLanes = lanes.map { $0.indications }
             usableApproachLanes = lanes.indices { $0.isValid }
             preferredApproachLanes = lanes.indices { ($0.isActive ?? false) }
-            usableLaneIndication = lanes.compactMap { $0.validIndication }[0]
-//            let usableIndications = lanes.compactMap { $0.validIndication }
-//            usableLaneIndication = usableIndications.reduce(ManeuverDirection(rawValue: "none")) { $0.union($1) }
+            if Set(lanes.compactMap { $0.validIndication}).count > 1 {
+                let context = EncodingError.Context(codingPath: decoder.codingPath, debugDescription: "Inconsistent valid indications.")
+                throw EncodingError.invalidValue(Set(lanes.compactMap { $0.validIndication}).count, context)
+            }
+            usableLaneIndication = lanes.compactMap { $0.validIndication }.first ?? ManeuverDirection.none
         } else {
             approachLanes = nil
             usableApproachLanes = nil
--- a/Sources/MapboxDirections/Intersection.swift
+++ b/Sources/MapboxDirections/Intersection.swift
@@ -325,7 +325,7 @@ extension Intersection: Codable {
                 let context = EncodingError.Context(codingPath: decoder.codingPath, debugDescription: "Inconsistent valid indications.")
                 throw EncodingError.invalidValue(validIndications, context)
             }
-            usableLaneIndication = lanes.compactMap { $0.validIndication }.first
+            usableLaneIndication = validIndications.first
         } else {
             approachLanes = nil
             usableApproachLanes = nil
--- a/Sources/MapboxDirections/Intersection.swift
+++ b/Sources/MapboxDirections/Intersection.swift
@@ -264,7 +264,6 @@ extension Intersection: Codable {
         if let approachLanes = approachLanes,
             let usableApproachLanes = usableApproachLanes,
             let preferredApproachLanes = preferredApproachLanes
-//            let usableLaneIndication = usableLaneIndication
         {
             lanes = approachLanes.map { Lane(indications: $0) }
             for i in usableApproachLanes {
@@ -326,10 +325,7 @@ extension Intersection: Codable {
                 let context = EncodingError.Context(codingPath: decoder.codingPath, debugDescription: "Inconsistent valid indications.")
                 throw EncodingError.invalidValue(validIndications, context)
             }
-//            let usableIndications = lanes.compactMap { $0.validIndication }
-//            usableLaneIndication = usableIndications.reduce(LaneIndication(rawValue: 0)) { $0.union($1) }
-//            let blah = validIndications.reduce(ManeuverDirection(rawValue: "none")) { $0 }
-            usableLaneIndication = lanes.compactMap { $0.validIndication }.first ?? ManeuverDirection(rawValue: "none")
+            usableLaneIndication = lanes.compactMap { $0.validIndication }.first
         } else {
             approachLanes = nil
             usableApproachLanes = nil
--- a/Sources/MapboxDirections/Intersection.swift
+++ b/Sources/MapboxDirections/Intersection.swift
@@ -15,7 +15,7 @@ public struct Intersection {
                 approachLanes: [LaneIndication]?,
                 usableApproachLanes: IndexSet?,
                 preferredApproachLanes: IndexSet?,
-                usableLaneIndication: LaneIndication?,
+                usableLaneIndication: ManeuverDirection?,
                 outletRoadClasses: RoadClasses? = nil,
                 tollCollection: TollCollection? = nil,
                 tunnelName: String? = nil,
@@ -165,7 +165,7 @@ public struct Intersection {
      
      If no lane information is available for the intersection, this property’s value is `nil`
      */
-    public let usableLaneIndication: LaneIndication?
+    public let usableLaneIndication: ManeuverDirection?
 }
 
 extension Intersection: Codable {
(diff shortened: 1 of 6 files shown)


jill-cardamon:
Tests are failing because of a nil value error.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -267,7 +267,11 @@ extension Intersection: Codable {
             lanes = approachLanes.map { Lane(indications: $0) }
             for i in usableApproachLanes {
                 lanes![i].isValid = true
-                if lanes![i].indications.contains(usableLaneIndication){
+//                if lanes![i].indications.contains(usableLaneIndication){
+//                    lanes![i].validIndication = usableLaneIndication
+//                }
+//                let usable = LaneIndication(descriptions: [usableLaneIndication.rawValue])
```
1ec5, line 273: Remember to remove this commented-out code before merging.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -267,7 +267,11 @@ extension Intersection: Codable {
...
+                if lanes![i].indications.contains(LaneIndication(descriptions: [usableLaneIndication.rawValue])!) {
```
1ec5, line 274: This is implicitly reinterpreting a maneuver direction description string as a lane indication description string. That is what we want, but we should make this step more explicit. Let’s create a convenience initializer on `LaneIndication` that takes a `ManeuverDirection`. It can switch over all the anticipated description strings to create the corresponding `LaneIndication`, or it can use the same `rawValue`-based approach here, but at least it would be better documented that way.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -319,8 +323,9 @@ extension Intersection: Codable {
             approachLanes = lanes.map { $0.indications }
             usableApproachLanes = lanes.indices { $0.isValid }
             preferredApproachLanes = lanes.indices { ($0.isActive ?? false) }
-            let usableIndications = lanes.compactMap { $0.validIndication }
-            usableLaneIndication = usableIndications.reduce(LaneIndication(rawValue: 0)) { $0.union($1) }
+            usableLaneIndication = lanes.compactMap { $0.validIndication }[0]
+//            let usableIndications = lanes.compactMap { $0.validIndication }
+//            usableLaneIndication = usableIndications.reduce(ManeuverDirection(rawValue: "none")) { $0.union($1) }
```
1ec5, line 328: This introduces a hard assumption that the valid indications of each lane will always match. We’re OK with that assumption, but if the Directions API were to ever violate the assumption, perhaps we should throw `EncodingError.invalidValue(_:_:)` instead of silently swallowing the inconsistency. To check for an inconsistency:

```swift
Set(lanes.compactMap { $0.validIndication }).count > 1
```

Sources/MapboxDirections/Intersection.swift
```diff
@@ -319,8 +323,9 @@ extension Intersection: Codable {
             approachLanes = lanes.map { $0.indications }
             usableApproachLanes = lanes.indices { $0.isValid }
             preferredApproachLanes = lanes.indices { ($0.isActive ?? false) }
-            let usableIndications = lanes.compactMap { $0.validIndication }
-            usableLaneIndication = usableIndications.reduce(LaneIndication(rawValue: 0)) { $0.union($1) }
+            usableLaneIndication = lanes.compactMap { $0.validIndication }[0]
```
1ec5, line 326: Some tests are [crashing](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/997/workflows/ed18f821-8a90-402f-85cc-c1a89f5f70b6/jobs/3045) because of indexing into an empty array:

```
Fatal error: Index out of range: file Swift/ContiguousArrayBuffer.swift, line 444
2021-04-12 14:40:33.761822+0000 xctest[3358:53526] Fatal error: Index out of range: file Swift/ContiguousArrayBuffer.swift, line 444
```

Use `Array.first` to avoid the crash.

Sources/MapboxDirections/Intersection.swift
jill-cardamon: That's a good idea! I'll add this check

Sources/MapboxDirections/Intersection.swift
jill-cardamon: A better way to compare without the convenience initializer is to compare the string descriptions of the `LaneIndication` to the `ManeuverDirection`'s raw value:
`lanes![i].indications.descriptions.contains(usableLaneIndication.rawValue)`. If this comparison is used, is the convenience initializer necessary?

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -263,6 +263,11 @@ public enum ManeuverDirection: String, Codable {
      Use the difference between the step’s initial and final headings to distinguish between a U-turn to the left (typical in countries that drive on the right) and a U-turn on the right (typical in countries that drive on the left). If the difference in headings is greater than 180 degrees, the maneuver requires a U-turn to the left. If the difference in headings is less than 180 degrees, the maneuver requires a U-turn to the right.
      */
     case uTurn = "uturn"
+    
+    /**
+     The maneuver has no direction.
+     */
+    case none
```
jill-cardamon, line 270: Adding this case was necessary for getting "road-tripping" to work properly since there is not initializer for `ManeuverDirection` that handles `none`.

Sources/MapboxDirections/Intersection.swift
1ec5: I think it would be preferable to make the conversion explicit in some fashion. Otherwise it would be more difficult to find this raw value–based conversion if the assumption ever breaks in the future.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -323,9 +319,11 @@ extension Intersection: Codable {
             approachLanes = lanes.map { $0.indications }
             usableApproachLanes = lanes.indices { $0.isValid }
             preferredApproachLanes = lanes.indices { ($0.isActive ?? false) }
-            usableLaneIndication = lanes.compactMap { $0.validIndication }[0]
-//            let usableIndications = lanes.compactMap { $0.validIndication }
-//            usableLaneIndication = usableIndications.reduce(ManeuverDirection(rawValue: "none")) { $0.union($1) }
+            if Set(lanes.compactMap { $0.validIndication}).count > 1 {
+                let context = EncodingError.Context(codingPath: decoder.codingPath, debugDescription: "Inconsistent valid indications.")
+                throw EncodingError.invalidValue(Set(lanes.compactMap { $0.validIndication}).count, context)
```
1ec5, line 324: Returning the set itself would be more informative than the count of unique indications.

Sources/MapboxDirections/RouteStep.swift
1ec5: Let’s make the `usableLaneIndication` property optional instead, since `Optional` already has a `none` case.

jill-cardamon:
> Let’s make the `usableLaneIndication` property optional instead, since Optional already has a none case.

`usableLaneIndication` is an optional already. However, lanes would not encode during the V5 `testEncoding()` test when `usableLaneIndication` was a nil value.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -319,8 +320,12 @@ extension Intersection: Codable {
...
             usableApproachLanes = lanes.indices { $0.isValid }
             preferredApproachLanes = lanes.indices { ($0.isActive ?? false) }
-            let usableIndications = lanes.compactMap { $0.validIndication }
-            usableLaneIndication = usableIndications.reduce(LaneIndication(rawValue: 0)) { $0.union($1) }
+            let validIndications = lanes.compactMap { $0.validIndication}
+            if Set(validIndications).count > 1 {
+                let context = EncodingError.Context(codingPath: decoder.codingPath, debugDescription: "Inconsistent valid indications.")
+                throw EncodingError.invalidValue(validIndications, context)
+            }
+            usableLaneIndication = lanes.compactMap { $0.validIndication }.first
```
1ec5, line 328:
```suggestion
            usableLaneIndication = validIndications.first
```

Sources/MapboxDirections/LaneIndication.swift
```diff
@@ -69,6 +69,10 @@ public struct LaneIndication: OptionSet, CustomStringConvertible {
         self.init(rawValue: laneIndication.rawValue)
     }
     
+    public init?(from direction: ManeuverDirection) {
+        self.init(descriptions: [direction.rawValue])
```
1ec5, line 73: Add a comment noting the assumption that every possible raw value of ManeuverDirection matches a valid raw value of LaneIndication.

Sources/MapboxDirections/LaneIndication.swift
```diff
@@ -69,6 +69,10 @@ public struct LaneIndication: OptionSet, CustomStringConvertible {
         self.init(rawValue: laneIndication.rawValue)
     }
     
+    public init?(from direction: ManeuverDirection) {
```
1ec5, line 72: Does this method need to be public? If so, add a documentation comment for it.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -217,10 +217,11 @@ extension VisualInstruction.Component: Codable {
...
             let indications = try container.decode(LaneIndication.self, forKey: .directions)
             let isUsable = try container.decode(Bool.self, forKey: .isActive)
-            var preferredDirection: LaneIndication? = nil
-            if let preferredDirectionDescription = try container.decodeIfPresent(String.self, forKey: .activeDirection) {
-                preferredDirection = LaneIndication(descriptions: [preferredDirectionDescription])
-            }
+//            var preferredDirection: ManeuverDirection? = nil
+//            if let preferredDirectionDescription = try container.decodeIfPresent(ManeuverDirection.self, forKey: .activeDirection) {
+//                preferredDirection = ManeuverDirection(rawValue: preferredDirectionDescription)
+//            }
```
1ec5, line 223: Remove this commented-out code.

1ec5 (approved):

Sources/MapboxDirections/Intersection.swift
```diff
@@ -260,15 +260,16 @@ extension Intersection: Codable {
         try container.encode(outletArray, forKey: .outletIndexes)
         
         var lanes: [Lane]?
+        print("lanes: \(String(describing: lanes))")
```
bamx23, line 263: @jill-cardamon This print should be removed

PR #547, opened 2021-05-31
Upgrade SPM version to 5.4 for Linux builds

### LINKED ISSUE #539: Enable automatic test discovery on Linux
As of apple/swift-package-manager#3053 and [Swift 5.4](https://swift.org/blog/swift-5-4-released/), SPM now supports automatic test discovery on Linux. LinuxMain.swift is now deprecated. We should remove that file and XCTestManifests.swift to take advantage of automatic test discovery, which will eliminate the need to maintain a list of tests by hand.

https://github.com/mapbox/mapbox-directions-swift/blob/2ffc5f0cf9cc5a912d6d2187de854931ed79ab7e/CONTRIBUTING.md#L57

We’ll need to wait until Swift 5.4 makes it into [an Xcode image on CircleCI](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions) before we can depend on these changes:

https://github.com/mapbox/mapbox-directions-swift/blob/2ffc5f0cf9cc5a912d6d2187de854931ed79ab7e/.circleci/config.yml#L42

/cc @mapbox/navigation-ios

S2Ler:
Fixes #539.

SPM 5.4 has automatic test discovery support, so that we can remove
manually generated files for running Linux tests.


S2Ler:
@1ec5 Do you know why tests are not running for this PR?

1ec5 (approved):
🙌

PR #550, opened 2021-06-04
Adds means of injecting accessToken from SPM tests

S2Ler:
SPM tests can't contain Info.plist which is used for injecting default
access token into the library.

With this addition SPM tests can set `injectedAccessToken` variable
which will be used in `defaultAccessToken` initialization.

**Motivation:**
`MapboxNavigation` uses `Directions.shared` in many places and there is no currently a way to set access token for this shared instance from SPM tests.

### DIFF
--- a/Sources/MapboxDirections/DirectionsCredentials.swift
+++ b/Sources/MapboxDirections/DirectionsCredentials.swift
@@ -1,9 +1,10 @@
 import Foundation
 
 /// The Mapbox access token specified in the main application bundle’s Info.plist.
-let defaultAccessToken =
+let defaultAccessToken: String? =
     Bundle.main.object(forInfoDictionaryKey: "MBXAccessToken") as? String ??
-    Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String
+    Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String ??
+    UserDefaults.standard.string(forKey: "MBXAccessToken")
 let defaultApiEndPointURLString = Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAPIBaseURL") as? String
 
 public struct DirectionsCredentials: Equatable {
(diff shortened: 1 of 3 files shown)


Sources/MapboxDirections/DirectionsCredentials.swift
```diff
@@ -1,11 +1,15 @@
...
-let defaultAccessToken =
+let defaultAccessToken: String? =
     Bundle.main.object(forInfoDictionaryKey: "MBXAccessToken") as? String ??
-    Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String
+    Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAccessToken") as? String ??
+    injectedAccessToken
 let defaultApiEndPointURLString = Bundle.main.object(forInfoDictionaryKey: "MGLMapboxAPIBaseURL") as? String
 
+/// Allows SPM tests to inject access token which is used for `Directions.shared`
+var injectedAccessToken: String?
```
1ec5, line 11: I assume this is for https://github.com/mapbox/mapbox-navigation-ios/pull/3029#discussion_r643777358? An `MBXAccessToken` key in UserDefaults, as described in https://github.com/mapbox/mapbox-navigation-ios/pull/3029#discussion_r643642729, would be preferable to a global variable. But this global variable is clearly named and documented, so I’ll leave it up to you as to whether to migrate it to UserDefaults.

1ec5 (approved):

Sources/MapboxDirections/DirectionsCredentials.swift
S2Ler: I've been thinking about the other means of injecting `accessToken` and didn't want to introduce a public API just for one place without a discussion. And the work here is needed just for tests right now. Shall we maybe discuss "MBXAccessToken" in UserDefaults for all libraries? Or it is already implemented in some of the SDKs? I haven't found the implementation in mapbox-navigation-ios repo.

1ec5: The `MBXAccessToken` user defaults key is undocumented, because we only needed to introduce that extra level of flexibility for tests. However, there is some precedent for coupling recognized Info.plist keys with recognized user defaults keys. We could document this mechanism if there’s enough demand for it, but for now, I think it would be reasonable to keep it undocumented but make use of it in MapboxNavigationTests, with the understanding that if it ever broke, the test would fail.

Issue #553, opened 2021-06-17
RouteOptions.alleyPriority, walkwayPriority, speed should be optional


1ec5:
`RouteOptions.alleyPriority`, `walkwayPriority`, and `speed` should be optional, and `RouteOptions.urlQueryItems` should add the corresponding URL query items if and only if these properties are non-nil.

## Problem

RouteOptions only adds the `alley_bias` parameter to the request URL if the profile is either `mapbox/driving` or `mapbox/walking`, but not for other profiles such as `mapbox/driving-traffic` that also support this parameter.

https://github.com/mapbox/mapbox-directions-swift/blob/2d9edcedef9506546fa390c8f7f10de848b2aec2/Sources/MapboxDirections/RouteOptions.swift#L191-L193

## Background

These options are limited to Valhalla-powered profiles and are unsupported in OSRM-based profiles, such as `mapbox/cycling` and formerly `mapbox/driving-traffic`. OSRM-based endpoints return an HTTP 422 error upon encountering a parameter that’s supported by Valhalla but not OSRM.

Originally, in #370, this library had to conditionalize the `alley_bias` parameter on the `mapbox/walking` profile because it’s supported by Valhalla but not OSRM. #416 later extended the parameter to the `mapbox/driving` profile once it migrated to Valhalla. We could conditionalize other parameters on the presence or absence of an explicit value, but `RouteOptions.alleyPriority` had to be non-optional because optional primitive types don’t bridge to Objective-C. Fortunately, #382 dropped Objective-C support; since then, we generally use optional types to represent incremental additions to the Directions and Map Matching APIs, even for primitive types.

## Solution

`RouteOptions.alleyPriority` should become optional, because this library still supports at least one profile (`mapbox/cycling`) that doesn’t support `alley_bias`. `RouteOptions.urlQueryItems` can thus conditionalize the addition of `alley_bias` on whether `RouteOptions.alleyPriority` is non-nil.

The `walkway_bias` and `walking_speed` parameters added in #370 are in the same situation, so `RouteOptions.walkwayPriority` and `speed` should get the same treatment. It would be pretty unlikely for a driving profile to support options specific to walkways, but `DirectionsProfileIdentifier` is an extensible enumeration and there could always be a walking-related profile other than `mapbox/walking`.

/cc @mapbox/navigation-ios @danpat @OttyLab

1ec5:
[...]

PR #557, opened 2021-06-30
RouteOptions.alleyPriority, walkwayPriority, speed are now optional

jill-cardamon:
This PR addresses #553 by making `RouteOptions.alleyPriority`, `walkwayPriority`, and `speed` optional properties and conditionally adding their corresponding URL query items to `RouteOptions.urlQueryItems`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Carthage v0.38 or above is now required for installing this SDK if you use Carthage. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
 * Xcode 12.0 or above is now required to build MapboxDirections from source. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
 * You can fully build this SDK on Macs with Apple Silicon. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
-* `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` are now optional, and thus must be set explicitly if the developer wants to include them in the route options. `DirectionsOptions.default` has been renamed to `DirectionsOptions.medium`. ([#557](https://github.com/mapbox/mapbox-directions-swift/pull/557))
+* `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` are now optional. Set them explicitly if you want to include them in the HTTP request. Renamed `DirectionsOptions.default` to `DirectionsOptions.medium`. ([#557](https://github.com/mapbox/mapbox-directions-swift/pull/557))
 
 ## v1.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Carthage v0.38 or above is now required for installing this SDK if you use Carthage. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
 * Xcode 12.0 or above is now required to build MapboxDirections from source. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
 * You can fully build this SDK on Macs with Apple Silicon. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
+* `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` are now optional, and thus must be set explicitly if the developer wants to include them in the route options. `DirectionsOptions.default` has been renamed to `DirectionsOptions.medium`. ([#557](https://github.com/mapbox/mapbox-directions-swift/pull/557))
 
 ## v1.2.0
 
--- a/Sources/MapboxDirections/DirectionsOptions.swift
+++ b/Sources/MapboxDirections/DirectionsOptions.swift
@@ -99,6 +99,7 @@ public struct DirectionsPriority: Hashable, RawRepresentable {
     /**
      The priority level with which a route neither avoids nor prefers a particular type of roadway or pathway.
      */
+    // should we remove this default (or change its naming) if we don't set it as such?
     static let `default` = DirectionsPriority(rawValue: 0.0)
     
     /**
--- a/Sources/MapboxDirections/DirectionsOptions.swift
+++ b/Sources/MapboxDirections/DirectionsOptions.swift
@@ -99,8 +99,7 @@ public struct DirectionsPriority: Hashable, RawRepresentable {
     /**
      The priority level with which a route neither avoids nor prefers a particular type of roadway or pathway.
      */
-    // should we remove this default (or change its naming) if we don't set it as such?
-    static let `default` = DirectionsPriority(rawValue: 0.0)
+    static let medium = DirectionsPriority(rawValue: 0.0)
     
     /**
      The priority level with which a route prefers a particular type of roadway or pathway.
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -187,16 +187,16 @@ open class RouteOptions: DirectionsOptions {
         if includesExitRoundaboutManeuver {
             params.append(URLQueryItem(name: "roundabout_exits", value: String(includesExitRoundaboutManeuver)))
         }
-        if alleyPriority != nil {
-            params.append(URLQueryItem(name: "alley_bias", value: String(alleyPriority.rawValue)))
+        if let alleyPriority = alleyPriority?.rawValue {
+            params.append(URLQueryItem(name: "alley_bias", value: String(alleyPriority)))
         }
         
-        if walkwayPriority != nil {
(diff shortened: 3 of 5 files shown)


jill-cardamon:
Tests are failing because some `DirectionsTests.swift` tests weren't updated.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -129,7 +129,7 @@ open class RouteOptions: DirectionsOptions {
      
      The value of this property must be at least `DirectionsPriority.low` and at most `DirectionsPriority.high`. The default value of `DirectionsPriority.default` neither prefers nor avoids alleys, while a negative value between `DirectionsPriority.low` and `DirectionsPriority.default` avoids alleys, and a positive value between `DirectionsPriority.default` and `DirectionsPriority.high` prefers alleys. A value of 0.9 is suitable for pedestrians who are comfortable with walking down alleys.
```
S2Ler, line 130: The comment needs an update. There is no "default" anymore. Also, it needs an explanation of what 'nil' means.

S2Ler (commented):
Do we also need update examples and some docs after this PR?

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -99,6 +99,7 @@ public struct DirectionsPriority: Hashable, RawRepresentable {
     /**
      The priority level with which a route neither avoids nor prefers a particular type of roadway or pathway.
      */
+    // should we remove this default (or change its naming) if we don't set it as such?
```
1ec5, line 102: Yes, `medium` (as in “medium priority”) would be better, since apparently the API reserves the right to use a different default at some point in the future, even if that would be somewhat nonsensical for a relative value.

Sources/MapboxDirections/RouteOptions.swift
1ec5: Same for the other properties too.

jill-cardamon: Thank you both! I will update the docs. Some examples can be updated after the next release. I can open a ticket for tail work.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Carthage v0.38 or above is now required for installing this SDK if you use Carthage. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
 * Xcode 12.0 or above is now required to build MapboxDirections from source. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
 * You can fully build this SDK on Macs with Apple Silicon. ([#548](https://github.com/mapbox/mapbox-directions-swift/pull/548))
+* `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` are now optional, and thus must be set explicitly if the developer wants to include them in the route options. `DirectionsOptions.default` has been renamed to `DirectionsOptions.medium`. ([#557](https://github.com/mapbox/mapbox-directions-swift/pull/557))
```
1ec5, line 13: Nit: The changelog and release notes are written in the second person with the developer as the audience:

```suggestion
* `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` are now optional. Set them explicitly if you want to include them in the HTTP request. Renamed `DirectionsOptions.default` to `DirectionsOptions.medium`. ([#557](https://github.com/mapbox/mapbox-directions-swift/pull/557))
```

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -123,31 +123,31 @@ open class RouteOptions: DirectionsOptions {
...
     /**
-     The expected uniform travel speed measured in meters per second.
+     An optional that represents the expected uniform travel speed measured in meters per second.
      
-     This property has no effect unless the profile identifier is set to `DirectionsProfileIdentifier.walking`. You can adjust this property to account for running or for faster or slower gaits. When the profile identifier is set to another profile identifier, such as `DirectionsProfileIdentifier.driving`, this property is ignored in favor of the expected travel speed on each road along the route. This property may be supported by other routing profiles in the future.
+     This property has no effect unless it is explicitly set as it defaults to nil, and works best when the profile identifier is set to `DirectionsProfileIdentifier.walking`. You can adjust this property to account for running or for faster or slower gaits. When the profile identifier is set to another profile identifier, such as `DirectionsProfileIdentifier.driving`, this property is ignored in favor of the expected travel speed on each road along the route. This property may be supported by other routing profiles in the future.
```
1ec5, line 146: We named the property just `speed` on the client side in case there’s ever something like a `cycling_speed`, to avoid a proliferation of profile-specific properties on the class, but for now it’s only hooked up to `walking_speed`.

Although this PR technically allows the property to be set in other profiles, I think we should continue to document it as a walking-only option for now. We don’t want developers to get the impression that it works with any profile and get surprised about an error. If for some reason the API were to support `walking_speed` for, say, the `mapbox/driving-traffic` profile, that usage would remain possible but undocumented, which I think is OK.

/cc @danpat

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -123,31 +123,31 @@ open class RouteOptions: DirectionsOptions {
...
+     The value of this property must be at least `DirectionsPriority.low` and at most `DirectionsPriority.high`. `DirectionsPriority.medium` neither prefers nor avoids walkways, while a negative value between `DirectionsPriority.low` and `DirectionsPriority.medium` avoids walkways, and a positive value between `DirectionsPriority.medium` and `DirectionsPriority.high` prefers walkways. A value of −0.1 results in less verbose routes in cities where sidewalks and crosswalks are generally mapped as separate footpaths.
      */
-    open var walkwayPriority: DirectionsPriority = .default
+    open var walkwayPriority: DirectionsPriority?
     
     /**
-     The expected uniform travel speed measured in meters per second.
+     An optional that represents the expected uniform travel speed measured in meters per second.
```
1ec5, line 144: It isn’t necessary to call out the optionality of the type; that’s already clear to the developer in Swift. It is important to note what happens if the property is unspecified: I think we can simply say for all three properties that the Directions API will choose the most reasonable value.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -123,31 +123,31 @@ open class RouteOptions: DirectionsOptions {
     open var roadClassesToAvoid: RoadClasses = []
     
     /**
-     A number that influences whether the route should prefer or avoid alleys or narrow service roads between buildings.
+     An optional that represents the number that influences whether the route should prefer or avoid alleys or narrow service roads between buildings.
      
-     This property has no effect unless the profile identifier is set to `DirectionsProfileIdentifier.automobile` or `DirectionsProfileIdentifier.walking`.
+     This property has no effect unless it is explicitly set as it defaults to nil, and works best when the profile identifier is set to `DirectionsProfileIdentifier.automobile` or `DirectionsProfileIdentifier.walking`.
```
1ec5, line 128: The [Directions API documentation](https://docs.mapbox.com/api/navigation/directions/#retrieve-directions) doesn’t explicitly restrict this option to any particular profile, even if it happens to be unsupported in the cycling profile, so I think we can remote this note entirely. It’s probably unnecessary to state that an unset property is set to `nil` and therefore has no effect, since no default value is present in the property signature.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -123,31 +123,31 @@ open class RouteOptions: DirectionsOptions {
...
     
     /**
-     A number that influences whether the route should prefer or avoid roads or paths that are set aside for pedestrian-only use (walkways or footpaths).
+     An optional that represents the number that influences whether the route should prefer or avoid roads or paths that are set aside for pedestrian-only use (walkways or footpaths).
      
-     This property has no effect unless the profile identifier is set to `DirectionsProfileIdentifier.walking`. You can adjust this property to avoid [sidewalks and crosswalks that are mapped as separate footpaths](https://wiki.openstreetmap.org/wiki/Sidewalks#Sidewalk_as_separate_way), which may be more granular than needed for some forms of pedestrian navigation.
+     This property has no effect unless it is explicitly set as it defaults to nil, and works best when the profile identifier is set to `DirectionsProfileIdentifier.walking`. You can adjust this property to avoid [sidewalks and crosswalks that are mapped as separate footpaths](https://wiki.openstreetmap.org/wiki/Sidewalks#Sidewalk_as_separate_way), which may be more granular than needed for some forms of pedestrian navigation.
```
1ec5, line 137: The [Directions API documentation](https://docs.mapbox.com/api/navigation/directions/#optional-parameters-for-the-mapboxwalking-profile) explicitly limits this option to the `mapbox/walking` profile. I think it’s safer to retain the documentation as it was, for consistency with the API documentation, even as we remove the formal client-side check.

1ec5 (approved):
Good to go as soon as the documentation feedback is addressed. Thanks for knocking this one out!

S2Ler (approved):

Issue #556, opened 2021-06-28
Leg source and destination are mismatched when route contains trackpoint or silent waypoint


1ec5:
When the RouteOptions contains a waypoint whose `separatesLegs` property is set to `false`, the resulting route’s legs have mismatching `source` and `destination` property values, as though the `separatesLegs` property were set to `true`.

The following code should pass `DirectionsOptions.legSeparators` into `Array.populate(waypoints:)`, rather than passing in `DirectionsOptions.waypoints`:

https://github.com/mapbox/mapbox-directions-swift/blob/2d9edcedef9506546fa390c8f7f10de848b2aec2/Sources/MapboxDirections/DirectionsResult.swift#L44-L45

This regression of #358 was introduced by #406.

/cc @mapbox/navigation-ios @ZiZasaurus

1ec5:
This issue is even more likely to affect Map Matching API usage, because any trackpoint would be handled like a Directions API silent waypoint.

1ec5:
Fixed in #560.

PR #560, opened 2021-07-08
Fix bug where leg source and destination are mismatched when route contains trackpoint or silent waypoint

jill-cardamon:
This PR addresses #556 by passing `DirectionsOptions.legSeparators` into `Array.populate(waypoints:)`, rather than passing in `DirectionsOptions.waypoints`. It fixes a regression of #358 was introduced by #406.

/cc @ZiZasaurus


S2Ler (approved):
LGTM.

One nit: Please start commit comments with upper case for consistency.

azarovalex (approved):

PR #566, opened 2021-07-22
[DRAFT] Extend Example with ability to edit waypoints

S2Ler:


S2Ler:
Unfortunately, current approach with SwiftUI didn't payed out. Need to find another implementation which won't crash.

DirectionsPlayground/DirectionsPlayground/Sources/QueriesList.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import SwiftUI
+
+struct QueriesList: View {
+    @State
+    var queries: [Query] = [] {
```
kried, line 6: Should we add the first Query to the screen or should we add text explaining that you need to add a query to see how Directions work? It might be confusing to see the empty list

DirectionsPlayground/DirectionsPlayground/Sources/WaypointsEditor.swift
```diff
@@ -0,0 +1,119 @@
...
+}
+
+struct WaypointView: View {
+    @Binding var waypoint: Waypoint
+    @State private var latitudeString: String
+    @State private var longitudeString: String    
+
+    init(waypoint: Binding<Waypoint>) {
+        _waypoint = waypoint
+        _latitudeString = State<String>(initialValue: waypoint.wrappedValue.latitude.description)
```
kried, line 88: Would it be possible to use the same format for both calls, either `State<String>(`, either `.init(`?

DirectionsPlayground/DirectionsPlayground/Sources/InfoButton.swift
```diff
@@ -0,0 +1,38 @@
...
+        }
+        .buttonStyle(.bordered)
+    }
+}
+
+extension InfoButton.DocUrl {
+    var urlString: String {
+        switch self {
+        case .waypointAllowsArrivingOnOppositeSide:
+            return "https://docs.mapbox.com/ios/directions/api/2.3.0/Classes/Waypoint.html#/s:16MapboxDirections8WaypointC28allowsArrivingOnOppositeSideSbvp"
```
kried, line 27: It looks like we would have to update the version here with each Directions SDK release.

Should we use 2.8.0 here for now?

DirectionsPlayground/DirectionsPlayground/Sources/QueryEditor.swift
```diff
@@ -0,0 +1,67 @@
+import Foundation
+import SwiftUI
+import Combine
+import MapboxDirections
+
+struct QueryEditor: View {
```
kried, line 6: As a suggestion, names like these sound as if it was a model, not a view. Would it be possible to use names like `QueryEditView`?

kried:
It is a great idea to rewrite the example to use SPM and SwiftUI!

PR #567, opened 2021-07-22
Fixed RouteStep.exitIndex parsing

### LINKED ISSUE #508: RouteStep exitIndex not filled with roundabout
Hello all,

According to documentation, exitIndex should indicate the number of exits from the approach to the recommended outlet when the maneuver is a roundabout.
From my experience, this is not working and exitIndex is always nil.

S2Ler:
It was expected that exitIndex was at the top level of RouteStep json
dictionary. But it is actually at "maneuver" sub dictionary.

Fixes #508 and mapbox/mapbox-navigation-ios#2742

### DIFF
--- a/Tests/MapboxDirectionsTests/RouteStepTests.swift
+++ b/Tests/MapboxDirectionsTests/RouteStepTests.swift
@@ -336,14 +336,19 @@ class RouteStepTests: XCTestCase {
             var newRoute: Route?
             XCTAssertNoThrow(newRoute = try decoder.decode(Route.self, from: jsonData))
             XCTAssertNotNil(newRoute)
+            guard let leg = newRoute?.legs.first else {
+                XCTFail("No legs found"); return
+            }
             
-            XCTAssert(newRoute!.legs.first!.incidents!.first!.kind == Incident.Kind.miscellaneous)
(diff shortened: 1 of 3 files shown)


Tests/MapboxDirectionsTests/RouteStepTests.swift
```diff
@@ -336,14 +336,19 @@ class RouteStepTests: XCTestCase {
             var newRoute: Route?
             XCTAssertNoThrow(newRoute = try decoder.decode(Route.self, from: jsonData))
             XCTAssertNotNil(newRoute)
+            guard let leg = newRoute?.legs.first else {
+                XCTFail("No Logs Found"); return
```
1ec5, line 340:
```suggestion
                XCTFail("No legs found"); return
```

1ec5 (approved):
Suggested changelog entry:

* Fixed an issue where `RouteStep.exitIndex` was always unset. ([#567](https://github.com/mapbox/mapbox-directions-swift/pull/567))

Tests/MapboxDirectionsTests/RouteStepTests.swift
S2Ler: done

PR #572, opened 2021-08-11
Option added to command line tool for outputting Xcode-compatible GPX file

### LINKED ISSUE #540: Add option to command line tool for outputting Xcode-compatible GPX file
The command line tool added in #469 supports outputting the route as plain text or JSON. But it would be super cool if it could instead output the route geometry as a GPX file formatted for use with [Wormhole](https://github.com/1ec5/wormhole/) (in other words, Xcode’s built-in location simulation feature).

https://github.com/mapbox/mapbox-directions-swift/blob/2ffc5f0cf9cc5a912d6d2187de854931ed79ab7e/Sources/MapboxDirectionsCLI/ProcessCommand.swift#L24

#375 tracks going in the other direction, converting a GPX file back into a Directions API request.

jill-cardamon:
Fixes #540.

### DIFF
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -44,13 +44,16 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions > 
             var time = Date()
             
             routes.forEach { route in
-                let coordinates = interpolate(route: route)
+                let coordinates = interpolate(route: route, timeInterval: timeInterval)
                 for coord in coordinates {
                     guard let lat = coord?.latitude, let lon = coord?.longitude else { continue }
                     gpxText.append("\n<wpt lat=\"\(lat)\" lon=\"\(lon)\">")
                     gpxText.append("\n\t<time> \(dateFormatter.string(from: time)) </time>")
                     gpxText.append("\n</wpt>")
                     time.addTimeInterval(timeInterval)
+                }
+                if routes.count > 1 {
+                    gpxText.append("<!--Moving to next route-->")
                 }
             }
             gpxText.append("\n</gpx>")
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -18,7 +18,9 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions > 
         return try encoder.encode(result)
     }
     
-    private func processOutput(_ data: Data, routeResponse: RouteResponse?) throws {
+    private func processOutput(_ data: Data,
+                               routeResponse: RouteResponse? = nil,
+                               matchResponse: MapMatchingResponse? = nil) throws {
         var outputText: String = ""
         
         switch options.outputFormat {
@@ -31,37 +33,30 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions > 
             }
             
         case .gpx:
-            #if !os(Linux)
             var gpxText: String = String("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
             gpxText.append("\n<gpx xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns=\"http://www.topografix.com/GPX/1/1\" xsi:schemaLocation=\"http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd\" version=\"1.1\">")
             
-            guard let routeResponse = routeResponse,
-                  let routes = routeResponse.routes else { return }
-            
-            let timeInterval: TimeInterval = 1
-            let dateFormatter = ISO8601DateFormatter()
-            dateFormatter.formatOptions = .withInternetDateTime
-            var time = Date()
-            
-            routes.forEach { route in
-                let coordinates = interpolate(route: route, timeInterval: timeInterval)
-                for coord in coordinates {
-                    guard let lat = coord?.latitude, let lon = coord?.longitude else { continue }
-                    gpxText.append("\n<wpt lat=\"\(lat)\" lon=\"\(lon)\">")
-                    gpxText.append("\n\t<time> \(dateFormatter.string(from: time)) </time>")
-                    gpxText.append("\n</wpt>")
-                    time.addTimeInterval(timeInterval)
+            if let routeResponse = routeResponse,
+               let routes = routeResponse.routes {
+                routes.forEach { route in
+                    let text = populateGPX(route)
+                    gpxText.append(text)
+                    if routes.count > 1 {
+                        gpxText.append("<!--Moving to next route-->")
+                    }
                 }
-                if routes.count > 1 {
-                    gpxText.append("<!--Moving to next route-->")
+            } else if let matchResponse = matchResponse,
+                      let matches = matchResponse.matches {
+                matches.forEach { match in
+                    let text = populateGPX(nil, match)
+                    gpxText.append(text)
+                    if matches.count > 1 {
+                        gpxText.append("<!--Moving to next match-->")
+                    }
                 }
             }
             gpxText.append("\n</gpx>")
             outputText = gpxText
-            #else
-            print("GPX output requires use of CoreLocation, a framework that is not available on Linux.")
-            exit(1)
-            #endif
         }
         
         if let outputPath = options.outputPath {
@@ -73,20 +68,53 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions > 
         }
     }
     
-    private func interpolate(route: Route, timeInterval: TimeInterval) -> [LocationCoordinate2D?] {
-        guard route.expectedTravelTime > 0, let polyline = route.shape,
-              let firstCoordinate = polyline.coordinates.first else { return [] }
+    private func populateGPX(_ route: Route? = nil, _ match: Match? = nil) -> String {
+        let timeInterval: TimeInterval = 1
+        let dateFormatter = ISO8601DateFormatter()
+        dateFormatter.formatOptions = .withInternetDateTime
+        var time = Date()
+        var text: String = ""
+        var coordinates: [LocationCoordinate2D?] = []
+        
+        if route != nil {
+            guard let route = route else { return "" }
+            coordinates = interpolate(shape: route.shape,
+                                      expectedTravelTime: route.expectedTravelTime,
+                                      distance: route.distance,
+                                      timeInterval: timeInterval)
+        } else if match != nil {
+            guard let match = match else { return "" }
+            coordinates = interpolate(shape: match.shape,
+                                      expectedTravelTime: match.expectedTravelTime,
+                                      distance: match.distance,
+                                     timeInterval: timeInterval)
+        }
+        
+        for coord in coordinates {
+            guard let lat = coord?.latitude, let lon = coord?.longitude else { continue }
+            text.append("\n<wpt lat=\"\(lat)\" lon=\"\(lon)\">")
+            text.append("\n\t<time> \(dateFormatter.string(from: time)) </time>")
+            text.append("\n</wpt>")
+            time.addTimeInterval(timeInterval)
+        }
+        return text
+    }
+    
+    private func interpolate(shape: LineString?, expectedTravelTime: TimeInterval, distance: LocationDistance, timeInterval: TimeInterval) -> [LocationCoordinate2D?] {
+        guard expectedTravelTime > 0, let polyline = shape,
+              let firstCoordinate = polyline.coordinates.first,
+              let lastCoordinate = polyline.coordinates.last else { return [] }
         
         var distanceAway: LocationDistance = 0
-        let distancePerTick = route.distance/route.expectedTravelTime
+        let distancePerTick = distance/expectedTravelTime
         var interpolatedCoordinates = [firstCoordinate]
-        while distanceAway <= route.distance {
+        while distanceAway <= distance {
             if let nextCoordinate = polyline.coordinateFromStart(distance: distanceAway) {
                 interpolatedCoordinates.append(nextCoordinate)
             }
             distanceAway += distancePerTick * timeInterval
         }
-        interpolatedCoordinates.append(polyline.coordinates.last)
+        interpolatedCoordinates.append(lastCoordinate)
         return interpolatedCoordinates
     }
     
@@ -109,12 +137,17 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions > 
                             .credentials: BogusCredentials]
         
         var routeResponse: RouteResponse?
+        var matchResponse: MapMatchingResponse?
         if options.outputFormat == .gpx {
             let gpxData = try String(contentsOfFile: options.inputPath).data(using: .utf8)!
-            routeResponse = try! decoder.decode(RouteResponse.self, from: gpxData)
+            do {
+                routeResponse = try decoder.decode(RouteResponse.self, from: gpxData)
+            } catch {
+                matchResponse = try decoder.decode(MapMatchingResponse.self, from: gpxData)
+            }
         }
         let data = try processResponse(decoder, type: ResponceType.self, from: input)
         
-        try processOutput(data, routeResponse: routeResponse)
+        try processOutput(data, routeResponse: routeResponse, matchResponse: matchResponse)
     }
 }
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -65,7 +65,7 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions > 
     }
     
     private func interpolate(route: Route) -> [CLLocationCoordinate2D?] {
-        guard let polyline = route.shape else { return [] }
+        guard route.expectedTravelTime > 0, let polyline = route.shape else { return [] }
         
 //        route.legs.forEach { leg in
 //            let expectedSegmentTravelTimes = leg.expectedSegmentTravelTimes
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -16,7 +16,7 @@ extension MapMatchingResponse: DirectionsResultsProvider {
     var directionsResults: [DirectionsResult]? { matches }
 }
 
-class CodingOperation<ResponceType : Codable & DirectionsResultsProvider, OptionsType : DirectionsOptions > {
+class CodingOperation<ResponseType : Codable & DirectionsResultsProvider, OptionsType : DirectionsOptions > {
     
     // MARK: - Parameters
     
@@ -24,8 +24,8 @@ class CodingOperation<ResponceType : Codable & DirectionsResultsProvider, Option
     
     // MARK: - Helper methods
     
-    private func processResponse(_ decoder: JSONDecoder, type: ResponceType.Type, from data: Data) throws -> (Data, ResponceType) {
-        let result = try decoder.decode(type, from: data)
+    private func processResponse(_ decoder: JSONDecoder, from data: Data) throws -> (Data, ResponseType) {
+        let result = try decoder.decode(ResponseType.self, from: data)
         let encoder = JSONEncoder()
         let data = try encoder.encode(result)
         return (data, result)
@@ -134,7 +134,7 @@ class CodingOperation<ResponceType : Codable & DirectionsResultsProvider, Option
         decoder.userInfo = [.options: directionsOptions,
                             .credentials: BogusCredentials]
         
-        let (data, directionsResultsProvider) = try processResponse(decoder, type: ResponceType.self, from: input)
+        let (data, directionsResultsProvider) = try processResponse(decoder, from: input)
         
         try processOutput(data, directionsResultsProvider: directionsResultsProvider)
     }
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -24,10 +24,11 @@ class CodingOperation<ResponceType : Codable & DirectionsResultsProvider, Option
     
     // MARK: - Helper methods
     
-    private func processResponse<T>(_ decoder: JSONDecoder, type: T.Type, from data: Data) throws -> Data where T : Codable {
+    private func processResponse(_ decoder: JSONDecoder, type: ResponceType.Type, from data: Data) throws -> (Data, ResponceType) {
         let result = try decoder.decode(type, from: data)
         let encoder = JSONEncoder()
-        return try encoder.encode(result)
+        let data = try encoder.encode(result)
+        return (data, result)
     }
     
     private func processOutput(_ data: Data,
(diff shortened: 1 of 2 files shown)


Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,22 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+            // do we need to include additional tags like metadata or the schema version?
+            
+            guard let routeResponse = routeResponse else { return }
+            guard let routes = routeResponse.routes else { return }
+            routes.forEach { route in
+                let shape = route.shape
+                for coord in shape!.coordinates {
+                    gpxText.append("\n<wpt lat=\"\(coord.latitude)\" lon=\"\(coord.longitude)\">")
+                    gpxText.append("</wpt>")
+                }
```
S2Ler, line 71: - Do you know how far away each coordinate from each other? 
- What about timestamp of the coordinate? 
- Have you tried running this gpx in Xcode?

jill-cardamon: I meant to open this as a draft. I am working on adding the timestamps now.

jill-cardamon: Timestamps are added (using `total distance/expected time`, and the output gpx works with Xcode. I'll check with a multi-leg trip too.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,29 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
                let jsonData = try? JSONSerialization.data(withJSONObject: object, options: [.prettyPrinted]) {
                 outputText = String(data: jsonData, encoding: .utf8)!
             }
+        case .gpx:
+            var gpxText: String = String("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
```
bamx23, line 60: Maybe it is worth adding a special library (like https://github.com/vincentneo/CoreGPX)?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,29 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+            gpxText.append("\n<gpx xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns=\"http://www.topografix.com/GPX/1/1\" xsi:schemaLocation=\"http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd\" version=\"1.1\">")
+            
+            guard let routeResponse = routeResponse else { return }
+            guard let routes = routeResponse.routes else { return }
+            
+            let dateFormatter = DateFormatter()
```
bamx23, line 66: Can `ISO8601DateFormatter` be used here?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -97,6 +123,13 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
         
         decoder.userInfo = [.options: directionsOptions!,
                             .credentials: BogusCredentials]
+        
+        var routeResponse: RouteResponse!
```
bamx23, line 127:
```suggestion
        var routeResponse: RouteResponse?
```

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,29 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
                let jsonData = try? JSONSerialization.data(withJSONObject: object, options: [.prettyPrinted]) {
                 outputText = String(data: jsonData, encoding: .utf8)!
             }
+        case .gpx:
+            var gpxText: String = String("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
+            gpxText.append("\n<gpx xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns=\"http://www.topografix.com/GPX/1/1\" xsi:schemaLocation=\"http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd\" version=\"1.1\">")
+            
+            guard let routeResponse = routeResponse else { return }
+            guard let routes = routeResponse.routes else { return }
```
bamx23, line 64: NIT
```suggestion
            guard let routeResponse = routeResponse,
                  let routes = routeResponse.routes else { return }
```

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -97,6 +123,13 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
         
         decoder.userInfo = [.options: directionsOptions!,
                             .credentials: BogusCredentials]
+        
+        var routeResponse: RouteResponse!
+        if outputFormat == .gpx {
+            guard let gpxData = try String(contentsOfFile: inputPath).data(using: .utf8) else { exit(1)}
```
bamx23, line 129: NIT
```suggestion
            guard let gpxData = try String(contentsOfFile: inputPath).data(using: .utf8) else { exit(1) }
```

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: What would be the benefit of using `ISO8601DateFormatter` over `DateFormatter`?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: We could, but looking at the examples from that library, it seems like it would be the almost the same number of lines of code (excluding the initial header stuff) since we would still need to parse the RouteResponse on our side to get the lat/lon.

bamx23: Sure, I just have some concerns about generating XML content by hand. All in all, it's just a recommendation as extra dependency has its own pros and cons.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
bamx23: You don't have to provide a format string as what you have is actually an ISO8601 format.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,29 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss"
+            var time = Date()
+            
+            routes.forEach { route in
+                let shape = route.shape
+                let timeInterval = TimeInterval(route.distance/route.expectedTravelTime)
+                for coord in shape!.coordinates {
+                    gpxText.append("\n<wpt lat=\"\(coord.latitude)\" lon=\"\(coord.longitude)\">")
+                    gpxText.append("\n\t<time> \(dateFormatter.string(from: time))Z </time>")
```
bamx23, line 75: Do all coordinates have the same timestamp?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: Ah I see.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: I'll switch to `ISO8601DateFormatter` then. Thanks for the recommendation

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: Nope. The `timeInterval` is added at the end of the inner loop.

bamx23: Oh, I see. But something looks wrong... That time interval stores an average speed actually. Should we divide estimated duration on coordinates count instead? Or can we use steps here?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
S2Ler: I wouldn't include a third part here as @jill-cardamon implementation is good enough for this limiting use case,  I think.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,29 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+                  let routes = routeResponse.routes else { return }
+            
+            let dateFormatter = ISO8601DateFormatter()
+            dateFormatter.formatOptions = .withInternetDateTime
+            var time = Date()
+            
+            routes.forEach { route in
+                let shape = route.shape
+                let timeInterval = TimeInterval(route.distance/route.expectedTravelTime)
+                for coord in shape!.coordinates {
```
S2Ler, line 73: I think you shall take into account the distance between coordinates to generate correct timestamp so that the speed will be ok.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
bamx23, line 60: Maybe it would be better to use an array of strings and to join them at the end of the case? I heard that it works faster and you don't have to start each line with `\n` manually.

1ec5: Also, if you find the backslashes annoying, you can replace each `"` with a `'`, because single-quote-delimited attributes are also valid XML.

(A more ambitious refactor could use the object-oriented [XMLDocument](https://developer.apple.com/documentation/foundation/xmldocument) API or a Swiftier library like [XMLCoder](https://github.com/MaxDesiatov/XMLCoder/), but an extra dependency might be overkill for this simple XML document.)

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -70,6 +97,22 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
         }
     }
     
+    private func interpolate(route: Route?) -> [CLLocationCoordinate2D?] {
+        guard let route = route else { return [] }
```
1ec5, line 101: It doesn’t appear to be necessary for this method’s argument to be optional.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -70,6 +97,22 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+        
+        var distanceAway: CLLocationDistance = 0
+        let distance = route.distance/route.expectedTravelTime
```
1ec5, line 104: This could result in division by zero if the route’s waypoints happen to snap to the same coordinate.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -70,6 +97,22 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+        guard let polyline = route.shape else { return [] }
+        var interpolatedCoordinates = [route.shape?.coordinates.first]
```
1ec5, line 106: The optional chaining on this line complicates the code that uses this method, because it now has to account for each individual element being `nil`, even though there should be no reason for a GPX route to contain nonexistent locations. If there’s no shape associated with the route, then `polyline` would be `nil` so this method would’ve already returned anyways.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -53,6 +56,30 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+            
+            guard let routeResponse = routeResponse,
+                  let routes = routeResponse.routes else { return }
+            
+            let timeInterval: TimeInterval = 1
+            let dateFormatter = ISO8601DateFormatter()
+            dateFormatter.formatOptions = .withInternetDateTime
+            var time = Date()
+            
+            routes.forEach { route in
```
1ec5, line 71: Each route in the response will appear in order in the GPX, causing it to jump from the arrival step of one route to the departure step of another route. This is reasonable for simulation purposes: if the user only wants one route, they can request no alternatives.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -70,6 +97,22 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+    private func interpolate(route: Route?) -> [CLLocationCoordinate2D?] {
+        guard let route = route else { return [] }
+        
+        var distanceAway: CLLocationDistance = 0
+        let distance = route.distance/route.expectedTravelTime
+        guard let polyline = route.shape else { return [] }
+        var interpolatedCoordinates = [route.shape?.coordinates.first]
+        
+        while distanceAway <= route.distance {
+            let nextCordinate = polyline.coordinateFromStart(distance: distanceAway)
```
1ec5, line 109: Typo: `nextCoordinate`.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: yep thanks for pointing that out!

Sources/MapboxDirectionsCLI/ProcessCommand.swift
jill-cardamon: oh my..I will fix this then.

Sources/MapboxDirectionsCLI/ProcessCommand.swift
1ec5: The template has said this for a few years, but every time I’ve tested it, you can’t get Xcode to interpolate anything.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -40,6 +73,24 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
         }
     }
     
+    #if !os(Linux)
+    private func interpolate(route: Route) -> [CLLocationCoordinate2D?] {
```
bamx23, line 77: What if we use `LocationCoordinate2D` from Turf here to allow Linux?

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -40,6 +73,24 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
+        guard route.expectedTravelTime > 0, let polyline = route.shape,
+              let firstCoordinate = polyline.coordinates.first else { return [] }
+        
+        var distanceAway: CLLocationDistance = 0
+        let distance = route.distance/route.expectedTravelTime
```
bamx23, line 82: It's a bit confusing to call `distance / time` a distance. Could you rename it to `averageSpeed` or `distancePerTick`?
And if you add `timeInterval` as an argument of this function, then you can have `distanceAway += speed * timeInterval` in the cycle below.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -40,6 +73,24 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
+        var interpolatedCoordinates = [firstCoordinate]
+        while distanceAway <= route.distance {
+            if let nextCoordinate = polyline.coordinateFromStart(distance: distanceAway) {
+                interpolatedCoordinates.append(nextCoordinate)
+            }
+            distanceAway += distance
+        }
+        return interpolatedCoordinates
```
bamx23, line 90: Not sure if it's important but we can get a bit off the final coordinate. What about adding the `polyline.coordinates.last` into `interpolatedCoordinates`?

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -29,6 +32,36 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
+            
+            guard let routeResponse = routeResponse,
+                  let routes = routeResponse.routes else { return }
+            
+            let timeInterval: TimeInterval = 1
+            let dateFormatter = ISO8601DateFormatter()
+            dateFormatter.formatOptions = .withInternetDateTime
+            var time = Date()
+            
+            routes.forEach { route in
```
bamx23, line 49: Should we somehow split different routes within the GPX file? At least by adding an XML comment.

1ec5: A comment would be helpful for debugging purposes; otherwise, I think it’s OK that the track jumps from the end of one route to the beginning of the next: https://github.com/mapbox/mapbox-directions-swift/pull/572#discussion_r719564488.

bamx23: It's OK, sure. But I can imagine the case when it might be inconvenient for a user to make a different route request, but they might still want to simulate only for one of the routes. We can add a CLI argument to filter a route, but this comment idea is just a bit simpler.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -29,6 +29,39 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
+                    time.addTimeInterval(timeInterval)
+                }
+                if routes.count > 1 {
+                    gpxText.append("<!--Moving to next route-->")
+                }
+            }
+            gpxText.append("\n</gpx>")
+            outputText = gpxText
+            #else
+            print("GPX output requires use of CoreLocation, a framework that is not available on Linux.")
```
bamx23, line 62: Is it still true?

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -57,8 +107,14 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
         
         decoder.userInfo = [.options: directionsOptions,
                             .credentials: BogusCredentials]
+        
+        var routeResponse: RouteResponse?
+        if options.outputFormat == .gpx {
+            let gpxData = try String(contentsOfFile: options.inputPath).data(using: .utf8)!
+            routeResponse = try! decoder.decode(RouteResponse.self, from: gpxData)
+        }
         let data = try processResponse(decoder, type: ResponceType.self, from: input)
```
bamx23, line 116: NIT: `ResponseType`

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -57,8 +107,14 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
         
         decoder.userInfo = [.options: directionsOptions,
                             .credentials: BogusCredentials]
+        
+        var routeResponse: RouteResponse?
+        if options.outputFormat == .gpx {
+            let gpxData = try String(contentsOfFile: options.inputPath).data(using: .utf8)!
+            routeResponse = try! decoder.decode(RouteResponse.self, from: gpxData)
```
bamx23, line 114:
```suggestion
            routeResponse = try! decoder.decode(RouteResponse.self, from: input)
```
And what if we do this within the `case .gpx:` in `processOutput` method using the existing `data`? Is it possible?

Sources/MapboxDirectionsCLI/CodingOperation.swift
jill-cardamon: removed!

Sources/MapboxDirectionsCLI/CodingOperation.swift
jill-cardamon: yep. included instead. with previous iteration, this wasn't possible.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -18,7 +18,9 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
         return try encoder.encode(result)
     }
     
-    private func processOutput(_ data: Data) throws {
+    private func processOutput(_ data: Data,
+                               routeResponse: RouteResponse? = nil,
+                               matchResponse: MapMatchingResponse? = nil) throws {
```
bamx23, line 23: It's possible to reduce code duplication with this trick:
```swift
protocol DirectionsResultsProvider {
    var directionsResults: [DirectionsResult]? { get }
}

extension RouteResponse: DirectionsResultsProvider {
    var directionsResults: [DirectionsResult]? { routes }
}

extension MapMatchingResponse: DirectionsResultsProvider {
    var directionsResults: [DirectionsResult]? { matches }
}
```
And then `processOutput` can work with `DirectionsResultsProvider` as `shape`, `expectedTravelTime` and `distance` are parts of `DirectionsResult` that is common for both `Route` and `Match`.

And even more, we have `ResponceType` generic type in `CodingOperation` that is either `RouteResponse` or `MapMatchingResponse` and we can mark `ResponceType: DirectionsResultsProvider` and use it directly. `processResponse` already decodes `ResponceType` so we can reuse it.

bamx23 (changes requested):

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -18,7 +30,8 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
         return try encoder.encode(result)
```
bamx23, line 30: The result here is exactly `directionsResultsProvider` needed below. We can return a tuple `(Data, ResponceType)` here (and also remove generic parameter as `T` is always `ResponceType` here.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -57,8 +129,17 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
         
         decoder.userInfo = [.options: directionsOptions,
                             .credentials: BogusCredentials]
+        
+        var directionsResultsProvider: DirectionsResultsProvider?
+        if options.outputFormat == .gpx {
+            do {
+                directionsResultsProvider = try decoder.decode(ResponceType.self, from: input) as DirectionsResultsProvider
+            } catch {
+                directionsResultsProvider = try decoder.decode(MapMatchingResponse.self, from: input) as DirectionsResultsProvider
```
bamx23, line 138: Do we really need to fallback to `MapMatchingResponse` here? The command will fail in `processResponse` in case data is not decodable with `ResponceType`.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -1,24 +1,38 @@
...
+class CodingOperation<ResponceType : Codable & DirectionsResultsProvider, OptionsType : DirectionsOptions > {
     
     // MARK: - Parameters
     
     let options: ProcessingOptions
     
     // MARK: - Helper methods
     
-    private func processResponse<T>(_ decoder: JSONDecoder, type: T.Type, from data: Data) throws -> Data where T : Codable {
+    private func processResponse(_ decoder: JSONDecoder, type: ResponceType.Type, from data: Data) throws -> (Data, ResponceType) {
```
bamx23, line 27: We don't need this "type" argument anymore

bamx23 (approved):

PR #575, opened 2021-08-17
 Add congestion_numeric attribute support

chezzdev:
Directions part of https://github.com/mapbox/navigation-sdks/issues/892.

This PR adds support for the new attribute of congestion level in numeric form.

Currently, it only adds a new parameter without removing the congestion level attribute. However, per https://github.com/mapbox/navigation-sdks/issues/892#issuecomment-870613463 it may be considered deprecated:
> `congestion` value should be deprecated. NavSDK is the only consumer of this value and API is providing it for legacy reasons.

Is there anything else preventing us from removing `CongestionLevel` from the library?

cc: @mapbox/navigation-ios


Sources/MapboxDirections/Congestion.swift
```diff
@@ -37,3 +37,8 @@ public enum CongestionLevel: String, Codable, CaseIterable {
      */
     case severe
 }
+
+/**
+ `NumericCongestionLevel` is the level of traffic congestion along a road segment in numeric form, from 0-100. A value of 0 indicates no congestion, a value of 100 indicates maximum congestion. Entries may be `nil` if congestion on that segment is not known.
+ */
+public typealias NumericCongestionLevel = Int
```
bamx23, line 44: Maybe it should be `Int?` as you say:
> Entries may be `nil` if congestion on that segment is not known.

bamx23 (approved):

Sources/MapboxDirections/Congestion.swift
chezzdev: Making a type alias of an optional makes the type a bit less clear to use. Concerning doc comment - good catch, I've moved it to `segmentNumericCongestionLevels` doc.

Udumft (approved):

PR #576, opened 2021-08-18
Add fall back to Directions API request when no input file is specified

### LINKED ISSUE #570: Fall back to Directions API request when no input file is specified
The `mapbox-directions-swift` command line tool’s `route` and `match` subcommands require the user to pass in an input file containing a valid Directions API or Map Matching API response in JSON format. This makes the tool useful only for round-tripping from and to JSON or for converting between the supported file formats.

The tool could be more generally useful if it could send a real HTTP request to the Directions API or Map Matching API when the input file is unspecified. For example, an iOS, macOS, or Linux developer could easily serialize a RouteOptions as JSON using JSONEncoder and print it out, then pass it into `mapbox-directions-swift` to get a test fixture. Once #540 is implemented, an iOS developer could drop the resulting GPX file into their application project or [Wormhole](https://github.com/1ec5/wormhole/) and simulate an arbitrary route somewhere far away from Infinite Loop and I-280.

For the most part, this would be just like the usual process for getting started with MapboxDirections. However, as a command line tool, it needs to wait until it receives a response from the server before exiting.

/cc @mapbox/navigation-ios

jill-cardamon:
Fixes #570.

Currently rebased on top of the gpx output enhancement for testing. Can be merged after gpx output PR is merged.

### DIFF
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -3,9 +3,10 @@ import MapboxDirections
 import Turf
 import CoreLocation
 
-private let BogusCredentials = Credentials(accessToken: "pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede")
-let accessToken = ProcessInfo.processInfo.environment["access_token"]
-let credentials = DirectionsCredentials(accessToken: accessToken)
+let accessToken: String? =
+    ProcessInfo.processInfo.environment["access_token"] ??
+    UserDefaults.standard.string(forKey: "MBXAccessToken")
+let credentials = DirectionsCredentials(accessToken: accessToken!)
 private let directions = Directions(credentials: credentials)
 
 protocol DirectionsResultsProvider {
@@ -146,8 +147,6 @@ class CodingOperation<ResponseType : Codable & DirectionsResultsProvider, Option
             }
             
             responseData = data
-            print("Fetched data: \(data)")
-            print("Fetched response: \(String(describing: response))")
             semaphore.signal()
         }
         
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -121,21 +121,15 @@ class CodingOperation<ResponseType : Codable & DirectionsResultsProvider, Option
         return interpolatedCoordinates
     }
     
-    private func requestResponse(_ coordinates: [Waypoint]?) -> (Data?, Data) {
+    private func requestResponse(_ coordinates: [Waypoint]?, includesSteps: Bool?) -> (Data?, Data) {
         let semaphore = DispatchSemaphore(value: 0)
         
-        var waypoints: [Waypoint]
-        if coordinates != nil {
-            waypoints = coordinates!
-        } else {
-            waypoints = [
-                Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.9131752, longitude: -77.0324047), name: "Mapbox"),
-                Waypoint(coordinate: CLLocationCoordinate2D(latitude: 38.8977, longitude: -77.0365), name: "White House"),
-            ]
-        }
+        guard let waypoints = coordinates, let includesSteps = includesSteps else {
+            print("Failed to request response without coordinates.")
+            exit(1) }
         
         let options = RouteOptions(waypoints: waypoints, profileIdentifier: .automobileAvoidingTraffic)
-        options.includesSteps = true
+        options.includesSteps = includesSteps
         var responseData: Data?
         
         let url = directions.url(forCalculating: options)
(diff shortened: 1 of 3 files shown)


Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -89,21 +131,33 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
     
     func execute() throws {
         
-        let input = FileManager.default.contents(atPath: NSString(string: options.inputPath).expandingTildeInPath)!
+//        let input = FileManager.default.contents(atPath: NSString(string: options.inputPath).expandingTildeInPath)!
```
1ec5, line 134: Remove this line if it’s no longer needed.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -89,21 +131,33 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
         let config = FileManager.default.contents(atPath: NSString(string: options.configPath).expandingTildeInPath)!
+        let input: Data!
```
1ec5, line 136: It looks like this variable is set in both branches below, and it’s only ever read by an `if let` statement, so it might not need to be implicitly unwrapped.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -89,21 +131,33 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
+                            .credentials: credentials]
         
         var routeResponse: RouteResponse?
         if options.outputFormat == .gpx {
-            let gpxData = try String(contentsOfFile: options.inputPath).data(using: .utf8)!
-            routeResponse = try! decoder.decode(RouteResponse.self, from: gpxData)
+            if let gpxData = input {
+                routeResponse = try! decoder.decode(RouteResponse.self, from: gpxData)
+            } else if let gpxData = try String(contentsOfFile: options.inputPath!).data(using: .utf8) {
+                routeResponse = try! decoder.decode(RouteResponse.self, from: gpxData)
```
1ec5, line 157: A failure to decode could be something other than an internal error, so instead of force-unwrapping, we should catch the error, print it to stderr, and exit with an error code.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -3,7 +3,11 @@ import MapboxDirections
 import Turf
 import CoreLocation
 
-private let BogusCredentials = DirectionsCredentials(accessToken: "pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede")
+let accessToken: String? =
+    ProcessInfo.processInfo.environment["access_token"] ??
```
1ec5, line 7: `MAPBOX_ACCESS_TOKEN` is more commonly used by Mapbox tools that get the token from the environment. It’s less likely to collide with a local variable in a shell script or an environment variable expected by another tool.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -81,6 +85,44 @@ class CodingOperation<ResponceType : Codable, OptionsType : DirectionsOptions >
...
+            print("Failed to request response without coordinates.")
+            exit(1) }
+        
+        let options = RouteOptions(waypoints: waypoints, profileIdentifier: .automobileAvoidingTraffic)
+        options.includesSteps = includesSteps
+        options.allowsUTurnAtWaypoint = true
+        options.includesAlternativeRoutes = false
+        options.includesExitRoundaboutManeuver = true
+        options.roadClassesToAvoid = []
+        options.routeShapeResolution = .full
```
1ec5, line 101: These settings should come from decoding the configuration file as a RouteOptions object.

1ec5 (commented):
Suggested changelog entry:

* The `mapbox-directions-swift` command line tool now requests routes from the Mapbox Directions API if no input file is specified. ([#576](https://github.com/mapbox/mapbox-directions-swift/pull/576))

1ec5 (approved):

1ec5:
[The Linux build is failing](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/1380/workflows/3a5efc4c-2d6e-4730-a1f3-a2831939d1d5/jobs/5271):

```
[121/127] Compiling MapboxDirectionsCLI CodingOperation.swiftftK/root/project/Sources/MapboxDirectionsCLI/CodingOperation.swift:129:26: error: type 'URLSession' (aka 'AnyObject') has no member 'init'
        let urlSession = URLSession(configuration: .ephemeral)
                         ^~~~~~~~~~
/root/project/Sources/MapboxDirectionsCLI/CodingOperation.swift:129:53: error: cannot infer contextual base in reference to member 'ephemeral'
        let urlSession = URLSession(configuration: .ephemeral)
                                                   ~^~~~~~~~~
[122/127] Compiling MapboxDirectionsCLI main.swift/root/project/Sources/MapboxDirectionsCLI/CodingOperation.swift:129:26: error: type 'URLSession' (aka 'AnyObject') has no member 'init'
        let urlSession = URLSession(configuration: .ephemeral)
                         ^~~~~~~~~~
/root/project/Sources/MapboxDirectionsCLI/CodingOperation.swift:129:53: error: cannot infer contextual base in reference to member 'ephemeral'
        let urlSession = URLSession(configuration: .ephemeral)
                                                   ~^~~~~~~~~
```

To fix these errors, import the FoundationNetworking module (but only if FoundationNetworking can be imported).

PR #580, opened 2021-08-30
Convert from Directions API URL to RouteOptions

### LINKED ISSUE #476: Convert from Directions API URL to RouteOptions
This library should have an easy way to convert a Directions API request URL (or perhaps a URLRequest or array of URLQueryItems) to a RouteOptions or MatchOptions object. This conversion would be especially useful when implementing Handoff or Universal Links. It would also be a useful alternative to the JSON input file expected by the command line tool in #469.

Here’s a crude version I wrote once as part of an internal dogfooding application that supports Handoff and Universal Links:

```swift
extension RouteOptions {
    convenience init?(url: URL) {
        let pathComponents = url.pathComponents
        guard pathComponents.count == 6 && pathComponents[1] == "directions" && pathComponents[2] == "v5" else {
            return nil
        }
        
        let waypoints = url.deletingPathExtension().lastPathComponent
            .split(separator: ";")
            .map { $0.split(separator: ",", maxSplits: 1) }
            .map { CLLocationCoordinate2D(latitude: Double($0[1])!, longitude: Double($0[0])!) }
            .map { Waypoint(coordinate: $0) }
        let profileIdentifier = DirectionsProfileIdentifier(rawValue: pathComponents[3..<5].joined(separator: "/"))
        self.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
    }
}
```

/cc @mapbox/navigation-ios

jill-cardamon:
This PR fixes #476 building off the code provided in the snippet.

### DIFF
--- a/Sources/MapboxDirections/Directions.swift
+++ b/Sources/MapboxDirections/Directions.swift
@@ -570,6 +570,42 @@ open class Directions: NSObject {
         request.setValue(userAgent, forHTTPHeaderField: "User-Agent")
         return request
     }
+    
+    public func convertToRouteOptions(from url: URL) -> RouteOptions? {
+        var routeOptions: RouteOptions!
+//        guard let queryItems = URLComponents(string: url.absoluteString)?.queryItems else { return nil }
+//        var optionsJSON = JSONDictionary()
+//        for queryItem in queryItems {
+//            optionsJSON[queryItem.name] = queryItem.value
+//        }
+//
+//        print("!!! JSON: \(optionsJSON)")
+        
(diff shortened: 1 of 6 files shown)


MaximAlien:
@jill-cardamon, would be good to update this PR as well. Thanks!

1ec5:
Does #655 supersede this PR, or is there anything else that needs to be folded into it?

jill-cardamon:
@1ec5 it looks like it does. we can call that method in cli

1ec5:
> it looks like it does. we can call that method in cli

OK, tracking in #692.

1ec5:
As mentioned above, #655 takes care of converting from a URL to an options object. Now #728 hooks this conversion up to the command line tool, loosely based on the approach taken in this PR.

PR #587, opened 2021-09-06
Fix segment indices

bamx23:
There was an issue with `segmentRangesByStep` having the wrong range for the arrival step. Instead of being empty, it had the length of 2 as the arrival step in real responses has a geometry that is a line with the same two points (zero-length line).

cc: @1ec5


S2Ler:
@bamx23 Do we have an issue that will be fixed by fix PR?

bamx23:
> @bamx23 Do we have an issue that will be fixed by fix PR?

We do, but it's in a private repo, linked.

1ec5 (approved):
Suggested changelog blurb:

* Fixed an issue where `RouteLeg.segmentRangesByStep` contained a range that was off by one for the arrival step of a leg. ([#587](https://github.com/mapbox/mapbox-directions-swift/pull/587))

PR #588, opened 2021-09-08
Don't append service sku token if it doesn't match access token used …

S2Ler:
…in the Directions request

### DIFF
--- a/Sources/MapboxDirections/DirectionsCredentials.swift
+++ b/Sources/MapboxDirections/DirectionsCredentials.swift
@@ -26,7 +26,12 @@ public struct DirectionsCredentials: Equatable {
         #if !os(Linux)
         guard let mbx: AnyClass = NSClassFromString("MBXAccounts") else { return nil }
         guard mbx.responds(to: Selector(("serviceSkuToken"))) else { return nil }
-        return mbx.value(forKeyPath: "serviceSkuToken") as? String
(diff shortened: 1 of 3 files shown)


Sources/MapboxDirections/DirectionsCredentials.swift
```diff
@@ -26,7 +26,12 @@ public struct DirectionsCredentials: Equatable {
         #if !os(Linux)
         guard let mbx: AnyClass = NSClassFromString("MBXAccounts") else { return nil }
         guard mbx.responds(to: Selector(("serviceSkuToken"))) else { return nil }
-        return mbx.value(forKeyPath: "serviceSkuToken") as? String
+        guard let serviceSkuToken = mbx.value(forKeyPath: "serviceSkuToken") as? String,
+              let serviceAccessToken = mbx.value(forKeyPath: "serviceAccessToken") as? String,
```
Udumft, line 30: What should happen if `serviceSkuToken` is `nil` while `serviceAccessToken` is present?

Udumft, line 30: What should happen if `serviceSkuToken` is `nil` while `serviceAccessToken` is present?

S2Ler: return nil token.

Udumft (approved):

PR #591, opened 2021-09-08
Backward compatible serviceAccessToken

S2Ler:


Udumft (approved):

Sources/MapboxDirections/DirectionsCredentials.swift
```diff
@@ -24,14 +24,21 @@ public struct DirectionsCredentials: Equatable {
...
-              serviceAccessToken == accessToken else {
-                  return nil
-              }
-        return serviceSkuToken
+        guard let mbx: AnyClass = NSClassFromString("MBXAccounts"),
+              mbx.responds(to: Selector(("serviceSkuToken"))),
+              let serviceSkuToken = mbx.value(forKeyPath: "serviceSkuToken") as? String
+        else { return nil }
+
+        if mbx.responds(to: Selector(("serviceAccessToken"))) {
```
chezzdev, line 32: What's the use-case for having `serviceSkuToken` and not having `serviceAccessToken`?

S2Ler: Backward compatibility basically. For somebody who might use this library outside of the SDK. We will have `serviceAccessToken` in SDK for sure.

PR #593, opened 2021-09-20
Raw coding keys for DirectionsOptions and its subclasses match the query parameter names

### LINKED ISSUE #581: Raw coding keys for DirectionsOptions, RouteOptions, and MatchOptions should match the query parameter names
Raw coding keys for `DirectionsOptions`, `RouteOptions`, and `MatchOptions` should match the Directions API query parameter names to make encoding and decoding more straightforward. 

### Background

`DirectionsOptions` and its subclasses `RouteOptions` and `MatchOptions` currently adhere to Swift and Cocoa naming conventions. Server API particulars like parameter names are implementation details to developers, but cross-platform developers might encounter a learning curve going from mapbox-java to mapbox-directions-swift (or vice versa) due to inconsistent naming. 

### Notes and Next Steps

To ease this learning curve, raw coding keys for `DirectionsOptions`, `RouteOptions`, and `MatchOptions` should match the Directions API query parameter like we did for responses. Documentation should also reflect these changes. This change will also make it easier to encode and decode `DirectionsOptions` which would ease the implementation of #476.

/cc: @mapbox/navigation-ios

jill-cardamon:
Fixes #581 by adding raw coding keys for DirectionsOptions and RouteOptions and the corresponding tests. MatchOptions already had raw coding keys, so this class was not changed.

### DIFF
--- a/Tests/MapboxDirectionsTests/RouteOptionsTests.swift
+++ b/Tests/MapboxDirectionsTests/RouteOptionsTests.swift
@@ -73,6 +73,27 @@ class RouteOptionsTests: XCTestCase {
         XCTAssertEqual(routeOptions.includesExitRoundaboutManeuver, true)
         XCTAssertEqual(routeOptions.roadClassesToAvoid, .toll)
         XCTAssertEqual(routeOptions.refreshingEnabled, false)
+        
+        let encodedRouteOptions: Data = try! JSONEncoder().encode(routeOptions)
+        let optionsString: String = String(data: encodedRouteOptions, encoding: .utf8)!
+        
+        let unarchivedOptions: RouteOptions = try! JSONDecoder().decode(RouteOptions.self, from: optionsString.data(using: .utf8)!)
+        
+        XCTAssertNotNil(unarchivedOptions)
+        XCTAssertEqual(unarchivedOptions.profileIdentifier, routeOptions.profileIdentifier)
+        XCTAssertEqual(unarchivedOptions.includesSteps, routeOptions.includesSteps)
+        XCTAssertEqual(unarchivedOptions.shapeFormat, routeOptions.shapeFormat)
+        XCTAssertEqual(unarchivedOptions.routeShapeResolution, routeOptions.routeShapeResolution)
+        XCTAssertEqual(unarchivedOptions.attributeOptions, routeOptions.attributeOptions)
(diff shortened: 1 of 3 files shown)


Tests/MapboxDirectionsTests/RouteOptionsTests.swift
```diff
@@ -31,6 +31,50 @@ class RouteOptionsTests: XCTestCase {
...
+            "continue_straight": true,
+            "alternatives": false,
+            "roundabout_exits": true,
+            "exclude": ["toll"],
+            "enable_refresh": false
+        ]
+        
+        let routeOptionsData = try! JSONSerialization.data(withJSONObject: routeOptionsJSON, options: [])
+        var routeOptions: RouteOptions?
+        XCTAssertNoThrow(routeOptions = try JSONDecoder().decode(RouteOptions.self, from: routeOptionsData))
```
azarovalex, line 60: mb it's easier to use `try!` there, so all the asserts can use routeOptions without unwrapping an optional?

Tests/MapboxDirectionsTests/RouteOptionsTests.swift
```diff
@@ -31,6 +31,50 @@ class RouteOptionsTests: XCTestCase {
...
+            "voice_units": "imperial",
+            "banner_instructions": true,
+            "continue_straight": true,
+            "alternatives": false,
+            "roundabout_exits": true,
+            "exclude": ["toll"],
+            "enable_refresh": false
+        ]
+        
+        let routeOptionsData = try! JSONSerialization.data(withJSONObject: routeOptionsJSON, options: [])
```
azarovalex, line 58: Should we test `JSONEncoder` as well?

azarovalex (approved):

Tests/MapboxDirectionsTests/RouteOptionsTests.swift
jill-cardamon: I'll do that. Thank you for the suggestion!

Tests/MapboxDirectionsTests/RouteOptionsTests.swift
jill-cardamon: It wouldn't hurt to road trip.

PR #598, opened 2021-09-30
Expose include-hov/hot parameters in `RouteOptions`

### LINKED ISSUE #592: Expose include-hov/hot parameters in `RouteOptions`
We need to expose the `include-hov` and `include-hot` parameters in `RouteOptions`.

### Background
`RouteOptions` already has `exclude`s parameters that can be used to exclude types of road classes like `toll`s. These parameters can be thought of as, "we will route onto tolls by default; to avoid tolls you have to opt-out using `exclude=toll`."

### Notes and Next Steps
The directions API is proposing a new `include` key with the following possible values:
```
include=hov2
include=hov3
include=hot
```
These parameters can be thought of as, "we will avoid routing onto hov/hot by default; if you want to route onto hov/hot you have to opt-in using `include=hov/hot`." The `include` parameter will only be available for `driving` and `driving-traffic` profiles. Other profiles will return an error.

cc/ @mapbox/navigation-ios @Guardiola31337

jill-cardamon:
Fixes #592.
This PR exposes the new `include hov/hot` parameters by adding `RouteOptions.roadClassesToAllow` property (and the corresponding `RoadClass`es) and its serialization to url parameters.

cc/ @mapbox/navigation-ios @Guardiola31337 @zugaldia


Sources/MapboxDirections/RoadClasses.swift
```diff
@@ -32,7 +32,7 @@ public struct RoadClasses: OptionSet, CustomStringConvertible {
     /**
      The user must travel this segment of the route by ferry.
      
-     The user should verify that the ferry is in operation. For driving and cycling directions, the user should also verify that his or her vehicle is permitted onboard the ferry.
+     The user should verify that the ferry is in operation. For driving and cycling directions, the user should also verify that their vehicle is permitted onboard the ferry.
```
1ec5, line 35: 👍

Sources/MapboxDirections/RoadClasses.swift
```diff
@@ -43,6 +43,21 @@ public struct RoadClasses: OptionSet, CustomStringConvertible {
...
+    
+    /**
+     The road segment is a HOV-3 road that requires a minumum of three vehicle occupants.
+    */
+    public static let hov3 = RoadClasses(rawValue: 1 << 7)
+    
+    /**
+     The road segment is a HOT road that is tolled if the user's vehicle does not meet the minimum occupant requirement.
+    */
+    public static let hot = RoadClasses(rawValue: 1 << 8)
```
1ec5, line 59: The term “HOT” is traffic engineering/management jargon that isn’t widely known among software developers in the U.S., let alone other countries. `hot` is also prone to be misinterpreted and look out of place as the adjective. Spell out `highOccupancyToll` and link “high-occupancy toll road” to [this Wikipedia article](https://en.wikipedia.org/wiki/High-occupancy_toll_lane). The article is about HOT lanes and express toll lanes, so we should clarify that the option is about HOT-only roads, not mere HOT lanes.

Sources/MapboxDirections/RoadClasses.swift
```diff
@@ -43,6 +43,21 @@ public struct RoadClasses: OptionSet, CustomStringConvertible {
      */
     public static let tunnel = RoadClasses(rawValue: 1 << 5)
     
+    /**
+     The road segment is a HOV-2 road that requires a minimum of two vehicle occupants.
+    */
+    public static let hov2 = RoadClasses(rawValue: 1 << 6)
```
1ec5, line 49: The term “HOV” is a little more widely known than “HOT” but still somewhat obscure, since some countries and U.S. states use other terminology, like “carpool lane”. For consistency with `highOccupancyToll`, spell these cases out as `highOccupancyVehicle{2,3}` and link to [this Wikipedia article](https://en.wikipedia.org/wiki/High-occupancy_vehicle_lane) or [this OpenStreetMap Wiki article](https://wiki.openstreetmap.org/wiki/Key:hov). Similarly, we should clarify that the option pertains to HOV-only roads, not mere HOV lanes.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -122,6 +126,13 @@ open class RouteOptions: DirectionsOptions {
      */
     open var roadClassesToAvoid: RoadClasses = []
     
+    /**
+     The route classes that the calculated routes will allow.
+     
+     This property has no effect unless the profile identifier is set to `DirectionsProfileIdentifier.automobile` or `DirectionsProfileIdentifier.automobileAvoidingTraffic`.
+    */
+    open var roadClassesToAllow: RoadClasses = []
```
1ec5, line 134: With this name, the developer may assume this property must contain a road class in order for the route to traverse roads of that class, even if in reality the routing profile would include it by default. If the developer redundantly sets this property to a value such as `RoadClasses.motorway`, will the Directions API silently ignore the redundant value, or will it return an error? If the API only allows certain values, then we should, split the enumeration in two, special-case and ignore the allowed-by-default classes, or document which cases are allowed with which property. None of the approaches is particularly attractive, but perhaps a future Directions API enhancement could allow us to design a more ergonomic, predictable client-side API.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * Added the `Waypoint.allowsSnappingToClosedRoad` property to allow snapping the waypoint’s location to a closed part of a road. ([#583](https://github.com/mapbox/mapbox-directions-swift/pull/583))
 * Added `AttributeOptions.numericCongestionLevel`, `RouteLeg.segmentNumericCongestionLevels`, `RouteLeg.Attributes.segmentNumericCongestionLevels` and `NumericCongestionLevel` to support receiving the numeric value for congestion level along each segment of a `RouteLeg`. ([#575](https://github.com/mapbox/mapbox-directions-swift/pull/575))
 * Fixed an issue where `RouteLeg.segmentRangesByStep` contained a range that was off by one for the arrival step of a leg. ([#587](https://github.com/mapbox/mapbox-directions-swift/pull/587))
+* Added a `RouteOption.roadClassesToAllow` property that allows hov-2, hov-3, or hot roads. ([#598](https://github.com/mapbox/mapbox-directions-swift/pull/598))
```
1ec5, line 19:
```suggestion
* Added a `RouteOptions.roadClassesToAllow` property that enables the resulting routes to use HOV-only roads or high-occupancy toll roads. ([#598](https://github.com/mapbox/mapbox-directions-swift/pull/598))
```

Sources/MapboxDirections/RoadClasses.swift
jill-cardamon: Oh that makes sense. I will fix this for hot and hov.

Sources/MapboxDirections/RouteOptions.swift
jill-cardamon: 😞 I will fix this one. I will check how the directions api behaves and make a decision from there.

CHANGELOG.md
jill-cardamon: thanks for catching that typo!

jill-cardamon:
my comment isn't showing up here in the conversation but it's in the file...to answer your question,

> Are we settling on a single enumeration or two?

Settled on a single enumeration with a [check](https://github.com/mapbox/mapbox-directions-swift/blob/5c665a1e7acf940d3f60df950be6e125fa31cd70/Sources/MapboxDirections/RouteOptions.swift#L226) when creating the urlQueryItems. I should add a check for `roadClassesToAvoid` as well. I'll add a comments to make it clear. 

Do you think `Intersection.outletRoadClasses` would contain any of these `hov/hot` values?

PR #601, opened 2021-09-30
Fix publish docs script

S2Ler:


scripts/publish-documentation.sh
```diff
@@ -21,7 +21,7 @@ git checkout $RELEASE_BRANCH
 git checkout $VERSION
 
 step "Installing dependencies…"
-./scripts/wcarthage.sh bootstrap --cache-builds
+./scripts/wcarthage.sh bootstrap --cache-builds --use-xcframeworks
```
MaximAlien, line 24: Do we still need `wcarthage.sh`? I think it was added as a workaround for beta versions of Xcode. I guess carthage can be used directly now.

MaximAlien (approved):

scripts/publish-documentation.sh
S2Ler: 1. Yeah, we shall make this file trivial similar to https://github.com/mapbox/mapbox-navigation-ios/blob/main/scripts/wcarthage.sh
2. I decided to keep this file so that users scripts wouldn't broke. We can remove it though and update all the docs.

S2Ler: https://github.com/mapbox/mapbox-directions-swift/issues/602

PR #606, opened 2021-10-05
Swift Argument Parser

### LINKED ISSUE #600: Replace SwiftCLI with Swift Argument Parser
#469 uses [SwiftCLI](https://github.com/jakeheis/SwiftCLI/), but [Swift Argument Parser](https://github.com/apple/swift-argument-parser/) has an even more elegant API that would facilitate future enhancements to this tool.

Technically, replacing the dependency would be a backwards-incompatible change. If we can take care of the migration by v2.0.0, that would be great. However, we probably have some additional leeway. I don’t expect that any developer would be using the command line tool as a helper process for another command line tool, and even if they did, that other command line tool can still depend on SwiftCLI; it just couldn’t depend on an incompatible 0._x_ version of Swift Argument Parser.

Thanks to @azarovalex for the suggestion.

/cc @mapbox/navigation-ios

Udumft:
Resolves #600 
This PR replaces `jakeheis/SwiftCLI` dependency with `apple/swift-argument-parser`. Updated the source code. Removed Cathage support for `mapbox-directions-swift` since `swift-argument-parser` does not support it.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,3 +1,3 @@
 github "raphaelmor/Polyline" ~> 5.0
 github "mapbox/turf-swift" "v2.0.0-rc.1"
-github "jakeheis/SwiftCLI" ~> 6.0.3
+github "apple/swift-argument-parser" ~> 1.0.0
--- a/Package.resolved
+++ b/Package.resolved
@@ -20,12 +20,12 @@
         }
       },
       {
-        "package": "SwiftCLI",
-        "repositoryURL": "https://github.com/jakeheis/SwiftCLI",
+        "package": "swift-argument-parser",
+        "repositoryURL": "https://github.com/apple/swift-argument-parser",
         "state": {
           "branch": null,
-          "revision": "2816678bcc37f4833d32abeddbdf5e757fa891d8",
-          "version": "6.0.2"
+          "revision": "d2930e8fcf9c33162b9fcc1d522bc975e2d4179b",
+          "version": "1.0.1"
         }
       },
       {
@@ -33,8 +33,8 @@
         "repositoryURL": "https://github.com/mapbox/turf-swift.git",
         "state": {
           "branch": null,
-          "revision": "555458bd67acce7322c513ddd3647c9a904f4197",
-          "version": "2.0.0-rc.1"
+          "revision": "007896bb945432d4683e114fcf2f089edff4ceeb",
+          "version": "2.0.0-rc.2"
         }
       }
     ]
--- a/Sources/MapboxDirectionsCLI/ProcessCommand.swift
+++ b/Sources/MapboxDirectionsCLI/ProcessCommand.swift
@@ -1,38 +1,14 @@
 
 import Foundation
 import MapboxDirections
-import SwiftCLI
-
 
 private let BogusCredentials = DirectionsCredentials(accessToken: "pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede")
 
-class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > : Command {
+class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > {
     
     // MARK: - Parameters
     
-    var name = "process"
-    
-    @Key("-i", "--input", description: "Filepath to the input JSON.")
-    var inputPath: String?
-    
-    @Key("-c", "--config", description: "Filepath to the JSON, containing serialized Options data.")
-    var configPath: String?
-    
-    @Key("-o", "--output", description: "[Optional] Output filepath to save the conversion result. If no filepath provided - will output to the shell.")
-    var outputPath: String?
-    
-    @Key("-f", "--format", description: "Output format. Supports `text` and `json` formats. Defaults to `text`")
-    var outputFormat: OutputFormat?
-    
-    enum OutputFormat: String, ConvertibleFromString {
-        case text
-        case json
-    }
-    
-    var customShortDescription: String = ""
-    var shortDescription: String {
-        return customShortDescription
-    }
+    let options: ProcessingOptions
     
     // MARK: - Helper methods
     
@@ -45,8 +21,8 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
     private func processOutput(_ data: Data) {
         var outputText: String = ""
         
-        switch outputFormat {
-        case .text, .none:
+        switch options.outputFormat {
+        case .text:
             outputText = String(data: data, encoding: .utf8)!
         case .json:
             if let object = try? JSONSerialization.jsonObject(with: data, options: []),
@@ -55,7 +31,7 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
             }
         }
         
-        if let outputPath = outputPath {
+        if let outputPath = options.outputPath {
             do {
                 try outputText.write(toFile: NSString(string: outputPath).expandingTildeInPath,
                                      atomically: true,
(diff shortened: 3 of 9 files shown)


Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,65 @@
...
+    var inputPath: String?
+    
+    @ArgumentParser.Option(name: [.short, .customLong("config")], help: "Filepath to the JSON, containing serialized Options data.")
+    var configPath: String?
+    
+    @ArgumentParser.Option(name: [.short, .customLong("output")], help: "[Optional] Output filepath to save the conversion result. If no filepath provided - will output to the shell.")
+    var outputPath: String?
+    
+    @ArgumentParser.Option(name: [.customShort("f"), .customLong("format")], help: "Output format. Supports `text` and `json` formats.")
+    var outputFormat: OutputFormat = .text
```
jill-cardamon, line 20: Why does this option use `.customShort` instead of `.short`?

bamx23: As it would conflict with `-o` for `outputPath`

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,65 @@
...
-                                                                 shortDescription: "Command to process Map Matching Data")
-let route = ProcessCommand< MapboxDirections.RouteResponse, RouteOptions>(name: "route",
-                                                                          shortDescription: "Command to process Routing Data")
+struct ProcessingOptions: ParsableArguments {
+    
+    @ArgumentParser.Option(name: [.short, .customLong("input")], help: "Filepath to the input JSON.")
+    var inputPath: String?
+    
+    @ArgumentParser.Option(name: [.short, .customLong("config")], help: "Filepath to the JSON, containing serialized Options data.")
+    var configPath: String?
```
bamx23, line 14: Do we still want these two arguments to be optional?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -1,38 +1,14 @@
 
 import Foundation
 import MapboxDirections
-import SwiftCLI
-
 
 private let BogusCredentials = DirectionsCredentials(accessToken: "pk.feedCafeDadeDeadBeef-BadeBede.FadeCafeDadeDeed-BadeBede")
 
-class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > : Command {
+class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > {
```
bamx23, line 7: Can we rename the class as it looks like a `ParsableCommand` but not like that anymore

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,65 @@
...
+}
 
-CLI(name: "mapbox-directions-swift",
-    description: "'mapbox-directions-swift' is a command line tool, designed to round-trip an arbitrary, JSON-formatted Directions or Map Matching API response through model objects and back to JSON.",
-    commands: [route, match]).goAndExit()
+struct Process: ParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "mapbox-directions-swift",
+        abstract: "'mapbox-directions-swift' is a command line tool, designed to round-trip an arbitrary, JSON-formatted Directions or Map Matching API response through model objects and back to JSON.",
+
```
bamx23, line 32:
```suggestion
        abstract: "'mapbox-directions-swift' is a command line tool, designed to round-trip an arbitrary, JSON-formatted Directions or Map Matching API response through model objects and back to JSON.",
```

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,65 @@
...
+import ArgumentParser
 
 
-let match = ProcessCommand< MapMatchingResponse, MatchOptions > (name: "match",
-                                                                 shortDescription: "Command to process Map Matching Data")
-let route = ProcessCommand< MapboxDirections.RouteResponse, RouteOptions>(name: "route",
-                                                                          shortDescription: "Command to process Routing Data")
+struct ProcessingOptions: ParsableArguments {
+    
+    @ArgumentParser.Option(name: [.short, .customLong("input")], help: "Filepath to the input JSON.")
```
bamx23, line 10: Do you need to add `ArgumentParser` module name everywhere?

Sources/MapboxDirectionsCLI/ProcessCommand.swift
```diff
@@ -70,16 +46,15 @@ class ProcessCommand<ResponceType : Codable, OptionsType : DirectionsOptions > :
...
+        self.options = options
     }
     
     // MARK: - Command implementation
     
     func execute() throws {
-        guard let inputPath = inputPath else { exit(1) }
-        guard let configPath = configPath else { exit(1) }
+        guard let inputPath = options.inputPath else { exit(1) }
+        guard let configPath = options.configPath else { exit(1) }
```
bamx23, line 57: Check [Validation](https://github.com/apple/swift-argument-parser/blob/main/Sources/ArgumentParser/Documentation.docc/Articles/Validation.md) guide, maybe it's better to use it to make errors more descriptive. But also, we can simply require these path arguments.

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,65 @@
...
+    @ArgumentParser.Option(name: [.short, .customLong("config")], help: "Filepath to the JSON, containing serialized Options data.")
+    var configPath: String?
+    
+    @ArgumentParser.Option(name: [.short, .customLong("output")], help: "[Optional] Output filepath to save the conversion result. If no filepath provided - will output to the shell.")
+    var outputPath: String?
+    
+    @ArgumentParser.Option(name: [.customShort("f"), .customLong("format")], help: "Output format. Supports `text` and `json` formats.")
+    var outputFormat: OutputFormat = .text
+    
+    enum OutputFormat: String, ExpressibleByArgument {
```
bamx23, line 22: I would recommend adding `CaseItarable` here to support [these extensions](https://github.com/apple/swift-argument-parser/blob/d2930e8fcf9c33162b9fcc1d522bc975e2d4179b/Sources/ArgumentParser/Parsable%20Types/ExpressibleByArgument.swift#L47-L61)

Cartfile
```diff
@@ -1,3 +1,3 @@
 github "raphaelmor/Polyline" ~> 5.0
 github "mapbox/turf-swift" "v2.0.0-rc.1"
-github "jakeheis/SwiftCLI" ~> 6.0.3
+github "apple/swift-argument-parser" ~> 1.0.0
```
bamx23, line 3: Should we support minor changes here?
```suggestion
github "apple/swift-argument-parser" ~> 1.0
```

Package.resolved
```diff
@@ -20,21 +20,21 @@
...
       },
       {
         "package": "Turf",
         "repositoryURL": "https://github.com/mapbox/turf-swift.git",
         "state": {
           "branch": null,
-          "revision": "555458bd67acce7322c513ddd3647c9a904f4197",
-          "version": "2.0.0-rc.1"
+          "revision": "007896bb945432d4683e114fcf2f089edff4ceeb",
+          "version": "2.0.0-rc.2"
```
bamx23, line 37: Should we discard non-related dependency changes?

Udumft: Well, I thought it will proc eventually to anyone who updates packages, so instead of moving it into special PR, why not update it when it pops out

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,82 @@
...
+        subcommands: [Match.self, Route.self]
+    )
+    
+    fileprivate static func validateInput(_ options: ProcessingOptions) throws {
+        guard FileManager.default.fileExists(atPath: options.inputPath) else {
+            throw ValidationError("Input JSON file `\(options.inputPath)`does not exist.")
+        }
+        
+        guard FileManager.default.fileExists(atPath: options.configPath) else {
+            throw ValidationError("Options JSON file `\(options.configPath)`does not exist.")
```
bamx23, line 41: NIT
```suggestion
            throw ValidationError("Input JSON file `\(options.inputPath)` does not exist.")
        }
        
        guard FileManager.default.fileExists(atPath: options.configPath) else {
            throw ValidationError("Options JSON file `\(options.configPath)` does not exist.")
```

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -2,14 +2,82 @@
...
+extension Command {
+    struct Match: ParsableCommand {
+        static var configuration =
+            CommandConfiguration(commandName: "match",
+                                    abstract: "Command to process Map Matching Data.")
+        
+        @ArgumentParser.OptionGroup var options: ProcessingOptions
+        
+        mutating func validate() throws {
+            try Command.validateInput(options)
```
bamx23, line 55: It might be better to have an extension for `ProcessingOptions` with this function.

bamx23 (approved):

PR #616, opened 2021-10-20
Waypoint.snappedDistance

### LINKED ISSUE #333: Add snapped distance property to Waypoint

Udumft:
Resolves #333 
PR adds `Waypoint.snappedDistance` property.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v2.x
 
-* Added `Waypoint.snappedDistance` property. ([#616](https://github.com/mapbox/mapbox-directions-swift/pull/616))
+* Added `Waypoint.snappedDistance` property to support receiving the waypoint's straight-line distance to the location it was snapped to in the `RouteResponse`. ([#616](https://github.com/mapbox/mapbox-directions-swift/pull/616))
 
 ## v2.0.0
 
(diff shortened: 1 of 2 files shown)


Udumft:
Corrected. Thanks for suggestions!

PR #626, opened 2021-11-08
Matrix API Integration

### LINKED ISSUE #419: Matrix API
This library should be able to make requests to and handle responses from the [Matrix API](https://docs.mapbox.com/api/navigation/#matrix).

/cc @mapbox/navigation-ios

jill-cardamon:
Fixes #419.

This PR adds a `Matrix` object to access the Matrix API. Integration is similar to `Directions` and `Isochrones`.
Needs to be rebased on top of main.

### DIFF
--- a/Sources/MapboxDirections/MatrixOptions.swift
+++ b/Sources/MapboxDirections/MatrixOptions.swift
@@ -38,7 +38,7 @@ public class MatrixOptions: Codable {
     /**
      Attribute options for the matrix.
 
-     Empty `attributeOptions` will result in default values assumed.
+     Only `distance` and `expectedTravelTime` are supported. Empty `attributeOptions` will result in default values assumed.
      */
     public var attributeOptions: AttributeOptions = []
     
@@ -81,7 +81,7 @@ public class MatrixOptions: Codable {
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encode(profileIdentifier, forKey: .profileIdentifier)
-        try container.encode(attributeOptions.queryDescription, forKey: .attributeOptions)
+        try container.encode(attributeOptions, forKey: .attributeOptions)
         try container.encodeIfPresent(destinations, forKey: .destinations)
         try container.encodeIfPresent(sources, forKey: .sources)
     }
@@ -89,9 +89,7 @@ public class MatrixOptions: Codable {
     public required init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         profileIdentifier = try container.decode(ProfileIdentifier.self, forKey: .profileIdentifier)
-        if let annotations = try container.decodeIfPresent(String.self, forKey: .attributeOptions) {
-            attributeOptions = Set<AttributeOption>(from: annotations)
-        }
+        self.attributeOptions = try container.decodeIfPresent(AttributeOptions.self, forKey: .attributeOptions) ?? []
         let destinations = try container.decodeIfPresent([Waypoint].self, forKey: .destinations) ?? []
         let sources = try container.decodeIfPresent([Waypoint].self, forKey: .sources) ?? []
         self.waypointsData = .init(sources: sources,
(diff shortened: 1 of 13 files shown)


MaximAlien:
@jill-cardamon, it'd be good to update this PR as it's pretty outdated.

MaximAlien:
@1ec5, could you please take a look when you have time?

Sources/MapboxDirections/MatrixOptions.swift
```diff
@@ -0,0 +1,200 @@
...
+    
+    /**
+     A string specifying the primary mode of transportation for the contours.
+    */
+    public var profileIdentifier: ProfileIdentifier
+    
+    /**
+     An array of `Waypoints` objects representing locations that will be in the matrix.
+     */
+    public var waypoints: [Waypoint]
```
1ec5, line 28: Should we reuse `Waypoint` for the Matrix API? Judging from [the public API documentation](https://docs.mapbox.com/api/navigation/matrix/), many of the options in `Waypoint` don’t apply to the Matrix API, either as input or output.

Sources/MapboxDirections/MatrixOptions.swift
```diff
@@ -0,0 +1,200 @@
...
+     By default, all waypoints in the `waypoints` array are used as destinations.
+     */
+    public var destinations: [Int]?
+    
+    /**
+     The coordinates at a given index in the `waypoints` array that should be used as sources.
+     
+     By default, all waypoints in the `waypoints` array are used as sources.
+     */
+    public var sources: [Int]?
```
1ec5, line 49: If we make a separate `Matrix.Waypoint` type, then it could have `isSource` and `isDestination` properties to avoid the need for maintaining parallel arrays of indices, which introduce the potential for out-of-bounds errors. Otherwise, a subclass of `Waypoint` might still be valuable as a place to put these properties.

Sources/MapboxDirections/MatrixOptions.swift
```diff
@@ -0,0 +1,200 @@
...
+extension MatrixOptions {
+    
+    /**
+     Attributes are metadata information for requesting matricies.
+     
+     Options selected affect the contents of the resulting `MatrixResponse`.
+     */
+    public typealias AttributeOptions = Set<AttributeOption>
+    
+    public struct AttributeOption: Hashable, RawRepresentable {
```
1ec5, line 146: This public type should have a documentation comment.

1ec5, line 146: I assume this is a separate type than the main `AttributeOptions` because the Matrix API only supports a limited number of `annotations` values. However, some of the `AttributeOptions` values are already limited to one profile or another, so I don’t think it would be entirely surprising if we were to reuse that type but document some values like `congestionLevel` as applying only to the Directions and Map Matching APIs.

The other difference seems to be that `Matrix.AttributeOptions` is a `Set` rather than an `OptionSet`. In hindsight, this would’ve been a pretty good move for `AttributeOptions` in v1.0 or v2.0, but I wonder if adopting that approach only for the Matrix API would prove confusing.

Sources/MapboxDirections/MatrixOptions.swift
Udumft: My thought was that using same `Waypoint` would be useful when after requesting the Matrix, user will pick these same waypoints to build an actual route.

Sources/MapboxDirections/MatrixOptions.swift
Udumft: Re-worked this approach to have separate arrays of waypoints for sources and destinations.

Sources/MapboxDirections/MatrixOptions.swift
Udumft: Not actual anymore

Udumft: Updated original AttributeOptions docs and re-used them for Matrix.

Udumft:
Detected breaking changes are not related to current PR:
```
  💔 API breakage: enumelement VisualInstruction.Component.guidanceView has declared type change from (MapboxDirections.VisualInstruction.Component.Type) -> (MapboxDirections.GuidanceViewImageRepresentation, MapboxDirections.VisualInstruction.Component.TextRepresentation) -> MapboxDirections.VisualInstruction.Component to (MapboxDirections.VisualInstruction.Component.Type) -> (MapboxDirections.GuidanceViewImageRepresentation, MapboxDirections.VisualInstruction.Component.TextRepresentation, MapboxDirections.VisualInstruction.Component.GuidanceViewKind?) -> MapboxDirections.VisualInstruction.Component
```

Sources/MapboxDirections/MatrixOptions.swift
```diff
@@ -0,0 +1,213 @@
...
+                        destinationsIndices.insert(source.offset)
+                        break
+                    }
+                }
+            }
+            
+            destinationsIndices.insert(integersIn: sources.endIndex..<(sources.endIndex + destinations.count))
+            
+            var sum = sources
+            sum.append(contentsOf: destinations)
```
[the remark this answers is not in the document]
Udumft: The way it works is that it puts all the sources at the beginning of the array while also removing duplicate waypoints in `destinations`. After that, what is left in `destinations` is appended to the result. In the end, there will be only unique waypoints left. There are `testRequest()` and `testWaypointParameters()` that also checks that waypoints are composed correctly and you can see there how `sources` and `destinations` arrays are handled.

kried (approved):

PR #627, opened 2021-11-10
RoadClasses restrictions violation.

Udumft:
Sometimes Directions may ignore `RouteOptions.roadClassesToAvoid` setting and violated such restrictions. This PR adds method to check for such cases and provides convenience type for searching the result.


Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,69 @@
...
+     */
+    public var legIndex: Int
+    /**
+     Index of a `RouteStep` inside `RouteLeg` where violation occured.
+     */
+    public var stepIndex: Int
+    /**
+     Index of an `Intersection` inside `RouteStep` where violation occured.
+     */
+    public var intersectionIndex: Int
```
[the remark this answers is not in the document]
Udumft: I used indices instead of object references because this is how we address entities in Nav SDK's RouteProgress for example. Also, having index it is easy to get an actual object using current `Route`, but vice versa is more complicated. I also wanted to put full indices path to allow user maximum precision in case he needs it.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -26,6 +26,49 @@ public struct RouteResponse {
...
     public var created: Date = Date()
+    
+    /**
+     Managed of `RoadClasses` restrictions specified to `RouteOptions.roadClassesToAvoid` which were violated during route calculation.
+     
+     Routing engine may still utilize `RoadClasses` meant to be avoided in cases when routing is impossible otherwise.
+     
+     Resulting array is in the same order as `routes`, showing exact `RoadClasses` restrictions were ignored for each particular route at specific leg/step/intersection. `nil` and empty return arrays correspond to `nil` and empty `routes` array.
+     */
+    public var ignoredRoadClassesAvoidance: [RouteRoadClassesViolations]? {
```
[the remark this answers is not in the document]
Udumft: Sorry I don't seem to understand the idea. Current feature only detects when RoadClasses were meant to be avoided but still got included into the route(s). What do you mean by adding a new property?

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -26,6 +26,49 @@ public struct RouteResponse {
...
+        
+        var result = [RouteRoadClassesViolations]()
+        let avoidedClasses = routeOptions.roadClassesToAvoid
+        
+        for route in routes {
+            var violations = [RoadClassExclusionViolation]()
+            
+            for (legIndex, leg) in route.legs.enumerated() {
+                for (stepIndex, step) in leg.steps.enumerated() {
+                    for (intersectionIndex, intersection) in (step.intersections ?? []).enumerated() {
```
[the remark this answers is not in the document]
Udumft: I think you are right. We can do this calculation during `init`s. The only variable value is `RouteResponse.routes` so it should be the only `didSet` that should be tracked.
We could use `decoder.userInfo` not only for getting decoding attributes but also for accumulating found violations. But there's a problem that when we decode an Intersection, we can check if it violates the RoadClasses, but we cannot mark what index this intersection has. So we'll have to re-iterate whole graph to build indices, unless we decide to use object references instead. But I would prefer to use index-based addressing as described in [this comment](https://github.com/mapbox/mapbox-directions-swift/pull/627#discussion_r748080527).

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,69 @@
...
+     - parameter legIndex: Index of a leg inside related `Route`to search in.
+     - parameter stepIndex: Index of a step inside given `Route`'s leg.
+     - parameter intersectionIndex: Index of an intersection inside given `Route`'s leg and step.
+     - returns: Array of `RoadClassExclusionViolation` under given indicies.
+     
+     Passing `nil` as one of the indicies will result in searching for all violations at given `legIndex`/`stepIndex`.
+     - note: Searching by `intersectionIndex` requires non-nil `stepIndex`.
+     */
+    public func violations(at legIndex: Int, stepIndex: Int? = nil, intersectionIndex: Int? = nil) -> [RoadClassExclusionViolation] {
+        assert(!(stepIndex == nil && intersectionIndex != nil), "It is forbidden to select `intersectionIndex` without specifying `stepIndex`.")
```
S2Ler, line 32: The invariant can be embedded in the type-system. For example by using two methods:
```swift
public func violations(at legIndex: Int, stepIndex: Int, intersectionIndex: Int) -> [RoadClassExclusionViolation]
```

```swift
public func violations(at legIndex: Int, stepIndex: Int? = nil) -> [RoadClassExclusionViolation]
```

That way it will be impossible for developers to misuse the method and simplify the methods.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,69 @@
+
+import Foundation
+
+/**
+ Description of `RoadClasses` that were meant to be avoided during routing but still got used in a `Route`.
+ 
+ Such ignoring may be done by the engine if it cannot provide a route otherwise.
```
S2Ler, line 7: I suggest to expand this comment a bit more. Questions raised in my head when I read this:
- What is the engine?
- Such ignoring? There is no mention of the ignoring in the doc comment except one this place. 
- "if it cannot provide a route otherwise", very hard to read. Maybe "otherwise" is a superfluous word. @1ec5 is it just me?

Tests/MapboxDirectionsTests/RouteResponseTests.swift
```diff
@@ -62,4 +62,33 @@ class RouteResponseTests: XCTestCase {
...
+        }
+    
+        let options = RouteOptions(coordinates: [.init(latitude: 14.758522,
+                                                       longitude: 55.193541),
+                                                 .init(latitude: 12.54072,
+                                                       longitude: 55.685213)])
+        options.roadClassesToAvoid = [.ferry, .toll]
+        let decoder = JSONDecoder()
+        decoder.userInfo[.options] = options
+        decoder.userInfo[.credentials] = DirectionsCredentials(accessToken: "foo", host: URL(string: "https://test.website")!)
```
S2Ler, line 83: There are `BogusCredentials` and `BogusToken` constants in the tests. Makes sense to extend `BogusCredentials` with a host and use it here.

Tests/MapboxDirectionsTests/RouteResponseTests.swift
```diff
@@ -62,4 +62,33 @@ class RouteResponseTests: XCTestCase {
...
+        var response: RouteResponse?
+        XCTAssertNoThrow(response = try decoder.decode(RouteResponse.self, from: fixtureData))
+        
+        let roadClassesViolations = response!.ignoredRoadClassesAvoidance
```
S2Ler, line 87: This code will crash if the code on line 85 throws. Better to guard optionals and use XCTFail. Similar thing on the next lines.

Sources/MapboxDirections/RouteResponse.swift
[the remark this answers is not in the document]
1ec5: I assumed you had included “avoidance” in the name with the expectation that we might need a parallel `ignoredRoadClassesAllowance` property for when `roadClassesToAllow` gets violated. So pushing a property down into each `RoadClassExclusionViolation` would allow us to combine the two properties into one in that future scenario. In other words, I don’t think it’s necessary to name this property so specifically; `roadClassViolations` is adequate.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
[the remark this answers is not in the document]
1ec5: What about keeping the current structure but replacing the Route reference with a route index? It wouldn’t be as convenient in some respects, but considering that `route` is currently mutable, there would be many opportunities for the object model to become inconsistent with itself.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
1ec5: If something in `roadClassesToAvoid` prevents the origin or destination from connecting to the main road network, the Directions API will temporarily ignore the road class constraints in order to reach the main road network.

Udumft: thanks for suggestion!

Sources/MapboxDirections/RouteRoadClassesViolations.swift
Udumft: Good idea. Code updated

Sources/MapboxDirections/RouteResponse.swift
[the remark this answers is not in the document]
Udumft: Got it. Renamed the property now. If we would like to expand this feature that way later, we would just extend `RouteRoadClassesViolations` to contain new data.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Added the `RouteOptions.maximumHeight` and `RouteOptions.maximumWidth` properties for ensuring that the resulting routes can accommodate a vehicle of a certain size. ([#623](https://github.com/mapbox/mapbox-directions-swift/pull/623))
 * The `DirectionsPriority` struct now conforms to the `Codable` protocol. ([#623](https://github.com/mapbox/mapbox-directions-swift/pull/623))
 * Fixed an issue where the `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` properties were excluded from the encoded representation of a `RouteOptions` object. ([#623](https://github.com/mapbox/mapbox-directions-swift/pull/623))
+* Added `RouteResponse.ignoredRoadClassesAvoidance` to verify if some of the specified `RouteOptions.roadClassesToAvoid` restrictions were violated during route calculation. Added `RouteRoadClassesViolations` and `RoadClassExclusionViolation` types to identify such points and provide quick lookup. ([#627](https://github.com/mapbox/mapbox-directions-swift/pull/627))
```
1ec5, line 14:
```suggestion
* Added the `RouteResponse.roadClassViolations` property, which indicates any requested `RouteOptions.roadClassesToAvoid` values that could not be satisfied when calculating the routes. ([#627](https://github.com/mapbox/mapbox-directions-swift/pull/627))
```

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,85 @@
...
+     Filters `violations` to search for specific leg, step and intersection.
+     
+     - parameter legIndex: Index of a leg inside related `Route`to search in.
+     - parameter stepIndex: Index of a step inside given `Route`'s leg.
+     - parameter intersectionIndex: Index of an intersection inside given `Route`'s leg and step.
+     - returns: Array of `RoadClassExclusionViolation` under given indicies.
+     
+     Passing `nil` as `intersectionIndex` will result in searching for all intersections of given step.
+     */
+    public func violations(at legIndex: Int, stepIndex: Int, intersectionIndex: Int?) -> [RoadClassExclusionViolation] {
```
1ec5, line 43: Remove the `at` label in favor of the default `legIndex`: there doesn’t seem to be anything about the leg index that qualifies for “at” more than the step and intersection indices do.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,85 @@
...
+        return fetchViolations(at: legIndex, stepIndex: stepIndex, intersectionIndex: intersectionIndex)
+    }
+    
+    private func fetchViolations(at legIndex: Int, stepIndex: Int? = nil, intersectionIndex: Int? = nil) -> [RoadClassExclusionViolation] {
```
1ec5, line 47: “Fetch” implies that something is being requested over the network. This is just about returning a filtered list, so `filteredViolations(…)` would be less misleading.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,85 @@
...
+        assert(!(stepIndex == nil && intersectionIndex != nil), "It is forbidden to select `intersectionIndex` without specifying `stepIndex`.")
+        
+        return violations.filter { violation in
+            let validStep = (stepIndex ?? violation.stepIndex) == violation.stepIndex
+            let validIntersection = (intersectionIndex ?? violation.intersectionIndex) == violation.intersectionIndex
+            
+            return violation.legIndex == legIndex &&
```
1ec5, line 54: Since these violations are stored in a per-route array on the route response, it could become necessary for client code to zip the array of violations with the array of routes. Maybe we should have a higher-level function in RouteResponse that takes a route index too.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,85 @@
...
+                validStep &&
+                validIntersection
+        }
```
1ec5, line 57: This logic works, but it would be simpler if we break it down into multiple lazy filters: first lazily filter by leg, then if a step index is specified, lazily filter by step, then if an intersection index, filter by intersection.

Sources/MapboxDirections/RouteRoadClassesViolations.swift
```diff
@@ -0,0 +1,69 @@
...
+/**
+ Description of `RoadClasses` that were meant to be avoided during routing but still got used in a `Route`.
+ 
+ Such ignoring may be done by the engine if it cannot provide a route otherwise.
+ */
+public struct RouteRoadClassesViolations {
+    /**
+     Related `Route` object
+     */
+    public let route: Route
```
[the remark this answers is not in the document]
1ec5: RouteRoadClassesViolations is an odd name that begs to be nested under Route somehow. I think my suggestions here and in https://github.com/mapbox/mapbox-directions-swift/pull/627#discussion_r747277835 weren’t very consistent when taken together, so let me try again.

Currently we have a per-route array of violation containers, each of which contains a route reference and list of actual violations. Instead, the response should have a flat array of violations, each of which stores a route index. Then a method on RouteResponse can filter by leg, step, and intersection index at the same time. We can’t avoid maintaining a structure parallel to the Route model, but that parallel structure shouldn’t refer to each Route.

1ec5 (commented):
Client code will have to be especially careful working with these violations because they’re fundamentally index-based. If the navigation SDK’s understanding of the current route or route leg changes, for example due to refreshing, then parallel data stored about the relations may become unreliable.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -26,6 +30,52 @@ public struct RouteResponse {
...
+        
+        guard let routes = routes else {
+            roadClassViolations = nil
+            return
+        }
+        
+        var result = [RouteRoadClassesViolations]()
+        let avoidedClasses = routeOptions.roadClassesToAvoid
+        
+        for (routeIndex, route) in routes.enumerated() {
```
S2Ler, line 57: If `avoidedClasses` is empty, can we avoid running through all intersections?

Sources/MapboxDirections/RouteRoadClassesViolations.swift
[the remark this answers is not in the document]
Udumft: I hope I got your idea right this time :) please check

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -33,6 +35,13 @@
...
 		2B54080A245B23BE006C820B /* incorrectRouteRefreshResponse.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B540808245B23BE006C820B /* incorrectRouteRefreshResponse.json */; };
 		2B54080B245B23BE006C820B /* incorrectRouteRefreshResponse.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B540808245B23BE006C820B /* incorrectRouteRefreshResponse.json */; };
+		2B5F0DFC273ACE3B00CC2C1A /* tollAndFerryDirectionsRoute.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B5F0DFB273ACE3B00CC2C1A /* tollAndFerryDirectionsRoute.json */; };
+		2B5F0DFD273ACE3B00CC2C1A /* tollAndFerryDirectionsRoute.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B5F0DFB273ACE3B00CC2C1A /* tollAndFerryDirectionsRoute.json */; };
+		2B5F0DFE273ACE3B00CC2C1A /* tollAndFerryDirectionsRoute.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B5F0DFB273ACE3B00CC2C1A /* tollAndFerryDirectionsRoute.json */; };
+		2B5F0E00273BEB3600CC2C1A /* RoadClassExclusionViolation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B5F0DFF273BEB3600CC2C1A /* RoadClassExclusionViolation.swift */; };
```
chezzdev, line 41: Lots of seeming duplicates. Are they all correct?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Added the `RouteOptions.maximumHeight` and `RouteOptions.maximumWidth` properties for ensuring that the resulting routes can accommodate a vehicle of a certain size. ([#623](https://github.com/mapbox/mapbox-directions-swift/pull/623))
 * The `DirectionsPriority` struct now conforms to the `Codable` protocol. ([#623](https://github.com/mapbox/mapbox-directions-swift/pull/623))
 * Fixed an issue where the `RouteOptions.alleyPriority`, `RouteOptions.walkwayPriority`, and `RouteOptions.speed` properties were excluded from the encoded representation of a `RouteOptions` object. ([#623](https://github.com/mapbox/mapbox-directions-swift/pull/623))
+* Added the `RouteResponse.roadClassViolations` property, which indicates any requested `RouteOptions.roadClassesToAvoid` values that could not be satisfied when calculating the routes. ([#627](https://github.com/mapbox/mapbox-directions-swift/pull/627))
```
chezzdev, line 14: I guess we should mention convenience methods in the changelog too.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -26,6 +30,15 @@ public struct RouteResponse {
      This property does not persist after encoding and decoding.
      */
     public var created: Date = Date()
+    
+    /**
+     Managed of `RoadClasses` restrictions specified to `RouteOptions.roadClassesToAvoid` which were violated during route calculation.
+     
+     Routing engine may still utilize `RoadClasses` meant to be avoided in cases when routing is impossible otherwise.
+     
+     Resulting array is in the same order as `routes`, showing exact `RoadClasses` restrictions were ignored for each particular route at specific leg/step/intersection. `nil` and empty return arrays correspond to `nil` and empty `routes` array.
```
chezzdev, line 39: This description is a bit confusing to me. Can arrays be in the same order if they are of different sizes?
If we want to describe the order 
```suggestion
     Violations are ordered by routes from the `routes` array, then by a leg, step, and intersection, where `RoadClasses` restrictions were ignored.
     `nil` and empty arrays correspond to `nil` and empty `routes` array.
```

MapboxDirections.xcodeproj/project.pbxproj
1ec5: There’s a separate target for each Apple platform the project supports.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -204,3 +167,128 @@ extension RouteResponse: Codable {
     }
 
 }
+
+extension RouteResponse {
+    
+    mutating func parseIgnoredRoadClassesAvoidance() {
```
1ec5, line 173: Nit: rename to `findRoadClassExclusionViolations` or somesuch.

1ec5 (approved):

PR #628, opened 2021-11-11
Update Turf 2.1.0

bamx23:
Updates minimal version of Turf to 2.1.0 as it is required by the upcoming version of Navigation SDK.


ShanMa1991 (approved):

bamx23:
As Directions SDK doesn't require this bump we've decided to rather make this explicit dependency in Navigation SDK itself.

PR #631, opened 2021-12-02
Link Turf symbols to docset

### LINKED ISSUE #605: Published docset should automatically link references to Turf symbols
The jazzy-generated MapboxDirections docset has quite a few references to Turf symbols, particularly `LineString`, that go without further explanation. Developers unfamiliar with Turf may be able to intuit what the types represent, but not necessarily what they’re capable of in terms of methods. The documentation script should automatically detect references to Turf symbols [in postprocessing](https://github.com/mapbox/mapbox-directions-swift/blob/67f806a44510180e142a9c2a2577b98789e291fd/scripts/document.sh#L55-L58) and link them to the Turf docset being added in mapbox/turf-swift#162.

The mapbox-navigation-ios repository’s documentation script [postprocesses](https://github.com/mapbox/mapbox-navigation-ios/blob/ccaeb85f5aa4bf1bc93ea78793846c22d1416f49/scripts/document.sh#L70-L73) the jazzy-generated docset to insert these cross-references. It works, but it’s a bit fragile because the script has to [hard-code all the top-level symbols](https://github.com/mapbox/mapbox-navigation-ios/blob/ccaeb85f5aa4bf1bc93ea78793846c22d1416f49/scripts/document.sh#L30-L31) to link. Fortunately, Turf tends to have a pretty stable set of top-level symbols, since it’s built around standard GeoJSON concepts. A more ambitious approach would be to automatically generate this list of linkable symbols, perhaps based on [the docset’s search index](https://mapbox.github.io/turf-swift/2.0.0-rc.2/search.json).

/cc @mapbox/navigation-ios @mapbox/docs

chezzdev:
Resolves #605.

Interesting parts:
- install jazzy via bundler
- use search index of the published Turf docset to get public symbols
- link to symbols not only in signatures but also in `code` comments
- post-processing is extracted to a separate script with a more stable approach


scripts/document.sh
```diff
@@ -51,8 +43,19 @@ jazzy \
     --theme ${THEME} \
     --output ${OUTPUT} \
     --build-tool-arguments CODE_SIGN_IDENTITY=,CODE_SIGNING_REQUIRED=NO,CODE_SIGNING_ALLOWED=NO
+    
+# Link to turf documentation
+TURF_VERSION=$(grep -A 5 Turf Package.resolved | grep version | cut -d '"' -f 4)
```
bamx23, line 48: Not sure if it's easier to read and understand, but
```suggestion
TURF_VERSION=$(cat Package.resolved | python -c 'import sys, json; package = next(p for p in json.load(sys.stdin)["object"]["pins"] if p["package"] == "Turf"); print(package["state"]["version"])')
```

scripts/document.sh
```diff
@@ -51,8 +43,19 @@ jazzy \
...
+TURF_BASE_URL="https://mapbox.github.io/turf-swift/${TURF_VERSION}"
+TURF_INDEX=$(curl -s "${TURF_BASE_URL}/search.json")
+TURF_STRUCTS=$(echo ${TURF_INDEX} | python -m json.tool | grep -A 2 \"Structs/ | grep parent_name | uniq | cut -d '"' -f 4 | paste -sd "|" -)
```
bamx23, line 51: It looks a bit "too magic" and fragile... What do you think, could it be written in a Python one-liner?

Gemfile
```diff
@@ -0,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'jazzy', '~> 0.14'
```
bamx23, line 2: New line, please)

bamx23 (approved):

scripts/document.sh
1ec5: By the way, if we’re worried about the shell scripting approach being too fragile, we could use a tool like [`jq`](https://stedolan.github.io/jq/), which works with JSON in a shell-scripty manner. Python could work too, but it would strongly suggest rewriting the whole script in a more understandable language. 😅

chezzdev: Went with a python oneliner not to introduce new dependecies.

scripts/document.sh
chezzdev: Definitely not a one-liner, but I extracted post-processing logic into Python script. PTAL

Gemfile
chezzdev: Done, thanks.

scripts/postprocess-docs.py
```diff
@@ -0,0 +1,47 @@
...
+  sys.exit(2)
+for opt, arg in opts:
+  if opt in ("-b", "--baseurl"):
+     base_url = arg
+  elif opt in ("-d", "--docsroot"):
+     docs_root = arg
+     
+# receive, parse and remove index file
+index_file = "search.json"
+urllib.request.urlretrieve('{base_url}/search.json'.format(base_url=base_url), index_file)
```
bamx23, line 19: There should be a way to load data in memory and use `json.loads`

scripts/postprocess-docs.py
```diff
@@ -0,0 +1,47 @@
...
+    file_path = os.path.join(root, file)
+    
+    for line in fileinput.FileInput(file_path, inplace=True):
+      line = re.sub(r"<span class=\"kt\">({keys})</span>".format(keys='|'.join(symbols.keys())), lambda x: '<span class=\"kt\"><a href=\"{link}\">{symbol}</a></span>'.format(link=symbols.get(x.group(1)), symbol=x.group(1)), line.rstrip('\n'))
+      line = re.sub(r"<code>({keys})</code>".format(keys='|'.join(symbols.keys())), lambda x: '<code><a href=\"{link}\">{symbol}</a></code>'.format(link=symbols.get(x.group(1)), symbol=x.group(1)), line.rstrip('\n'))
+      
+      # Mapbox Directions specific
+      line = re.sub(r'MapboxDirections\s+(Docs|Reference)', lambda x: 'Mapbox Directions for Swift {section}'.format(section=x.group(1)), line.rstrip('\n'))
+      
+      print(line)
```
bamx23, line 47: Is it a debug print?

scripts/postprocess-docs.py
chezzdev: Right, that's a bit simpler. Updated.

scripts/postprocess-docs.py
chezzdev: No, this works in conjunction with `fileinput.FileInput(file_path, inplace=True):` for replacing lines in file inplace.

bamx23: NIT: I would prefer `sys.stdout.write(line)` then. (And/or add a comment)

scripts/postprocess-docs.py
```diff
@@ -0,0 +1,41 @@
...
+  sys.exit(2)
+for opt, arg in opts:
+  if opt in ("-b", "--baseurl"):
+     base_url = arg
+  elif opt in ("-d", "--docsroot"):
+     docs_root = arg
+     
+# receive, parse and remove index file
+with urllib.request.urlopen('{base_url}/search.json'.format(base_url=base_url)) as url:
+  index_data = json.loads(url.read().decode())
```
bamx23, line 19: Oh, maybe we can go this way?
```suggestion
  index_data = json.load(url)
```

bamx23 (approved):

chezzdev:
All the feedback was incorporated in mapbox/mapbox-navigation-ios#3655 too.

PR #634, opened 2021-12-09
Route refreshing

Udumft:
Added `RouteRefreshing` protocol to allow refreshing route with both `RefreshedRoute` and another `Route` instance.

This is required since NN's `RouterInterface` returns a complete `Route` instance instead of `RefreshedRoute` as was returned by `Directions` endpoint. Both cases should be handled similarly. Replacing `Route` objects causes a [visual glitch when](https://github.com/mapbox/mapbox-navigation-ios/issues/3621) route line is re-drawn.


Sources/MapboxDirections/RouteRefreshing.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+/**
+ A skeletal route containing infromation to refresh `Route` object annotations.
+ */
+public protocol RouteRefreshing {
```
chezzdev, line 6: If we want `Route` and `RefreshedRoute` to conform, could it be `RefreshableRoute`?

Sources/MapboxDirections/RouteRefreshing.swift
```diff
@@ -0,0 +1,37 @@
...
+    var refreshedLegs: [RouteLegRefreshing] { get }
```
chezzdev, line 7: Can we just call the properties as they are names right now so that conformance is automatic and there's no duplicate "refresh" info in the type and properties?

Sources/MapboxDirections/RouteRefreshing.swift
Udumft: Conformance to this protocol means that this object "may be used to refresh a route", but not "a route that can be refreshed". Naming is hard, but it's the best idea I had so far...

Sources/MapboxDirections/RouteRefreshing.swift
Udumft: The protocol requires `leg` to be a specific protocol conformance type, so it is problematic to allow both `Route` and `RefreshedRoute` to adopt it at the same time. Also, `RouteLeg.attributes` is internal, so we'll have to add a getter synonym there.

Sources/MapboxDirections/RouteRefreshResponse.swift
```diff
@@ -99,9 +99,9 @@ extension Route {
      
      - parameter refreshedRoute: The route containing leg attributes to merge into the receiver. If this route contains fewer legs than the receiver, this method skips legs from the beginning of the route to make up the difference, so that merging the attributes from a one-leg route affects only the last leg of the receiver.
      */
-    public func refreshLegAttributes(from refreshedRoute: RefreshedRoute) {
-        for (leg, refreshedLeg) in zip(legs.suffix(refreshedRoute.legs.count), refreshedRoute.legs) {
-            leg.attributes = refreshedLeg.attributes
+    public func refreshLegAttributes(from refreshedRoute: RouteRefreshing) {
+        for (leg, refreshedLeg) in zip(legs.suffix(refreshedRoute.refreshedLegs.count), refreshedRoute.refreshedLegs) {
```
bamx23, line 103: Should we handle the case where `refreshedRoute` has more legs than the current one?

Sources/MapboxDirections/RouteRefreshing.swift
bamx23: What about `RouteRefreshSource`? Agree, difficult naming

Sources/MapboxDirections/RouteRefreshing.swift
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ A skeletal route leg containing infromation to refresh `RouteLeg` object annotations.
```
bamx23, line 11: "annotations" or "attributes"? (here an in the comments above)

bamx23 (approved):

Sources/MapboxDirections/RouteRefreshResponse.swift
Udumft: This should never happen in real life. But in such case it would just skip excessive legs as it does when there are fewer legs

Sources/MapboxDirections/RouteRefreshing.swift
Udumft: These are synonyms.

bamx23: A bit confusing to me as a non-native speaker

Udumft: Corrected

Sources/MapboxDirections/RouteRefreshing.swift
Udumft: I like `RouteRefreshSource`. Updated the code

bamx23 (approved):

PR #644, opened 2022-01-24
Add support for mapbox designed route shields

jill-cardamon:
This PR adds support for mapbox designed route shields by creating the `MapboxShield` struct and adding it as a parameter for `VisualInstruction.Component.image`.

cc/ @abhishek1508

### DIFF
--- a/Sources/MapboxDirections/VisualInstructionComponent.swift
+++ b/Sources/MapboxDirections/VisualInstructionComponent.swift
@@ -45,9 +45,9 @@ public extension VisualInstruction {
          
          - parameter image: The component’s preferred image representation.
          - parameter alternativeText: The component’s alternative text representation. Use this representation if the image representation is unavailable or unusable, but consider formatting the text in a special way to distinguish it from an ordinary `.text` component.
-         - parameter mapboxShield: Optional for the component's mapbox shield representation.
+         - parameter mapboxShield: Optionally, a structured image representation for displaying a [highway shield](https://en.wikipedia.org/wiki/Highway_shield).
          */
-        case image(image: ImageRepresentation, alternativeText: TextRepresentation, mapboxShield: MapboxShield?)
+        case image(image: ImageRepresentation, alternativeText: TextRepresentation, mapboxShield: ShieldRepresentation? = nil)
 
         /**
          The component is an image of a zoomed junction, with a fallback text representation.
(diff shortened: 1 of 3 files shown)


Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -171,6 +172,64 @@ public extension VisualInstruction.Component {
...
+    
+    /**
+     String indicating the color of the text to be rendered on the route shield.
+     */
+    public let textColor: String
+    
+    /**
+     String indicating the display ref.
+     */
+    public let displayRef: String
```
1ec5, line 207: “Ref” is OpenStreetMap jargon that most developers would be unlikely to recognize. The Directions API uses this term for consistency with the `ref` property on the Step object, but this library uses the more intuitive term “reference code” with `code` as shorthand:

https://github.com/mapbox/mapbox-directions-swift/blob/0f20e9a3d5928333eb311dc3a6f6ae363a05a8a3/Sources/MapboxDirections/RouteStep.swift#L812-L819

Even so, what this property contains differs from the raw reference code in the Step object. Typically, a raw reference code would include an alphabetic network code, such as `US` for a U.S. Route, but many route networks omit that code from the shield, as in , which bears just the route number “101” instead of “US 101”. For this reason, the Directions API appends “display” to the property name, but that would have the effect of making it sound like a method in Swift. Instead, perhaps we could name this property `number`, as in route number, or `contents` if the possibility of an alphanumeric number would lead to confusion.

Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
```diff
@@ -23,32 +23,126 @@ class VisualInstructionComponentTests: XCTestCase {
...
+    func testShieldImageComponent() {
+        let componentJSON: [String : Any] = [
+            "text": "I 95",
             "type": "icon",
-            "imageBaseURL": "https://s3.amazonaws.com/mapbox/shields/v3/us-42",
+            "mapbox_shield": [
+                "base_url": "https://api.mapbox.com/styles/v1/",
+                "name": "us-interstate",
+                "text_color": "black",
+                "display_ref": "242"
```
1ec5, line 120: `text` is for I-95, but `display_ref` is for I-242 and `text_color` is for something other than an Interstate shield. Let’s make sure that this test is internally consistent, referring to the same highway shield throughout, to avoid confusion in the future.

Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
```diff
@@ -23,32 +23,126 @@ class VisualInstructionComponentTests: XCTestCase {
...
         XCTAssertNoThrow(component = try JSONDecoder().decode(VisualInstruction.Component.self, from: componentData))
         XCTAssertNotNil(component)
         if let component = component {
             switch component {
-            case .image(let image, let alternativeText):
-                XCTAssertEqual(image.imageBaseURL?.absoluteString, "https://s3.amazonaws.com/mapbox/shields/v3/us-42")
-                XCTAssertEqual(image.imageURL(scale: 1, format: .svg)?.absoluteString, "https://s3.amazonaws.com/mapbox/shields/v3/us-42@1x.svg")
-                XCTAssertEqual(image.imageURL(scale: 3, format: .svg)?.absoluteString, "https://s3.amazonaws.com/mapbox/shields/v3/us-42@3x.svg")
-                XCTAssertEqual(image.imageURL(scale: 3, format: .png)?.absoluteString, "https://s3.amazonaws.com/mapbox/shields/v3/us-42@3x.png")
-                XCTAssertEqual(alternativeText.text, "US 42")
```
1ec5, line 42: The Directions API will continue to return `imageBaseURL`, so we need to test that it gets decoded and round-tripped correctly. The `VisualInstruction.Component.imageURL(scale:format:)` case hasn’t been removed, so we should continue to cover the `scale` associated value in this manner.

Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
jill-cardamon: I was following a [test](https://github.com/mapbox/mapbox-java/blob/2141e8bdff2dabd01967500ce29fd60be226cf48/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/BannerComponentTest.java#L165-L180) from mapbox-java. Thanks for this catch - I will make sure it's consistent.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -45,9 +45,10 @@ public extension VisualInstruction {
          
          - parameter image: The component’s preferred image representation.
          - parameter alternativeText: The component’s alternative text representation. Use this representation if the image representation is unavailable or unusable, but consider formatting the text in a special way to distinguish it from an ordinary `.text` component.
+         - parameter mapboxShield: Optional for the component's mapbox shield representation.
          */
-        case image(image: ImageRepresentation, alternativeText: TextRepresentation)
-        
+        case image(image: ImageRepresentation, alternativeText: TextRepresentation, mapboxShield: MapboxShield?)
```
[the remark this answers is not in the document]
jill-cardamon: Ah I had this originally. Don't know why I deleted it. I'll make sure to set this default.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -171,6 +172,64 @@ public extension VisualInstruction.Component {
...
+
+    /**
+     String indicating the name of the route shield.
+     */
+    public let name: String
+    
+    /**
+     String indicating the color of the text to be rendered on the route shield.
+     */
+    public let textColor: String
```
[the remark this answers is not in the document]
jill-cardamon: Yes let's resolve it here. The current shield text color is limited to 7 colors (black, blue, green, orange, red, white, and yellow), and these are not standard CSS colors.

Sources/MapboxDirections/VisualInstructionComponent.swift
jill-cardamon: `text` would be most appropriate here since there is the possibility of alphanumeric route numbering (Australia and the UK, among others, use this system).

Sources/MapboxDirections/VisualInstructionComponent.swift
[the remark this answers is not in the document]
1ec5: Oh, my bad, I was confusing this property with the property in the shield spritesheet metadata file, which can contain CSS color values. The keywords you’re referring to are meant to be resolved based on the map or UI style. Since it can differ by context, a concrete UIColor/NSColor would be too inflexible here.

If the set of possible values is a closed set that we don’t foresee being extended in the future, we could make it into an enumeration. (The only standard sign color I don’t see represented here is purple, but I don’t think it ever occurs as a foreground text color, only a background text color.

Tests/MapboxDirectionsTests/VisualInstructionComponentTests.swift
1ec5: Never mind, the diff obscured the fact that you left the test in place but changed the route number throughout.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -238,7 +299,7 @@ extension VisualInstruction.Component: Codable {
                 imageBaseURL = URL(string: imageBaseURLString)
             }
             let imageRepresentation = ImageRepresentation(imageBaseURL: imageBaseURL)
-            self = .image(image: imageRepresentation, alternativeText: textRepresentation)
+            self = .image(image: imageRepresentation, alternativeText: textRepresentation, mapboxShield: mapboxShield)
```
1ec5, line 302: The changelog should mention that this enumeration case has been renamed from `VisualInstruction.Component.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -171,6 +172,64 @@ public extension VisualInstruction.Component {
     }
 }
 
+/**
+ A mapbox shield representation of a visual instruction component.
+ */
+public struct MapboxShield: Equatable, Codable {
```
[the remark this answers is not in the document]
1ec5: To nest this type under `VisualInstruction.Component`, put it inside an extension of `VisualInstruction.Component`. See the definition of `TextRepresentation` for an example.

Sources/MapboxDirections/VisualInstructionComponent.swift
1ec5: Rename `mapboxShield` to `shield` for consistency with the renamed type.

1ec5 (approved):
Good to merge once the struct is nested under `VisualInstruction.Component` and renamed to `ShieldRepresentation` and the default value for `shield` is added.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -45,9 +45,10 @@ public extension VisualInstruction {
          
          - parameter image: The component’s preferred image representation.
          - parameter alternativeText: The component’s alternative text representation. Use this representation if the image representation is unavailable or unusable, but consider formatting the text in a special way to distinguish it from an ordinary `.text` component.
+         - parameter shield: Optionally, a structured image representation for displaying a [highway shield](https://en.wikipedia.org/wiki/Highway_shield).
          */
-        case image(image: ImageRepresentation, alternativeText: TextRepresentation)
-        
+        case image(image: ImageRepresentation, alternativeText: TextRepresentation, shield: ShieldRepresentation? = nil)
```
1ec5, line 50: @bamx23 pointed out that, even with the default case, a switch statement will break because of this extra argument. We’ll need to move this into a separate case for when there is a shield representation.

1ec5: #646

PR #647, opened 2022-02-02
Fix backwards incompatible associated value in shield implementation

### LINKED ISSUE #646: Backwards-incompatible associated value added to VisualInstruction.Component case
#644 renamed `VisualInstruction.Component.image(image:alternativeText:)` to `VisualInstruction.Component.image(image:alternativeText:shield:)`. Even though the `shield` parameter has a default argument, `switch` statements that capture the arguments of `image` will be broken because the number of arguments has changed. The only way to maintain backwards compatibility is to revert `image` to how it was and introduce a new `shield` case that has the `shield` parameter.

/ref https://github.com/mapbox/mapbox-directions-swift/pull/644#discussion_r795025081
/cc @jill-cardamon @bamx23

jill-cardamon:
This PR fixes #646 by making `shield` an optional property of `VisualInstruction.Component.ImageRepresentation`.

cc/ @1ec5 @bamx23

### DIFF
--- a/Sources/MapboxDirections/VisualInstructionComponent.swift
+++ b/Sources/MapboxDirections/VisualInstructionComponent.swift
@@ -183,11 +183,11 @@ public extension VisualInstruction.Component {
         /**
          Initializes a mapbox shield with the given name, text color, and display ref.
          */
-        public init(baseURL: URL, name: String, textColor: String, code: String) {
+        public init(baseURL: URL, name: String, textColor: String, text: String) {
             self.baseURL = baseURL
             self.name = name
             self.textColor = textColor
-            self.code = code
+            self.text = text
         }
         
         /**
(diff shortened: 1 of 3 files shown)


Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -178,11 +183,11 @@ public extension VisualInstruction.Component {
         /**
          Initializes a mapbox shield with the given name, text color, and display ref.
          */
-        public init(baseURL: URL, name: String, textColor: String, text: String) {
+        public init(baseURL: URL, name: String, textColor: String, code: String) {
```
bamx23, line 186: Why do we need this change? As it's a backward-incompatible one.

jill-cardamon: It was to not interfere with `text` in the coding keys. I can actually change it back.

bamx23 (approved):

PR #655, opened 2022-02-18
Options & Credentials decoding

Udumft:
This PR adds convenience initialization methods to decode `RouteOptions` and `Credentials` from given route request URL. This feature is needed for NavSDK for interacting with NavNative and user code methods which report such entities in string format (in particular, needed for [RerouteController integration](https://github.com/mapbox/mapbox-navigation-ios/pull/3754)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Added `VisualInstruction.Component.ShieldRepresentation` struct and the `VisualInstruction.Component.ImageRepresentation.shield` property containing metadata for displaying a highway shield consistent with map styles used by the Mapbox Maps SDK. ([#644](https://github.com/mapbox/mapbox-directions-swift/pull/644), [#647](https://github.com/mapbox/mapbox-directions-swift/pull/647))
 * Added a `RouteLeg.viaWaypoints` property that lists the non-leg-separating waypoints (also known as “silent waypoints”) along a `RouteLeg`. Previously, you had to filter `DirectionsOptions.waypoints` to include only the items whose `Waypoints.separatesLegs` property was set to `true`, then zip them with `RouteResponse.routes`. This approach still works in some cases but is not guaranteed to be reliable for all Mapbox Directions API responses in the future. ([#656](https://github.com/mapbox/mapbox-directions-swift/pull/656))
+* Added `DirectionsOptions(url:)`, `RouteOptions.restore(from:)` and `DirectionsOptions.restore(from:)` methods for deserializing corresponding options object using appropriate request URL. ([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
 
 ## v2.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Added `VisualInstruction.Component.ShieldRepresentation` struct and the `VisualInstruction.Component.ImageRepresentation.shield` property containing metadata for displaying a highway shield consistent with map styles used by the Mapbox Maps SDK. ([#644](https://github.com/mapbox/mapbox-directions-swift/pull/644), [#647](https://github.com/mapbox/mapbox-directions-swift/pull/647))
 * Added a `RouteLeg.viaWaypoints` property that lists the non-leg-separating waypoints (also known as “silent waypoints”) along a `RouteLeg`. Previously, you had to filter `DirectionsOptions.waypoints` to include only the items whose `Waypoints.separatesLegs` property was set to `true`, then zip them with `RouteResponse.routes`. This approach still works in some cases but is not guaranteed to be reliable for all Mapbox Directions API responses in the future. ([#656](https://github.com/mapbox/mapbox-directions-swift/pull/656))
-* Added `DirectionsOptions(url:)`, `RouteOptions.restore(from:)` and `DirectionsOptions.restore(from:)` methods for deserializing corresponding options object using appropriate request URL. ([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
+* Added `DirectionsOptions(url:)`, `RouteOptions(url:)` and extended existing `DirectionsOptions(waypoints:profileIdentifier:queryItems:)`, `RouteOptions(waypoints:profileIdentifier:queryItems:)`, `MatchOptions(waypoints:profileIdentifier:queryItems:)` and related convenience init methods for deserializing corresponding options object using appropriate request URL or it's query items. ([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
 
 ## v2.2.0
 
--- a/Sources/MapboxDirections/Credentials.swift
+++ b/Sources/MapboxDirections/Credentials.swift
@@ -63,6 +63,17 @@ public struct Credentials: Equatable {
             self.host = URL(string: "https://api.mapbox.com")!
         }
     }
+    
+    public init(requestURL url: URL) {
+        var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        let accessToken = components?
+            .queryItems?
+            .first { $0.name == "access_token" }?
+            .value
+        components?.path = "/"
+        components?.queryItems = nil
+        self.init(accessToken: accessToken, host: components?.url)
+    }
 }
 
 @available(*, deprecated, renamed: "Credentials")
--- a/Sources/MapboxDirections/Credentials.swift
+++ b/Sources/MapboxDirections/Credentials.swift
@@ -64,6 +64,12 @@ public struct Credentials: Equatable {
         }
     }
     
+    /**
+     :nodoc:
+     Attempts to get `host` and `accessToken` from provided URL to create `Credentials` instance.
+     
+     If it is impossible to extract parameter(s) - default values will be used.
+     */
     public init(requestURL url: URL) {
         var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
         let accessToken = components?
--- a/Sources/MapboxDirections/DirectionsOptions.swift
+++ b/Sources/MapboxDirections/DirectionsOptions.swift
@@ -245,7 +245,7 @@ open class DirectionsOptions: Codable {
         }
             
         guard waypoints.count >= 2 else {
-                    return nil
+            return nil
         }
         
         let profileIdentifier = ProfileIdentifier(rawValue: url.pathComponents.dropLast().suffix(2).joined(separator: "/"))
--- a/Sources/MapboxDirections/DirectionsOptions.swift
+++ b/Sources/MapboxDirections/DirectionsOptions.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Polyline
+import Turf
 
 /**
  Maximum length of an HTTP request URL for the purposes of switching from GET to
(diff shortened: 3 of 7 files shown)


1ec5:
Does the URL-to-RouteOptions portion of this PR supersede #580?

/cc @jill-cardamon

Sources/MapboxDirections/Credentials.swift
```diff
@@ -63,6 +63,23 @@ public struct Credentials: Equatable {
             self.host = URL(string: "https://api.mapbox.com")!
         }
     }
+    
+    public init?(requestURL url: URL) {
+        guard let scheme = url.scheme,
+              let host = url.host,
+              let hostURL = URL(string: "\(scheme)://\(host)") else {
```
1ec5, line 70: Use URLComponents to construct a URL securely and reliably.

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -369,6 +370,94 @@ open class DirectionsOptions: Codable {
         return queryItems
     }
     
+    /**
+     :nodoc:
+     Method to deserialize options parameters from given `url`.
+     */
+    open func restore(from url: URL) {
```
1ec5, line 377: This is generally useful functionality that application would find useful when implementing Handoff, so we should make it a formal part of the public API.

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -369,6 +395,96 @@ open class DirectionsOptions: Codable {
...
+        
+        let getElement: ((_ array: [String]?, _ index: Int) -> String?) = { array, index in
+            if array?.count ?? -1 > index {
+                return array?[index]
+            }
+            return nil
+        }
+        
+        waypoints.enumerated().forEach {
+            if let approach = getElement(waypointsData[0], $0.offset) {
```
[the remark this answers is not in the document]
Udumft: using `zip`s was my previous approach. It results in multiple cycles iterating over `waypoints` for each parameter. This one at least is easier to understand and see that it's all the properties that are applied to waypoints parameters. Probably there could be a method to decode a waypoints array from an URL for clarity, but I am not sure it's worth it.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## v2.3.0
 
 * Added `VisualInstruction.Component.ShieldRepresentation` struct for displaying a highway shield. Added `VisualInstruction.Component.ImageRepresentation.shield` property. ([#644](https://github.com/mapbox/mapbox-directions-swift/pull/644), [#647](https://github.com/mapbox/mapbox-directions-swift/pull/647))
+* Added `DirectionsOptions.init?(url:)`, `RouteOptions.restore(from:)` and `DirectionsOptions.restore(from:)` methods to allow deserializing corresponding options object using appropriate request URL.([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
```
1ec5, line 6: I think this is the form that jazzy requires for autolinking:

```suggestion
* Added `DirectionsOptions(url:)`, `RouteOptions.restore(from:)` and `DirectionsOptions.restore(from:)` methods for deserializing corresponding options object using appropriate request URL. ([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
```

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -129,6 +129,44 @@ open class DirectionsOptions: Codable {
...
     }
     
+    /**
+     Creates new options object by deserializing given `url`
+     
+     Initialization fails if it is unable to extract `waypoints` list and `profileIdentifier`. If other properties are failed to decode - it will just skip them.
+     
+     - parameter url: An URL, used to make a route request.
+     */
+    public convenience init?(url: URL) {
```
1ec5, line 139: What’s the difference between `init(url:)` and `restore(from:)`? Why would the developer call `restore(from:)` on an existing RouteOptions object instead of initializing a new RouteOptions object? Sorry that https://github.com/mapbox/mapbox-directions-swift/pull/655#discussion_r810275117 didn’t make this point clearly.

Udumft: Well, my original thought was that this method would unlock use-cases where user would want "the same exact type of route, but between different points". But now I see that it is simply achieved just by having an `init`. Probably should make `restore(from:)` internal to avoid cluttering the API.

Sources/MapboxDirections/DirectionsOptions.swift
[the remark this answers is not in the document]
bamx23: The main issue here is that using hardcoded indices is a dangerous source of bugs. Do you have any ideas on how to avoid this? I can imagine use of some enum here, not sure

Sources/MapboxDirections/DirectionsOptions.swift
1ec5: It looks like the distinction is actually between URL path components and URL query items. How about folding everything into the initializer, then factoring out an initializer that takes the waypoints, profile identifier, and an array of query items?

Sources/MapboxDirections/DirectionsOptions.swift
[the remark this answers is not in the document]
Udumft: I don't think something like enum would make sense here because `waypointsData` is still filled manually here. Ideally I would factor `waypointsData` into special wrapper which does this data manipulations, but I think it is an overkill for the sake of this method.

Sources/MapboxDirections/DirectionsOptions.swift
Udumft: Updated. Init now looks a bit more than I would love to, and I am not 100% sure about that mix of arguments on the `DirectionsOptions(waypoints:profileIdentifier:queryItems:)` as it has some SDK types together with url request parts in the same row which might look weird.  Other than that, API looks more natural to me.

Sources/MapboxDirections/Credentials.swift
```diff
@@ -63,6 +63,17 @@ public struct Credentials: Equatable {
...
+    
+    public init(requestURL url: URL) {
+        var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        let accessToken = components?
+            .queryItems?
+            .first { $0.name == "access_token" }?
+            .value
+        components?.path = "/"
+        components?.queryItems = nil
+        self.init(accessToken: accessToken, host: components?.url)
```
bamx23, line 75: Should we return `nil` in case there was an issue while parsing the URL? As at the moment we might implicitly return an object with default values

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -122,10 +123,133 @@ open class DirectionsOptions: Codable {
...
+                  let latitude = LocationDegrees(latitudeString),
+                  let longitude = LocationDegrees(longitudeString) else {
+                      return nil
+            }
+            return Waypoint(coordinate: .init(latitude: latitude,
+                                              longitude: longitude))
+        }
+            
+        guard waypoints.count >= 2 else {
+                    return nil
```
bamx23, line 245: NIT: Too many white spaces

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -122,10 +123,133 @@ open class DirectionsOptions: Codable {
...
-    required public init(waypoints: [Waypoint], profileIdentifier: ProfileIdentifier? = nil) {
+    required public init(waypoints: [Waypoint], profileIdentifier: ProfileIdentifier? = nil, queryItems: [URLQueryItem]? = nil) {
         self.waypoints = waypoints
         self.profileIdentifier = profileIdentifier ?? .automobile
+        
+        guard let queryItems = queryItems else {
+            return
+        }
+        
+        let mappedQueryItems = Dictionary<String, String>(uniqueKeysWithValues: queryItems.compactMap {
```
bamx23, line 136: I would suggest `Dictionary(_:uniquingKeysWith:)` as this one might led to a runtime error if there are duplicates in the query items (which is possible in general). `uniquingKeysWith` might be like `(_, latestValue) in latestValue`

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -122,10 +123,133 @@ open class DirectionsOptions: Codable {
...
+            }
+            return nil
+        }
+        
+        waypoints.enumerated().forEach {
+            if let approach = getElement(waypointsData[0], $0.offset) {
+                $0.element.allowsArrivingOnOppositeSide = approach == "unrestricted" ? true : false
+            }
+            
+            if let descriptions = waypointsData[1]?[$0.offset].components(separatedBy: ",") {
```
bamx23, line 197: Have you thought about using `getElement` here and below?

Sources/MapboxDirections/MapMatching/MatchOptions.swift
```diff
@@ -32,16 +33,39 @@ open class MatchOptions: DirectionsOptions {
...
-    public required init(waypoints: [Waypoint], profileIdentifier: ProfileIdentifier? = nil) {
-        super.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
+    public required init(waypoints: [Waypoint], profileIdentifier: ProfileIdentifier? = nil, queryItems: [URLQueryItem]? = nil) {
+        super.init(waypoints: waypoints, profileIdentifier: profileIdentifier, queryItems: queryItems)
+        
+        guard let queryItems = queryItems else {
+            return
+        }
+        
+        let mappedQueryItems = Dictionary<String, String>(uniqueKeysWithValues: queryItems.compactMap {
```
bamx23, line 52: The same idea about `Dictionary(_:uniquingKeysWith:)`

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -17,11 +17,73 @@ open class RouteOptions: DirectionsOptions {
...
         let profilesDisallowingUTurns: [ProfileIdentifier] = [.automobile, .automobileAvoidingTraffic]
         allowsUTurnAtWaypoint = !profilesDisallowingUTurns.contains(profileIdentifier ?? .automobile)
-        super.init(waypoints: waypoints, profileIdentifier: profileIdentifier)
+        super.init(waypoints: waypoints, profileIdentifier: profileIdentifier, queryItems: queryItems)
+        
+        guard let queryItems = queryItems else {
+            return
+        }
+        
+        let mappedQueryItems = Dictionary<String, String>(uniqueKeysWithValues: queryItems.compactMap {
```
bamx23, line 31: And once again about `Dictionary(_:uniquingKeysWith:)`

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -122,10 +123,133 @@ open class DirectionsOptions: Codable {
...
+                waypoints.enumerated().forEach {
+                    $0.element.separatesLegs = indicies.contains($0.offset)
+                }
+            }
+        }
+        
+        let waypointsData = [mappedQueryItems["approaches"]?.components(separatedBy: ";"),
+                             mappedQueryItems["bearings"]?.components(separatedBy: ";"),
+                             mappedQueryItems["radiuses"]?.components(separatedBy: ";"),
+                             mappedQueryItems["waypoint_names"]?.components(separatedBy: ";"),
```
1ec5, line 181: At a glance, I don’t see any glaring logic errors in this conversion. However, I’m a bit concerned about the maintainability of this code as the Directions API evolves. Let’s at least increase the code coverage level on these additions by expanding the unit tests.

As possible tail work, it would be nice to find a way to fold it into the existing Codable implementation, which expects a JSON representation of the same data that’s pretty consistent with the URL query parameters following #593. Maybe we could improve maintainability by first converting the `URLQueryItem`s into the JSON representation very literally, then decoding that JSON data. If there are any roundtripping issues that way, they would also affect codability.

Sources/MapboxDirections/DirectionsOptions.swift
Udumft: Yes, you are right. I forgot to spread this usage across all items. My bad.

Sources/MapboxDirections/Credentials.swift
Udumft: The use-case this method is designed for is fine to subtly fallback to a default. I didn't think this method would be usable or visible to wide auditory, but indeed this case should be at least documented.

Sources/MapboxDirections/DirectionsOptions.swift
Udumft: Covered all of these by unit test check

Sources/MapboxDirections/DirectionsOptions.swift
Udumft: Updated. Thanks for the suggestion

bamx23 (approved):

PR #664, opened 2022-03-14
Support multiple road classes in `RouteOptions.roadClassesToAvoid` and `RouteOptions.roadClassesToAllow` properties

azarovalex:
[Directions server-side API](https://docs.mapbox.com/api/navigation/directions/) now allows passing multiple road classes in these fields, so the corresponding SDK code is updated.

Looks like `RoadClasses.tunnel` and `RoadClasses.restricted` are not supported by Directions API in `RouteOptions.roadClassesToAvoid` anymore, so I updated documentation for these properties.

Error when trying to avoid tunnels:
image

@1ec5 Looks like `RoadClasses.tunnel` and `RoadClasses.restricted` were only used in `RouteOptions.roadClassesToAvoid `, now that Directions API doesn't support them, should we mark these properties as deprecated?

### DIFF
--- a/Sources/MapboxDirections/RoadClasses.swift
+++ b/Sources/MapboxDirections/RoadClasses.swift
@@ -185,3 +185,14 @@ extension RoadClasses: Codable {
         }
     }
 }
+
+extension RoadClasses {
+    static var validAvoidClasses: RoadClasses {
+        // .restricted and .tunnel road types are not supported by Directions API anymore
+        return [.toll, .motorway, .ferry, .unpaved, .cashTollOnly]
+    }
+
+    static var validAllowClasses: RoadClasses {
+        return [.highOccupancyVehicle2, .highOccupancyVehicle3, .highOccupancyToll]
+    }
+}
--- a/Sources/MapboxDirections/RoadClasses.swift
+++ b/Sources/MapboxDirections/RoadClasses.swift
@@ -21,6 +21,8 @@ public struct RoadClasses: OptionSet, CustomStringConvertible {
      The road segment has access restrictions.
      
      A road segment may have this class if there are [general access restrictions](https://wiki.openstreetmap.org/wiki/Key:access) or a [high-occupancy vehicle](https://wiki.openstreetmap.org/wiki/Key:hov) restriction.
+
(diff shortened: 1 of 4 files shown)


Sources/MapboxDirections/RoadClasses.swift
```diff
@@ -185,3 +185,13 @@ extension RoadClasses: Codable {
         }
     }
 }
+
+extension RoadClasses {
+    static var validAvoidClasses: RoadClasses {
+        return [.toll, .motorway, .ferry, .unpaved, .cashTollOnly]
```
1ec5, line 191: I’m wary of hard-coding these allowlists. We’ve done that in the past and it has gotten outdated. Better to document our current expectations and allow the request to return an error.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -281,20 +281,21 @@ open class RouteOptions: DirectionsOptions {
...
-        
-        if !roadClassesToAvoid.isEmpty && roadClassesToAvoid.isDisjoint(with: [.highOccupancyVehicle2, .highOccupancyVehicle3, .highOccupancyToll]) {
-            let allRoadClasses = roadClassesToAvoid.description.components(separatedBy: ",").filter { !$0.isEmpty }
-            precondition(allRoadClasses.count < 2, "You can only avoid one road class at a time.")
-            if let firstRoadClass = allRoadClasses.first {
-                params.append(URLQueryItem(name: CodingKeys.roadClassesToAvoid.stringValue, value: firstRoadClass))
+
+        if !roadClassesToAvoid.isEmpty {
+            if !roadClassesToAvoid.isSubset(of: .validAvoidClasses) {
+                assertionFailure("Provided road classes cannot be avoided")
```
1ec5, line 287: In general, we avoid client-side validation, except for options that don’t make logical sense.

Sources/MapboxDirections/RoadClasses.swift
azarovalex: Ok, I removed the hardcode and filtering of allowed road classes, now the server will perform the check

Sources/MapboxDirections/RouteOptions.swift
azarovalex: Agree, removed the assertion.

bamx23 (approved):

PR #669, opened 2022-03-25
Foreign JSON members roundtrip

### LINKED ISSUE #637: Codable types should roundtrip unrecognized JSON properties
Codable types defined by this package, such as RouteStep, should roundtrip from and to JSON in its entirety, including any unrecognized properties. This would ensure forward compatibility in the event that the Directions API includes content in a response that had not been publicly documented at the time of this package’s release.

When decoding from JSON, any unrecognized property in `KeyedDecodingContainer.allKeys` but not `CodingKeys` should be stored in a `userInfo` property of type `JSONObject`, which would get encoded back to JSON as the original properties. If the `userInfo` dictionary has a key that’s in `CodingKeys`, the Encodable implementation could ignore it in favor of the built-in support.

This would effectively be a lighterweight alternative to #60.

/ref mapbox/turf-swift#174
/cc @mapbox/navigation-ios @SiarheiFedartsou

Udumft:
Resolves #637 
This PR utilizes `Turf.ForeignMemberContainer` and it's analog for reference types: `ForeignMemberClassContainer`. 
Revisiting types, related to processing various responses from the Directions API, found and refactored some outdated coding logic.


Sources/MapboxDirections/Extensions/ForeignMemberContainer.swift
```diff
@@ -0,0 +1,99 @@
...
+        self.stringValue = String(intValue)
+        self.intValue = intValue
+    }
+}
+
+extension ForeignMemberContainer {
+    /**
+     Decodes any foreign members using the given decoder.
+     */
+    mutating func decodeForeignMembers<WellKnownCodingKeys>(notKeyedBy _: WellKnownCodingKeys.Type, with decoder: Decoder) throws where WellKnownCodingKeys: CodingKey {
```
Udumft, line 26: Turf didn't expose these methods. It doesn't break existing conformances, but if we want to use it in Directions, we need these methods.
Should `ForeignMemberClassContainer` follow the same rule? I feel that it's `decoding` method should remain public to allow users successfully subclass our `open` types. Otherwise, the mechanism will not handle new subclass properties and put them into `foreignMembers` as unrecognized.

Sources/MapboxDirections/Extensions/ForeignMemberContainer.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+}
+
+/**
+ A GeoJSON *class* that can contain [foreign members](https://datatracker.ietf.org/doc/html/rfc7946#section-6.1) in arbitrary keys.
+ 
+ When subclassing `ForeignMemberClassContainer` type, you should call `decodeForeignMembers(notKeyedBy:with:)` during your `Decodable.init(from:)` initializer if your subclass has added any new properties.
+ */
+public protocol ForeignMemberClassContainer: AnyObject {
```
Udumft, line 54: I'm open to naming suggestions here!

Sources/MapboxDirections/Intersection.swift
```diff
@@ -4,7 +4,10 @@ import Turf
 /**
  A single cross street along a step.
  */
-public struct Intersection {
+public struct Intersection: ForeignMemberContainer {
+    public var foreignMembers: JSONObject = [:]
+    public var lanesForeignMembers: [JSONObject] = []
```
Udumft, line 9: This falls our of usual pattern, but since `Intersection` takes care of Coding lanes data - there is no better place to store Lane's foreign members.

Sources/MapboxDirections/RouteLeg.swift
```diff
@@ -7,8 +7,11 @@ import Turf
 
  You do not create instances of this class directly. Instead, you receive route leg objects as part of route objects when you request directions using the `Directions.calculate(_:completionHandler:)` method.
  */
-open class RouteLeg: Codable {
-    public enum CodingKeys: String, CodingKey {
+open class RouteLeg: Codable, ForeignMemberClassContainer {
+    public var foreignMembers: JSONObject = [:]
+    public var attributesForeignMembers: JSONObject = [:]
```
Udumft, line 12: Same as for `Intersection.Lane`

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -128,13 +130,15 @@ extension RouteResponse: Codable {
                 let waypoint = Waypoint(coordinate: decodedWaypoint.coordinate,
                                         coordinateAccuracy: waypointInOptions.coordinateAccuracy,
                                         name: waypointInOptions.name?.nonEmptyString ?? decodedWaypoint.name)
-
+                waypoint.snappedDistance = decodedWaypoint.snappedDistance
```
Udumft, line 133: For some reason not all of the Waypoint properties were copied here. Was that intended?

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -43,7 +46,6 @@ public struct RouteResponse {
 
 extension RouteResponse: Codable {
     enum CodingKeys: String, CodingKey {
-        case code
```
Udumft, line 46: I found this Key is unused.

Tests/MapboxDirectionsTests/RouteStepTests.swift
```diff
@@ -225,12 +225,8 @@ class RouteStepTests: XCTestCase {
                 var encodedStepJSON: Any?
                 XCTAssertNoThrow(encodedStepJSON = try JSONSerialization.jsonObject(with: encodedStepData, options: []))
                 XCTAssertNotNil(encodedStepJSON)
-
-                // https://github.com/mapbox/mapbox-directions-swift/issues/125
-                var referenceStepJSON = stepJSON
-                referenceStepJSON.removeValue(forKey: "weight")
```
Udumft, line 231: This is now covered by the new feature!

Tests/MapboxDirectionsTests/RouteTests.swift
```diff
@@ -10,7 +10,7 @@ class RouteTests: XCTestCase {
             "legs": [
                 [
                     "summary": "West 6th Avenue Freeway, South University Boulevard",
-                    "weight": 1346.3,
+                    "weight": 2346.3,
```
Udumft, line 13: Updated just for variety with another `weight` item in this JSON

Sources/MapboxDirections/Extensions/ForeignMemberContainer.swift
1ec5: I think that’s reasonable. Turf’s structs can’t be subclassed, but MapboxDirections’ classes are often subclassed, making it more important to provide conveniences like this method.

Sources/MapboxDirections/RestStop.swift
```diff
@@ -22,4 +24,22 @@ public struct RestStop: Codable, Equatable {
...
+    
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(type, forKey: .type)
+        
+        try encodeForeignMembers(notKeyedBy: CodingKeys.self, to: encoder)
+    }
+    
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        return lhs.type == rhs.type
```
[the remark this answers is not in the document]
1ec5: In Turf, any struct that conforms to ForeignMemberContainer considers two instances with different foreign members to be unequal.

Sources/MapboxDirections/RouteResponse.swift
1ec5: Probably an oversight.

Sources/MapboxDirections/RouteResponse.swift
1ec5: This key is [part of the Directions API response](https://docs.mapbox.com/api/navigation/directions/#response-retrieve-directions), but it looks like Directions is obtaining the `code` through ResponseDisposition:

https://github.com/mapbox/mapbox-directions-swift/blob/f46b02eeb196c4b23e96bc3f1a303266010e96ef/Sources/MapboxDirections/Directions.swift#L200-L215

Sources/MapboxDirections/Extensions/ForeignMemberContainer.swift
1ec5: How about `ForeignMemberContainerClass` or `ForeignMemberContainingClass`?

Sources/MapboxDirections/Extensions/ForeignMemberContainer.swift
Udumft: >Is ForeignMemberClassContainer only separate from ForeignMemberContaining because of the need to make these methods open rather than public?

No, the difference is caused by value and reference types difference. `ForeignMemberContainer` uses `mutating` method which is fine to use in struct's `init`, but classes do not allow it since `self` is immutable during initialization methods (such as `init(from decoder)`. 
Also, encoding is different too. Method should be called only in the base class and since structs does not care about subclassing it can use provided `CodingKeys`, while class implementation has to rely on `foreignMembers` to be correctly populated beforehand.
Given that I've decided to put it into another protocol.

Sources/MapboxDirections/RestStop.swift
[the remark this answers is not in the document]
Udumft: But Turf basically provides a set of data containers. I am afraid that accounting for foreignMembers may introduce subtle bugs in Directions or Navigation logic when 'suddenly' objects will become unequal because of members which are unimportant for the SDK.

Tests/MapboxDirectionsTests/V5Tests.swift
```diff
@@ -63,7 +63,7 @@ class V5Tests: XCTestCase {
         }
         XCTAssertNotNil(task)
         
-        waitForExpectations(timeout: 10) { (error) in
+        waitForExpectations(timeout: 15) { (error) in
```
Udumft, line 66: Decoding now takes a lot more time... This especially affects `geoJSON` formatted responses (see `testGeoJSON` time increased from [~2.7s](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/1478/workflows/2ce59a0c-b58d-462f-a3ca-1e4e0f030a44/jobs/5838) to [~12s](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift/1480/workflows/1835e71e-dee8-4796-b998-031fdbf2b0ba/jobs/5850)). I have no ideas on how to improve performance on this without re-designing the entire JSON roundtripping mechanics, starting from Turf. On my local machine, this test time is increased from `1s` to `5s`.

Sources/MapboxDirections/RestStop.swift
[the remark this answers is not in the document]
1ec5: Agreed, this is more likely for classes than structs anyhow.

Tests/MapboxDirectionsTests/V5Tests.swift
1ec5: Since it’s extremely unlikely for a GeoJSON-formatted geometry in a Directions API response to contain any foreign members or for these foreign members to have any significance, perhaps Turf should look for an optional `userInfo` key indicating that foreign members should be ignored when encoding or decoding. Turf could recursively pass down this userInfo key. This would be a generally useful feature.

In the meantime, I’m not terribly concerned about a performance hit limited to GeoJSON-formatted geometries, since the navigation SDK’s NavigationRouteOptions defaults to Polyline encoding, partly for performance reasons.

1ec5: > Since it’s extremely unlikely for a GeoJSON-formatted geometry in a Directions API response to contain any foreign members or for these foreign members to have any significance, perhaps Turf should look for an optional `userInfo` key indicating that foreign members should be ignored when encoding or decoding. Turf could recursively pass down this userInfo key. This would be a generally useful feature.

Tracking in mapbox/turf-swift#186.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fixed a crash that occurred when `RouteOptions.roadClassesToAvoid` or `RouteOptions.roadClassesToAllow` properties contained multiple road classes.
 * `RoadClasses.tunnel` and `RoadClasses.restricted` are no longer supported in `RouteOptions.roadClassesToAvoid` or `RouteOptions.roadClassesToAllow` properties
 * Added `DirectionsOptions(url:)`, `RouteOptions(url:)` and extended existing `DirectionsOptions(waypoints:profileIdentifier:queryItems:)`, `RouteOptions(waypoints:profileIdentifier:queryItems:)`, `MatchOptions(waypoints:profileIdentifier:queryItems:)` and related convenience init methods for deserializing corresponding options object using appropriate request URL or it's query items. ([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
+* `RouteResponse`, `RouteRefreshResponse`, `MatchResponse` and all underlying types now correctly handle unrecogized (foreign) JSON values and preserve them on coding/decoding using `ForeignMemberContainer` and `ForeignMemberClassContainer` implementations. ([#669](https://github.com/mapbox/mapbox-directions-swift/pull/669))
```
1ec5, line 8: Typo:

```suggestion
* `RouteResponse`, `RouteRefreshResponse`, `MatchResponse` and all underlying types now correctly handle unrecognized (foreign) JSON values and preserve them on coding/decoding using `ForeignMemberContainer` and `ForeignMemberClassContainer` implementations. ([#669](https://github.com/mapbox/mapbox-directions-swift/pull/669))
```

Should we say something about how this could be useful for using properties in the response that are still in beta?

Sources/MapboxDirections/RouteResponse.swift
1ec5: The changelog should mention that we’ve fixed the issue where `Waypoint.snappedDistance` was getting lost when decoding a RouteResponse.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -319,17 +321,7 @@ struct Road {
     }
     
     init(name: String, ref: String?, exits: String?, destination: String?, rotaryName: String?) {
-        if !name.isEmpty, let ref = ref {
-            // Mapbox Directions API v5 encodes the ref separately from the name but redundantly includes the ref in the name for backwards compatibility. Remove the ref from the name.
```
1ec5, line 323: Directions API v5 profiles powered by Valhalla no longer include the ref in the name. However, the mapbox/cycling profile, which is powered by OSRM, [still includes the ref](https://api.mapbox.com/directions/v5/mapbox/cycling/-84.49966248187295,39.10237033604159;-84.49950427649345,39.101791247428054.json?steps=true&overview=full&geometries=geojson&roundabout_exits=true&access_token=).

/ref Project-OSRM/osrm-backend#2704

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -526,21 +585,23 @@ open class RouteStep: Codable {
...
-        var maneuver = container.nestedContainer(keyedBy: ManeuverCodingKeys.self, forKey: .maneuver)
-        try maneuver.encode(instructions, forKey: .instruction)
-        try maneuver.encode(maneuverType, forKey: .type)
-        try maneuver.encodeIfPresent(exitIndex, forKey: .exitIndex)
-
-        try maneuver.encodeIfPresent(maneuverDirection, forKey: .direction)
-        try maneuver.encode(LocationCoordinate2DCodable(maneuverLocation), forKey: .location)
-        try maneuver.encodeIfPresent(initialHeading, forKey: .initialHeading)
-        try maneuver.encodeIfPresent(finalHeading, forKey: .finalHeading)
+        var maneuver = Maneuver(instructions: instructions,
```
1ec5, line 588: Good idea, long overdue.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -582,27 +675,24 @@ open class RouteStep: Codable {
...
         if let polyLineString = try container.decodeIfPresent(PolyLineString.self, forKey: .shape) {
             shape = try LineString(polyLineString: polyLineString)
         } else {
             shape = nil
         }
         
-        if let instruction = try? maneuver.decode(String.self, forKey: .instruction) {
-            instructions = instruction
-        } else {
-            instructions = "\(maneuverType) \(maneuverDirection?.rawValue ?? "")"
```
1ec5, line 604: FYI, this was a fallback for custom OSRM endpoints that had not integrated OSRM Text Instructions. We primarily used it for debugging purposes, since it was easier to fashion a test fixture using OSRM’s demo page than using the Directions API at the time. These days, any OSRM endpoint would be expected to integrate OSRMTI.

1ec5 (approved):
This looks good to me, other than probably needing to bring back the ref-in-name code for now. The performance issue is significant but scoped such that we can treat it as tail work if necessary.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -377,7 +380,7 @@ extension Road: Codable {
             }
             try container.encodeIfPresent(name, forKey: .name)
         } else {
-            try container.encodeIfPresent(ref, forKey: .name)
+            try container.encode(ref ?? "", forKey: .name)
```
Udumft, line 383: Had to change that to maintain encoding and decoding to preserve JSON structure as it is reported by Directions.

Udumft:
I've tried disabling foreign members coding in Turf to measure performance hit (using [this mechanism](https://github.com/mapbox/turf-swift/pull/187)) but it still slowed down the coding significantly. Probably need to find another approach...

1ec5:
Can we narrow the performance penalty down to a particular property or object in the response? For example, what if the response lacks annotations, which already cause the leg to be doubly decoded even before this change?

Udumft:
I found that `Road` object does not need a foreign member container since it basically codes a subset inside a route step. But instead, it did double coding, treating all other members as foreign, thus decresing the performance significantly. After the update, performance hit has smoothed:
image

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -558,21 +626,53 @@ open class RouteStep: Codable {
     
     
     /// Used to Decode `Intersection.admin_index`
-    private struct AdministrativeAreaIndex: Codable {
+    private struct AdministrativeAreaIndex: Codable, ForeignMemberContainer {
+        var foreignMembers: JSONObject = [:]
```
1ec5, line 630: As you pointed out in chat, this struct also contributes to the performance hit. I believe this is because it’s an artificial construct that redecodes Intersection, which is already a foreign member container. The `AdministrativeAreaIndex.foreignMembers` property would end up containing a duplicate copy of the whole intersection object.

Sources/MapboxDirections/RouteStep.swift
```diff
@@ -558,21 +626,53 @@ open class RouteStep: Codable {
...
+            try container.encodeIfPresent(administrativeRegionIndex, forKey: .administrativeRegionIndex)
+            
+            try encodeForeignMembers(notKeyedBy: CodingKeys.self, to: encoder)
+        }
     }
     
     /// Used to Decode `Intersection.geometry_index`
-    private struct IntersectionShapeIndex: Codable {
+    private struct IntersectionShapeIndex: Codable, ForeignMemberContainer {
+        var foreignMembers: JSONObject = [:]
```
1ec5, line 655: Same with this struct – it’s a redecoding of Intersection.

1ec5 (commented):
Hopefully eliminating protocol conformance for the [`union`](https://en.wikipedia.org/wiki/Union_type)-style redecoding structs will mitigate the performance hit to an acceptable degree.

Udumft:
After updating IntersectionShapeIndex and AdministrativeAreaIndex, test performance seems to stabilize:

image

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
...
 * `RoadClasses.tunnel` and `RoadClasses.restricted` are no longer supported in `RouteOptions.roadClassesToAvoid` or `RouteOptions.roadClassesToAllow` properties
 * Added `DirectionsOptions(url:)`, `RouteOptions(url:)` and extended existing `DirectionsOptions(waypoints:profileIdentifier:queryItems:)`, `RouteOptions(waypoints:profileIdentifier:queryItems:)`, `MatchOptions(waypoints:profileIdentifier:queryItems:)` and related convenience init methods for deserializing corresponding options object using appropriate request URL or it's query items. ([#655](https://github.com/mapbox/mapbox-directions-swift/pull/655))
+* `RouteResponse`, `RouteRefreshResponse`, `MatchResponse` and all underlying types now correctly handle unrecognized (foreign) JSON values and preserve them on coding/decoding using `ForeignMemberContainer` and `ForeignMemberClassContainer` implementations. ([#669](https://github.com/mapbox/mapbox-directions-swift/pull/669))
+* Fixed `Waypoint.snappedDistance` value was missing after decoding a `RouteResponse`. ([#669](https://github.com/mapbox/mapbox-directions-swift/pull/669))
```
1ec5, line 13:
```suggestion
* Types that correspond to objects in the Mapbox Directions API response, such as `RouteResponse`, `RouteRefreshResponse`, `MatchResponse`, and `RouteStep`, now conform to the `ForeignMemberContainer` and `ForeignMemberClassContainer` protocols. Types that conform to these protocols can persist unrecognized properties in the response, such as properties that are in beta, even after coding and decoding. You can access these properties using the `ForeignMemberContainer.foreignMembers` and `ForeignMemberClassContainer.foreignMembers` properties. ([#669](https://github.com/mapbox/mapbox-directions-swift/pull/669))
* Fixed an issue where decoding a `RouteResponse` incorrectly set the `Waypoint.snappedDistance` property to `nil`. ([#669](https://github.com/mapbox/mapbox-directions-swift/pull/669)) 
```

1ec5 (approved):

PR #672, opened 2022-03-31
Additional Incident members

Udumft:
This PR exposes missing [Incident](https://docs.mapbox.com/api/navigation/directions/#incident-object) members as country codes, closed road flag, long description, etc.


Sources/MapboxDirections/Incident.swift
```diff
@@ -65,6 +72,17 @@ public struct Incident: Codable, Equatable {
...
     }
 
+    private struct CongestionContainer: Codable {
+        enum CodingKeys: String, CodingKey {
+            case value
+        }
+        
+        let value: Int
+        var clampedValue: Int? {
+            value == 101 ? nil : value
```
Udumft, line 82: [The doc](https://docs.mapbox.com/api/navigation/directions/#incident-object) says that value can range from 0 up to 101. Directions team clarified that `101` value indicates something like "invalid state" which effectively means that there is no congestion data available.  So, for the sake of simplifying, and since both cases means the same for the end user, I've added this conversion.

purew: It comes from the LTS telemetry source here

> 0-101 -> higher is more congestion. Special values: 100 -> the road is closed, 101 -> no congestion calculated

https://github.com/mapbox/live-traffic-service-interface/blob/master/service.proto#L18

purew: Oh I see, it's what's stated in the public doc you linked too.

S2Ler: Worth have a named constant for 101 with a link to docs if possible.

S2Ler (approved):

PR #677, opened 2022-04-07
Breaking changes flow

S2Ler:
- Uses `swift package diagnose-api-breaking-changes` to detect breaking changes.
- I wasn't able to setup mbx-ci/gh tools to post pr comments, so if job fails, you will have to go to circleci and check what's wrong. The list of breaking changes will be available in artificats.


S2Ler:
Test

swift-package-baseline/baseline.txt
```diff
@@ -0,0 +1 @@
+v2.4.0-rc.1
```
bamx23, line 1: For this baseline, there should be a baseline file `swift-package-baseline/8284de2e35ce98b414d317f00eed26988d7bbbae/MapboxDirections.json` as `swift-package-baseline/2b14b8c276523a01dc730897593990ef83c71320/MapboxDirections.json` is for release `2.3.0`.

.circleci/config.yml
```diff
@@ -42,6 +42,36 @@ commands:
...
+  detect-breaking-changes:
+    macos:
+      xcode: 13.3.0
+    steps:
+      - checkout
+      - install-mapbox-token
+      - run:
+          name: "Diagnose breaking changes"
+          command: |
+            swift package diagnose-api-breaking-changes --products MapboxDirections --baseline-dir swift-package-baseline --breakage-allowlist-path swift-package-baseline/breakage-allowlist-path.txt $(cat swift-package-baseline/baseline.txt) > diagnose-output.txt || TRUE
```
bamx23, line 54: I would recommend keeping tool output in the CI logs:
```suggestion
            swift package diagnose-api-breaking-changes --products MapboxDirections --baseline-dir swift-package-baseline --breakage-allowlist-path swift-package-baseline/breakage-allowlist-path.txt $(cat swift-package-baseline/baseline.txt) | tee diagnose-output.txt || TRUE
```

.circleci/config.yml
```diff
@@ -157,6 +187,9 @@ jobs:
       - save-cache
 
 workflows:
+  detect-breaking-changes:
+    jobs:
+      - detect-breaking-changes
```
bamx23, line 192: NIT: We use custom names in other jobs
```suggestion
      - detect-breaking-changes:
          name: "Detect Breaking Changes"
```

.circleci/config.yml
```diff
@@ -157,6 +187,9 @@ jobs:
       - save-cache
 
 workflows:
+  detect-breaking-changes:
```
bamx23, line 190: Do we need to have a separate workflow for this task?

bamx23:
We should also add a new step to releasing instructions:
> After releasing a final or RC version, update `swift-package-baseline/baseline.txt`, run `swift package diagnose-api-breaking-changes --products MapboxDirections --baseline-dir swift-package-baseline --breakage-allowlist-path swift-package-baseline/breakage-allowlist-path.txt $(cat swift-package-baseline/baseline.txt)` and remove previous baseline directory(the one with commit hash).

We can also automate this as part of the release process on CI by creating a pipeline that runs this on tag push and creates a PR.

.circleci/config.yml
```diff
@@ -42,6 +42,41 @@ commands:
...
+    macos:
+      xcode: 13.3.0
+    steps:
+      - checkout
+      - install-mapbox-token
+      - run: 
+          name: "Unzip baseline interface"
+          command: |
+            cd swift-package-baseline
+            unzip baseline-interface.zip
```
bamx23, line 55: Don't think zipping is a good idea. If we keep only the latest baseline, this file can be updated (move + several changes for each baseline update).

scripts/breaking-changes/parse-breaking-changes-output.py
```diff
@@ -0,0 +1,24 @@
...
+        sys.exit(0)
+
+    breaking_changes_list = []
+
+    for line in output.split("\n"):
+        if line.find(BREAKING_CHANGES_STRING) != -1:
+            breaking_changes_list.append(line)
+
+    print("\n".join(breaking_changes_list))
+    sys.exit(2)
```
bamx23, line 24: NIT: empty new line
```suggestion
    sys.exit(2)

```

.circleci/config.yml
```diff
@@ -42,6 +42,41 @@ commands:
...
+      - run:
+          name: "Diagnose breaking changes"
+          command: |
+            swift package diagnose-api-breaking-changes --products MapboxDirections --baseline-dir swift-package-baseline --breakage-allowlist-path swift-package-baseline/breakage-allowlist-path.txt $(cat swift-package-baseline/baseline.txt) | tee diagnose-output.txt || TRUE
```
bamx23, line 59: NIT: (for readability)
```suggestion
            swift package diagnose-api-breaking-changes \
                --products MapboxDirections \
                --baseline-dir swift-package-baseline \
                --breakage-allowlist-path swift-package-baseline/breakage-allowlist-path.txt \
                "$(cat swift-package-baseline/baseline.txt)" \
              | tee diagnose-output.txt \
              || TRUE
```

swift-package-baseline/baseline.txt
S2Ler: nice catch, fixed

.circleci/config.yml
S2Ler: need, no. Can, yes) Merged for simplicity.

.circleci/config.yml
S2Ler: The baseline isn't supposed to be changed frequently. Only once per major release. So, if we analyse the size:
- raw: 1.6 MB
- zipped: 74 KB.
- Number of major release before zipped content will become more than first raw (excluding the size increase with raw diffs): 21 Major release.
- Estimated number of years before the parity reached assuming 1 major release per 2 years: 42 years.
- Probability that library will reach 21st major release: ~0

So I would keep zipped content because it has immediate effect on users network traffic consumption.

scripts/breaking-changes/parse-breaking-changes-output.py
S2Ler: Thanks, accepted.

S2Ler:
> We can also automate this as part of the release process on CI by creating a pipeline that runs this on tag push and creates a PR.

As I mentioned in the previous step it is not designed to be updated frequently. The script is added though. CI integration isn't important enough given that you don't need to run it frequently. But can be done in a separate PR if needed.

.circleci/config.yml
1ec5: Large binary files would still have an immediate impact on the time it takes to clone the repository, something SPM does with abandon. Can we host this file anywhere else – say, as an attachment to a release?

S2Ler: I don't think it worth it moving baseline out of the repository. For comparison, one snapshot image in the mapbox-navigation-ios can be bigger than this zipped content. Maps team also stores baseline in the repo zipped: https://github.com/mapbox/mapbox-maps-ios/tree/main/scripts/api-compatibility-check

bamx23: What if we run a separate job or update this one and cache baseline on CI?

And it's not true that the baseline is changed on major releases only. We should update it whenever we add new APIs. So it could be that we need to update the baseline for each `rc.1` (and maybe other RCs and GA if they introduce a new API).

Storing baseline in GitHub release artifacts is an interesting idea. It can be automated later, but it's an easy release step that we should have anyway (it's either updating it within the repo or updating the release on Github). And the CI script is not going to be modified as we can read the tag from `swift-package-baseline/baseline.txt`.

S2Ler: > And it's not true that the baseline is changed on major releases only.

I mean you don't to update it on each API change. In ideal world it should point to first major release and that's it. 

If you are so agains this baseline caching, I can remove it. It is not necessary for the tool to work. Just speeds things up a bit.

bamx23: My point is that if we use `2.0.0` as a baseline, then add a new API in `2.1.0` and break it in `2.2.0`, it's a breaking change for `2.1.0`, but not for `2.0.0`. That's why we should use the latest RC or GA as a baseline in this tool.

S2Ler: Fair point, this will need a redesign of this PR. Coming.

S2Ler: PR reworked:
- README.md updated with new instructions.
- update-version.sh script updated with automatic baseline bump during rc.1 release.
- Baseline isn't cached any more.
- Baseline is updated during each RC release.

scripts/update-baseline.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+set -u
+
+if [[ "$(pwd)" != *mapbox-directions-swift ]]; then
```
bamx23, line 7: Should we go with `SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` as it's not required to have this repo in `mapbox-directions-swift` dir.

scripts/update-version.sh
```diff
@@ -46,3 +46,9 @@ fi
 step "Updating copyright year to ${YEAR}…"
 
 sed -i '' -E "s/© ([0-9]{4})[–-][0-9]{4}/© \\1–${YEAR}/g" LICENSE.md docs/jazzy.yml Sources/MapboxDirections/Info.plist
+
+# Starting with first release candidate, we should not have breaking changes.
+if [[ $1 == *rc.1 ]]; then
```
bamx23, line 51: I would suggest updating the baseline for GA releases too. Not sure about patch releases (this suggestion includes them, maybe it shouldn't and we need `&& $SHORT_VERSION == *.0`). It isn't expected that we add new APIs in RCs, but it might happen.
```suggestion
if [[ $1 == *rc.1 || $SHORT_VERSION == $SEM_VERSION ]]; then
```

swift-package-baseline/README.md
```diff
@@ -0,0 +1,30 @@
...
+This folder contains metadata required for `swift package diagnose-api-breaking-changes` tool:
+- baseline.txt: contains the version of the package that is used as a baseline for detecting breaking changes.
+- breakage-allowlist-path.txt: contains the list of breaking changes that we accepted to exist.
+
+## The procedure of accepting breaking changes
+
+1. Obtain the list of breaking changes from Circle CI "Breaking changes" job using "Artifacts" tab.
+1. Add the list of breaking changes to `breakage-allowlist-path.txt` file removing the preceding emoji. For example, for generated breaking change line of `💔 API breakage: constructor Incident.init(identifier:type:description:creationDate:startDate:endDate:impact:subtype:subtypeDescription:alertCodes:lanesBlocked:shapeIndexRange:) has been removed` the following line should be added to `breakage-allowlist-path.txt`: `API breakage: constructor Incident.init(identifier:type:description:creationDate:startDate:endDate:impact:subtype:subtypeDescription:alertCodes:lanesBlocked:shapeIndexRange:) has been removed`
+
+## Bumping baseline
```
bamx23, line 12: It is worth mentioning that baseline is updated automatically when you run `scripts/update-version.sh` for RCs (and maybe GAs)

bamx23 (approved):
Looks good, just several suggestions.
And it's only 2-3 minutes to run this new CI job without a cached/vendored baseline, nice!

swift-package-baseline/README.md
S2Ler: Included, thanks for suggestion.

bamx23 (approved):

PR #685, opened 2022-04-21
Provide convenience initializers for options

### LINKED ISSUE #684: Compiler error subclassing RouteOptions — required initializer must be provided by subclass
#655 changed the `RouteOptions(waypoints:profileIdentifier:)` initializer to take an additional `queryItems` parameter. Even though a default argument was provided, a subclass such as the navigation SDK’s NavigationRouteOptions would become incompatible and fail to build (albeit with a fix-it suggestion) because the signature differs:

> /path/to/mapbox-navigation-ios/Sources/MapboxCoreNavigation/NavigationRouteOptions.swift:70:1: 'required' initializer 'init(waypoints:profileIdentifier:queryItems:)' must be provided by subclass of 'RouteOptions'

This change probably got missed in code review because it was marked `public` rather than `open`, but it is `required` nonetheless. It would’ve been caught by an API diffing tool: #683.

Can we get around this issue by adding a convenience initializer with the old set of arguments, or do we need to revisit https://github.com/mapbox/mapbox-directions-swift/pull/655#discussion_r819995251? Either way, we need to issue the fix in v2.4.1 per semver rules.

/cc @mapbox/navigation-ios

jill-cardamon:
This PR fixes #684 by adding convenience initializers for `RouteOptions`, `MatchOptions`, and `DirectionsOptions`.

### DIFF
--- a/Sources/MapboxDirections/DirectionsOptions.swift
+++ b/Sources/MapboxDirections/DirectionsOptions.swift
@@ -219,6 +219,18 @@ open class DirectionsOptions: Codable {
     }
     
     /**
+     Initializes an options object for routes between the given waypoints and an optional profile identifier.
+
+     Do not call `DirectionsOptions(waypoints:profileIdentifier:)` directly; instead call the corresponding initializer of `RouteOptions` or `MatchOptions`.
+
+     - parameter waypoints: An array of `Waypoint` objects representing locations that the route should visit in chronological order. The array should contain at least two waypoints (the source and destination) and at most 25 waypoints. (Some profiles, such as `ProfileIdentifier.automobileAvoidingTraffic`, [may have lower limits](https://docs.mapbox.com/api/navigation/#directions).)
+     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. `ProfileIdentifier.automobile` is used by default.
+     */
+    public convenience init(waypoints: [Waypoint], profileIdentifier: ProfileIdentifier? = nil) {
+        self.init(waypoints: waypoints, profileIdentifier: profileIdentifier, queryItems: nil)
+    }
+    
+    /**
      Creates new options object by deserializing given `url`
      
      Initialization fails if it is unable to extract `waypoints` list and `profileIdentifier`. If other properties are failed to decode - it will just skip them.
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -98,6 +98,16 @@ open class RouteOptions: DirectionsOptions {
             self.arriveBy = arriveBy
         }
     }
+    
+    /**
+     Initializes a route options object for routes between the given waypoints and an optional profile identifier.
+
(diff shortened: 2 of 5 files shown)


Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -218,6 +218,18 @@ open class DirectionsOptions: Codable {
         }
     }
     
+    /**
+     Initializes an options object for routes between the given waypoints and an optional profile identifier.
+
+     Do not call `DirectionsOptions(waypoints:profileIdentifier:)` directly; instead call the corresponding initializer of `RouteOptions` or `MatchOptions`.
```
jill-cardamon, line 224: Since developers should not call `Directionsoptions(waypoints:profileIdentifier:)` directly, is it safe to not include this change?

1ec5: We should include it, because it is valid, if somewhat contrived, to subclass DirectionsOptions directly. The use case would be building support for another navigation API service endpoint. Normally that’s something this library would have built right in, but in principle, client code could implement that support all the same.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -98,6 +98,16 @@ open class RouteOptions: DirectionsOptions {
...
         }
     }
+    
+    /**
+     Initializes a route options object for routes between the given waypoints and an optional profile identifier.
+
+     - parameter waypoints: An array of `Waypoint` objects representing locations that the route should visit in chronological order. The array should contain at least two waypoints (the source and destination) and at most 25 waypoints. (Some profiles, such as `ProfileIdentifier.automobileAvoidingTraffic`, [may have lower limits](https://www.mapbox.com/api-documentation/#directions).)
+     - parameter profileIdentifier: A string specifying the primary mode of transportation for the routes. `ProfileIdentifier.automobile` is used by default.
+     */
+    public convenience init(waypoints: [Waypoint], profileIdentifier: ProfileIdentifier? = nil) {
```
1ec5, line 108: Let’s add a test that subclasses RouteOptions and overrides just this initializer, not `init(waypoints:profileIdentifier:queryItems:)`. It doesn’t have to do anything; we just need to see if it would compile. I’m concerned that the `required` bit means it still needs to be implemented even if the convenience initializer would call the required one.

1ec5 (commented):
Remember to add a changelog entry about fixing the incompatibility introduced in #655 by restoring these initializers.

Sources/MapboxDirections/RouteOptions.swift
jill-cardamon: Is it sufficient to subclass `RouteOptions`, override the convenience init, and initialize a `RouteOptionsSubclass` object through the override?

jill-cardamon:
This PR did not fix the stability issues with the Nav SDK. Instead, several patch releases of MapboxCoreNavigation were issued and can be found here: [v2.1.2](https://github.com/mapbox/mapbox-navigation-ios/releases/tag/v2.1.2), [v2.2.1](https://github.com/mapbox/mapbox-navigation-ios/releases/tag/v2.2.1), and [v2.3.2](https://github.com/mapbox/mapbox-navigation-ios/releases/tag/v2.3.2).

PR #687, opened 2022-05-02
Codecov is back!

jill-cardamon:
This PR reenables codecov for mapbox-directions-swift.


jill-cardamon:
Nothing left. Should be good to go. Let me rebase.

PR #688, opened 2022-05-03
Fix crash that occurs when directions are requested without waypoints

### LINKED ISSUE #515: Crash if request Directions without Waypoints
## Problem:

An attempt to request directions without start/end/waypoints results in a crash.

## Where / Repro

Location:
```
let unparameterizedURL = URL(string: includesQuery ? options.path : options.abridgedPath, relativeTo: credentials.host)!
```
https://github.com/mapbox/mapbox-directions-swift/blob/fcc33ed07dba17df88fc6a305ff4732fae81533f/Sources/MapboxDirections/Directions.swift#L529

Triggered through the use of Directions calculate:
https://github.com/mapbox/mapbox-directions-swift/blob/fcc33ed07dba17df88fc6a305ff4732fae81533f/Sources/MapboxDirections/Directions.swift#L148

## Context:
While this is avoidable by inspecting params prior to sending, given the interface definition I would expect this not to crash in the event the minimum requirements were not met.

Could be avoided with a prerequisite check which could then construct invalid input result

eg. inspect input, complete with invalidInput(message: "Missing x/y/z")

jill-cardamon:
This PR fixes #515 by including an assertion failure when any options (`DirectionsOptions` and its affiliated subclasses) are initialized with less than 2 waypoints or coordinates. Some tests that initialize `RouteOptions` with less than 2 waypoints in the Nav SDK will need to be updated.

### DIFF
--- a/Sources/MapboxDirections/Directions.swift
+++ b/Sources/MapboxDirections/Directions.swift
@@ -521,7 +521,6 @@ open class Directions: NSObject {
      - returns: The URL to send the request to.
      */
     open func url(forCalculating options: DirectionsOptions) -> URL {
-        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
         return url(forCalculating: options, httpMethod: "GET")
     }
     
--- a/Sources/MapboxDirections/Directions.swift
+++ b/Sources/MapboxDirections/Directions.swift
@@ -521,6 +521,7 @@ open class Directions: NSObject {
      - returns: The URL to send the request to.
      */
     open func url(forCalculating options: DirectionsOptions) -> URL {
+        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
         return url(forCalculating: options, httpMethod: "GET")
     }
     
@@ -536,8 +537,6 @@ open class Directions: NSObject {
      - returns: The URL to send the request to.
      */
     open func url(forCalculating options: DirectionsOptions, httpMethod: String) -> URL {
-        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
-
         let includesQuery = httpMethod != "POST"
         var params = (includesQuery ? options.urlQueryItems : [])
         params.append(contentsOf: authenticationParams)
@@ -559,6 +558,7 @@ open class Directions: NSObject {
      - returns: A GET or POST HTTP request to calculate the specified options.
      */
     open func urlRequest(forCalculating options: DirectionsOptions) -> URLRequest {
+        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
         let getURL = self.url(forCalculating: options, httpMethod: "GET")
         var request = URLRequest(url: getURL)
         if getURL.absoluteString.count > MaximumURLLength {
--- a/Sources/MapboxDirections/Directions.swift
+++ b/Sources/MapboxDirections/Directions.swift
@@ -536,6 +536,8 @@ open class Directions: NSObject {
      - returns: The URL to send the request to.
      */
     open func url(forCalculating options: DirectionsOptions, httpMethod: String) -> URL {
+        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
+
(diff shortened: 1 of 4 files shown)


Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -129,6 +129,8 @@ open class DirectionsOptions: Codable {
         self.waypoints = waypoints
         self.profileIdentifier = profileIdentifier ?? .automobile
         
+        if waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
```
1ec5, line 132: The `waypoints` property is read-write, so it’s valid and a common pattern to initialize the object first before filling in the waypoints. Instead of asserting here, move the assertion to where we build a request URL in `abridgedPath` or `Directions.url(forCalculating:httpMethod:)`.

https://github.com/mapbox/mapbox-directions-swift/blob/62b9689daa60fe0d9ee245868cb2bdaec3e09c95/Sources/MapboxDirections/DirectionsOptions.swift#L312

1ec5 (commented):
The changelog should mention that there’s now an assertion for this case and when it gets thrown.

Sources/MapboxDirections/DirectionsOptions.swift
bamx23, line 132: This doesn't actually fix the crash. The assertion will be visible only in the debugger, but the app will still crash in production.

Sources/MapboxDirections/Directions.swift
```diff
@@ -536,6 +536,8 @@ open class Directions: NSObject {
      - returns: The URL to send the request to.
      */
     open func url(forCalculating options: DirectionsOptions, httpMethod: String) -> URL {
+        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
```
bamx23, line 539: What URL can we return here? We can't make this method returning optional URL, but there is no valid URL for the case when `options.waypoints.count < 2`.

1ec5: Unfortunately, making the return type optional would be a backwards-incompatible change. Should we shift the check to where Directions calls this method for now?

Sources/MapboxDirections/Directions.swift
```diff
@@ -557,6 +558,7 @@ open class Directions: NSObject {
      - returns: A GET or POST HTTP request to calculate the specified options.
      */
     open func urlRequest(forCalculating options: DirectionsOptions) -> URLRequest {
+        if options.waypoints.count < 2 { assertionFailure("waypoints array requires at least 2 waypoints") }
```
bamx23, line 561: Moving this assertion to these methods doesn't seem to make any difference. We will still crash if `options.waypoints.count < 2` in production.

What if we return an invalid URL here? Or an URL of Directions API but with `{coordinates}` as is including something like `[]` or `empty` if the waypoints array is empty.

jill-cardamon: I think what this does is prevents the crash from happening before we fetch a response. We could provide an invalidInput instead of an assertion failure, but again we'd see this crash in production. Providing an invalid URL could be a good idea. If we provide a URL of the Directions API with empty endpoints, I think we'd see a similar crash.

bamx23: My point here is that SDK should not crash on incorrect input. We have a way to return an error when requesting routes and should not crash there too.

All these cases should be fixed and covered with tests (we can mock a response that happens if you provide 1 or 0 waypoints).

Sources/MapboxDirections/DirectionsOptions.swift
```diff
@@ -431,10 +431,15 @@ open class DirectionsOptions: Codable {
      The path of the request URL, not including the hostname or any parameters.
      */
     var path: String {
-        guard let coordinates = coordinates, !coordinates.isEmpty else {
+        guard let coordinates = coordinates else {
```
bamx23, line 434: By the way, why it's nil? I see that the `coordinates` property can never be `nil` and we can make it non-nil as it's internal.

jill-cardamon:
Sorry @azarovalex I didn't see this notification. Is this still causing any issues on 1Tap?

PR #697, opened 2022-05-24
RouteStep encoding rounding

### LINKED ISSUE #670: RouteStep rounds distances and times when encoding
@Udumft noticed that RouteStep rounds distances to increments of 0.1&nbsp;meters and times to increments of 0.1&nbsp;seconds when encoding to JSON:

https://github.com/mapbox/mapbox-directions-swift/blob/f40fcea9d26fa2cb59570f21546abeffb643a02b/Sources/MapboxDirections/RouteStep.swift#L493-L495

This is not a lot of precision loss, but in principle, this library should be able to round-trip the same number that it got from the Directions API. Otherwise, there could be some inconsistency with other libraries like MapboxNavigationNative and MapboxMaps.

This behavior was introduced in 9cffdd3772668152b8c1e9571fd1687a7d8bd2fd for #382, apparently to make an integration test pass. Most likely, the test had been failing because of an overly strict `XCTAssertEquals(_:_:_:)` call, which should’ve been replaced by either `XCTAssertEqualsWithAccuracy(_:_:accuracy:_:)` or `JSONSerialization.assertObjectsAreEqual(_:_:approximate:)`. The latter method was implemented in 42923910e1d94348f6640f06bc192dd122087198, later in the same PR.

/cc @mapbox/navigation-ios

Udumft:
Resolves #670 
PR removes rounding distance and travel times when encoding RouteStep. Unit test updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes to Mapbox Directions for Swift
 
+## main
+
+* Removed limiting `RouteStep.distance`, `RouteStep.expectedTravelTime` and `RouteStep.typicalTravelTime` precision to 1 decimal digit when being encoded. ([#697](https://github.com/mapbox/mapbox-directions-swift/pull/697))
+
 ## v2.5.0
 
 * Added the `RestStop.name` property. ([#689](https://github.com/mapbox/mapbox-directions-swift/pull/689))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes to Mapbox Directions for Swift
 
+## main
+
+* Removed limiting `RouteStep.distance`, `RouteStep.expectedTravelTime` and `RouteStep.typicalTravelTime` precision to 1 decimal digit when being encoded. ([#697](https://github.com/mapbox/mapbox-directions-swift/pull/697))
```
1ec5, line 5:
```suggestion
* Fixed an issue where `RouteStep.distance`, `RouteStep.expectedTravelTime` and `RouteStep.typicalTravelTime` were rounded to one decimal place when being encoded. ([#697](https://github.com/mapbox/mapbox-directions-swift/pull/697))
```

1ec5 (approved):
Thanks!

PR #702, opened 2022-05-30
Bump Turf requirement to v2.4

### LINKED ISSUE #701: Update minimum Turf version to v2.4.0
MapboxDirections depends on new types in v2.4.0 and we should bump minimum version.

S2Ler:
Fixes #701


Udumft (approved):
Probably should also update `.resolved` files

Udumft:
Oh, sorry, I didn't notice that. Then it LGTM

PR #704, opened 2022-06-01
Incidents refreshing

Udumft:
This PR adds `Route.refreshLegIncidents` method to refresh incidents data. Unit tests updated

### DIFF
--- a/Sources/MapboxDirections/RouteRefreshSource.swift
+++ b/Sources/MapboxDirections/RouteRefreshSource.swift
@@ -15,11 +15,18 @@ public protocol RouteLegRefreshSource {
     var refreshedIncidents: [Incident]? { get }
 }
 
+public extension RouteLegRefreshSource {
+    var refreshedIncidents: [Incident]? {
+        return nil
+    }
+}
+
 extension Route: RouteRefreshSource {
     public var refreshedLegs: [RouteLegRefreshSource] {
         legs
     }
 }
+
 extension RouteLeg: RouteLegRefreshSource {
     public var refreshedAttributes: Attributes {
         attributes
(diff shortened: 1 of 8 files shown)


Udumft:
PR has a breaking change:
```
  💔 API breakage: var RouteLegRefreshSource.refreshedIncidents has been added as a protocol requirement
```
I don't think there is a way to avoid that, since the meaning of this PR is to add more data to be refreshed.

chezzdev:
> I don't think there is a way to avoid that, since the meaning of this PR is to add more data to be refreshed.

Not sure if this tool will accept this, but technically if we provide a default implementation for the getter, this property would no longer be required to implement.

Udumft:
But should we add a default value? I'd like to avoid defaulting everything we add.
I guess not many users actually implement `RouteLegRefreshSource`. Do we know know someone who definitely does that? If so, though it is a breaking change, it is quite simple to update for it.

chezzdev:
This particular protocol may not be implemented by users, but it's much easier to stick to the rule than think about it every time. If we have the chance to avoid breaking the API, I'd try it.

Sources/MapboxDirections/RouteRefreshSource.swift
```diff
@@ -31,6 +36,10 @@ extension RefreshedRoute: RouteRefreshSource {
     }
 }
 extension RefreshedRouteLeg: RouteLegRefreshSource {
+    public var refreshedIncidents: [Incident]? {
+        incidents
+    }
+    
     public var refreshedAttributes: RouteLeg.Attributes {
         attributes
     }
```
bamx23, line 45: NIT: it would be nice to keep the order of methods the same
```suggestion
    public var refreshedAttributes: RouteLeg.Attributes {
        attributes
    }

    public var refreshedIncidents: [Incident]? {
        incidents
    }
```

bamx23 (approved):

PR #706, opened 2022-06-06
Traffic control devices flags

Udumft:
This PR adds traffic control devices flags `Intersection object`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * MapboxDirections now requires [Turf v2.4](https://github.com/mapbox/turf-swift/releases/tag/v2.4.0). ([#703](https://github.com/mapbox/mapbox-directions-swift/pull/703))
 * Added the `RouteOptions.maximumWeight` property to compute routes that excludes roads with a lower weight limit. ([#694](https://github.com/mapbox/mapbox-directions-swift/pull/694))
 * Added `Route.refreshLegIncidents(from:)` method to allow refreshing `RouteLeg.incidents` during a route refresh. ([#704](https://github.com/mapbox/mapbox-directions-swift/pull/704))
+* Added `railwayCrossing`, `trafficSignal`, `stopSign` and `yieldSign` properties to `Intersection` object to flag corresponding traffic control devices installed. ([#706](https://github.com/mapbox/mapbox-directions-swift/pull/706))
 
 ## v2.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * MapboxDirections now requires [Turf v2.4](https://github.com/mapbox/turf-swift/releases/tag/v2.4.0). ([#703](https://github.com/mapbox/mapbox-directions-swift/pull/703))
 * Added the `RouteOptions.maximumWeight` property to compute routes that excludes roads with a lower weight limit. ([#694](https://github.com/mapbox/mapbox-directions-swift/pull/694))
 * Added `Route.refreshLegIncidents(from:)` method to allow refreshing `RouteLeg.incidents` during a route refresh. ([#704](https://github.com/mapbox/mapbox-directions-swift/pull/704))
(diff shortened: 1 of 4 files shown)


Udumft:
Accepted API breakage:
```
💔 API breakage: constructor Intersection.init(location:headings:approachIndex:outletIndex:outletIndexes:approachLanes:usableApproachLanes:preferredApproachLanes:usableLaneIndication:outletRoadClasses:tollCollection:tunnelName:restStop:isUrban:regionCode:outletMapboxStreetsRoadClass:) has been removed
```
Added new parameters to this constructor with default values.

Sources/MapboxDirections/Intersection.swift
```diff
@@ -42,6 +46,10 @@ public struct Intersection: ForeignMemberContainer {
         self.restStop = restStop
         self.regionCode = regionCode
         self.outletMapboxStreetsRoadClass = outletMapboxStreetsRoadClass
+        self.railwayCrossing = railwayCrossing
+        self.trafficSignal = trafficSignal
+        self.stopSign = stopSign
+        self.yieldSign = yieldSign
```
1ec5, line 52: There’s a mix of American English and British English vocabulary here:

* American English: railroad crossing, yield sign
* British English: railway crossing, give way sign

We should be consistent in these names to reduce confusion. This package has been using American English spelling and vocabulary for the most part, except for the occasional oversight perhaps. But if “railroad crossing” is confusing or strange to an international audience, “level crossing” would be a decent compromise, as a more specific technical term.

Tests/MapboxDirectionsTests/IntersectionTests.swift
```diff
@@ -19,6 +19,10 @@ class IntersectionTests: XCTestCase {
                     "type": "toll_booth",
                     "name": "test toll booth"
                 ],
+                "railway_crossing": true,
+                "traffic_signal": true,
+                "stop_sign": false,
+                "yield_sign": false
```
1ec5, line 25: FYI, some of these combinations are possible, like a railroad crossing controlled by a traffic signal at an intersection. However, some combinations don’t make much sense, like both a stop sign and a yield sign. Unfortunately, a series of independent properties is probably the best we can do. An OptionSet would prevent some of these valid combinations and constrain us in the future if we need to add other forms of traffic control devices or other details about these devices, such as the kind of traffic signal (ramp meter, blink-mode, etc.).

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * MapboxDirections now requires [Turf v2.4](https://github.com/mapbox/turf-swift/releases/tag/v2.4.0). ([#703](https://github.com/mapbox/mapbox-directions-swift/pull/703))
 * Added the `RouteOptions.maximumWeight` property to compute routes that excludes roads with a lower weight limit. ([#694](https://github.com/mapbox/mapbox-directions-swift/pull/694))
 * Added `Route.refreshLegIncidents(from:)` method to allow refreshing `RouteLeg.incidents` during a route refresh. ([#704](https://github.com/mapbox/mapbox-directions-swift/pull/704))
+* Added `railwayCrossing`, `trafficSignal`, `stopSign` and `yieldSign` properties to `Intersection` object to flag corresponding traffic control devices installed. ([#706](https://github.com/mapbox/mapbox-directions-swift/pull/706))
```
1ec5, line 8:
```suggestion
* Added the `Intersection.railwayCrossing`, `Intersection.trafficSignal`, `Intersection.stopSign` and `Intersection.yieldSign` properties to indicate the kind of traffic control devices at an intersection along the route. ([#706](https://github.com/mapbox/mapbox-directions-swift/pull/706))
```

Sources/MapboxDirections/Intersection.swift
Udumft: I didn't even think about it, just reflected original names from Directions. Thanks for clarification.

ShanMa1991 (approved):

PR #709, opened 2022-06-08
Add a job that will build documentation and create PR

### LINKED ISSUE #651: Generate and publish documentation on CI
We should adapt the work @chezzdev did in mapbox/mapbox-navigation-ios#3439 mapbox/mapbox-navigation-ios#3469 to run the jazzy documentation generation script on CI for every PR and push it to publisher-production for release tags. This will completely automate two of the steps of [the release process](https://github.com/mapbox/mapbox-directions-swift/blob/main/CONTRIBUTING.md#releasing-a-new-version). It would also obviate #650, since no one would need to run publish-documentation.sh manually anymore.

/cc @mapbox/navigation-ios @mapbox/docs

S2Ler:
I'm no longer able to run jazzy locally and decided to move the task to a stable environment. 

Fixes #651, although #651 says that docs should be published for each PR. I don't think it is necessary. But can be solved separately.

- [x] Update [Contributing Guide](https://github.com/mapbox/mapbox-directions-swift/blob/main/CONTRIBUTING.md) after general idea is accepted.
- [x] Test this job e2e.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,8 +82,9 @@ To release a new version of the MapboxDirections package:
 1. Tag the merged changes as `v#.#.#`. Push the tag by running `git pull && git push origin v#.#.#`.
 1. [Create a new release](https://github.com/mapbox/mapbox-directions-swift/releases/new/). Add release notes based on the release’s section in the changelog. (Unlike the changelog, release notes accept `#123` syntax for linking to PRs.) Title the release `v#.#.#`. Check “This is a pre-release” if applicable, then click “Publish release”.
 1. Run `pod repo update && pod trunk push MapboxDirections.podspec` (or `pod trunk push MapboxDirections-pre.podspec` for a prerelease) to publish the release on CocoaPods trunk.
-1. Run `./scripts/publish-documentation.sh v#.#.#` to generate and publish the documentation. Create a pull request and set the base branch to `publisher-production`.
-1. Wait for new documentation to be live. Once you merge the branch into `publisher-production`, the new version will be available within 10 minutes. (Mapbox employees can check the #publisher channel in Slack for a notification of when the commit has been published.)
+1. Open [CircleCI](https://app.circleci.com/pipelines/github/mapbox/mapbox-directions-swift?branch=main) and approve "Approve Publish Documentation" workflow for the release tag to publish documentation.
+1. Make sure that job runs successfully.
+1. Wait for new documentation to be live. The new version will be available within 10 minutes after job succeeds. (Mapbox employees can check the #publisher channel in Slack for a notification of when the commit has been published.)
 1. _(Mapbox employees only.)_ [Update various links](https://github.com/mapbox/ios-sdk#mapboxdirectionsswift) to the current docset in the [iOS documentation] site.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -192,6 +204,26 @@ jobs:
...
+    steps:
+      - checkout
+      - install-mapbox-token
+      - install-carthage
+      - install-mbx-ci
+      - install-gh
+      - run:
+          name: "Publish Documentation"
+          command: |
+            ./scripts/publish-documentation.sh << pipeline.git.tag >>
```
bamx23, line 219: I would recommend splitting this step into several separate ones:
* Build Documentation
* Setup Repo Write Access
* Create Docs PR

.circleci/config.yml
```diff
@@ -192,6 +204,26 @@ jobs:
...
+            git remote set-url origin "https://x-access-token:$(mbx-ci github writer public token)@github.com/mapbox/mapbox-directions-swift.git"
+            git config user.email "release-bot@mapbox.com"
+            git config user.name "Mapbox Releases"
+            git push --set-upstream origin docs-<< pipeline.git.tag >>
+            export GITHUB_TOKEN="$(mbx-ci github writer public token)"
+            gh pr create --reviewer "mapbox/navigation-ios" --fill
```
bamx23, line 225: By the way, do we need a PR? Why dot pushing directly to `publisher-production` as we do in Navigation SDK repo?

bamx23, line 225: Does this command create a PR to `publisher-production` branch? And I'm not sure that reviewer group will be added (there was an issue with this for `mbx-ci` generated tokens)

.circleci/config.yml
```diff
@@ -219,3 +251,13 @@ workflows:
...
+      - approve-publish-documentation:
+          name: "Approve Publish Documentation"
+          type: approval
+          filters:
+            tags:
+              only: /^v\d+\.\d+\.\d+(-.+)?$/
+      - publish-documentation:
+          name: "Publish Documentation"
+          requires:
+            - "Approve Publish Documentation"
```
bamx23, line 263: We can run this for all valid version tags (without approval)

scripts/publish-documentation.sh
```diff
@@ -35,6 +35,7 @@ mkdir -p "./$FOLDER"
 mv -v $OUTPUT/* "./$FOLDER"
 
 step "Switching branch to publisher-production"
+git checkout Gemfile.lock
```
bamx23, line 38: Should we remove `--depth=1 --prune` from `git fetch --depth=1 --prune` above _(line 17)_?

.circleci/config.yml
```diff
@@ -219,3 +253,13 @@ workflows:
           name: "SPM Ubuntu build"
       - example-app-build:
           name: "Build example app"
+      - approve-publish-documentation:
+          name: "Approve Publish Documentation"
```
bamx23, line 257: Just as an idea: 
* We can build docs for every commit and store it [in artifacts](https://github.com/mapbox/mapbox-navigation-ios/blob/fc37d6661e7ef73959e0a30108f00b0e487e281b/.circleci/config.yml#L648-L650)
* Then there is an approval job for publishing (that works is for tags only)
* And once approved we can push docs to `publisher-production`

scripts/publish-documentation.sh
1ec5: Yes, please remove the shallow clone. I always avoided the publish-documentation.sh script because this step trashes my clone.

Udumft (approved):
LGTM

PR #710, opened 2022-06-23
Add simulator to user agent string in case if it's present.

MaximAlien:
PR adds simulator to user agent string in case if it's present.


MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -1056,8 +1056,8 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = DA1A10C41D00F8FF009F82FA /* Build configuration list for PBXNativeTarget "MapboxDirectionsMac" */;
 			buildPhases = (
-				DA1A10AA1D00F8FF009F82FA /* Sources */,
```
MaximAlien, line 1059: While using Xcode 13.4.1 build issue sometimes occurs because of configuration list sequence.

1ec5 (approved):

PR #720, opened 2022-08-11
Bump Turf version to 2.5

azarovalex:
https://github.com/mapbox/turf-swift/releases/tag/v2.5.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes to Mapbox Directions for Swift
 
+## v2.7.0
+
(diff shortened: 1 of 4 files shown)


1ec5 (commented):
Turf v2.5.0 doesn’t introduce any new APIs that MapboxDirections depends on, so I think we should keep the minimum dependency version at v2.4.0. When we do start to need something new from Turf, we should coordinate the dependency change with the map SDK. I don’t think we’d expect an SPM conflict like mapbox/mapbox-navigation-ios#4047, since both MapboxDirections and MapboxMaps use the `upToNextMajor` operator with Turf, but consistency would give us some peace of mind.

In the meantime, anyone updating their package dependencies should already be getting Turf v2.5.0 automatically. For this PR, the Package.resolved change can be merged without the other changes.

Issue #564, opened 2021-07-20
RouteOptions and Waypoint should be structs


1ec5:
RouteOptions, MatchOptions, Waypoint, and Tracepoint should all be structs, not classes. We pass these types around as if they’re grab bags of values; ideally, they would have no identity apart from their values.

## Background

This library has always represented the set of options for an API request as an object, even though a struct would be a more correct representation of pass-by-value semantics. We kept RouteOptions as a class because the library had to be compatible with Objective-C code; Swift structs don’t bridge to Objective-C. From there, we started using inheritance to add flexibility to the request functionality. To support the Map Matching API, #236 introduced DirectionsOptions as an abstract base class for RouteOptions and MatchOptions. Meanwhile, mapbox/mapbox-navigation-ios#531 added a subclass of RouteOptions, NavigationRouteOptions, to provide good default values optimized for turn-by-turn navigation.

Dropping Objective-C compatibility in #382 gave us an opportunity to convert RouteOptions and MatchOptions into structs. It’s more complicated than that, because waypoints also had to be converted to structs at the same time. It turns out that the Directions API has two different kinds of waypoints, with disjoint sets of information: one that goes in requests and another that goes in responses.

## Next steps

#388 replaces the Waypoint class with separate structs for request and response waypoints, with a goal of eventually converting RouteOptions to a struct as well. This PR has the potential to complicate client code, particularly in the navigation SDK, because client code now has to distinguish between request and response waypoints and potentially convert between the two at various times. Unfortunately, we were unable to prioritize landing this PR for a long time due to unrelated constraints, but the use of classes for input types remains an impediment to adopting more server API features and fully supporting client customization.

/cc @mapbox/navigation-ios

1ec5:
This is one half of #392: that issue also tracks eliminating the DirectionsResult abstract base class, which is also desirable but not as urgent.

PR #728, opened 2022-08-20
Accept request URL in lieu of configuration file

### LINKED ISSUE #692: Command line tool should accept URL as input
The command line tool should accept a Mapbox Directions API request URL as input instead of a JSON file. This functionality was previously implemented in #580, which was obsoleted by #655.

/ref #476
/cc @jill-cardamon @Udumft

### LINKED ISSUE #724: Unable to configure API endpoint using command line tool
The command line tool reads the access token from the `MAPBOX_ACCESS_TOKEN` environment variable, but an access token is only valid for a particular API endpoint. The tool should also read the API endpoint from the environment, defaulting to api.mapbox.com.

This code needs to pass the API endpoint into the `Credentials(accessToken:host:)` initializer:

https://github.com/mapbox/mapbox-directions-swift/blob/5af910694c7459a066b29939c112ccf9a334ffc6/Sources/MapboxDirectionsCLI/CodingOperation.swift#L8-L12

While we’re at it, the `MAPBOX_ACCESS_TOKEN` environment variable isn’t documented in `--help`.

/cc @mapbox/navigation-ios @purew

1ec5:
Overhauled the command line tool’s configuration option. Instead of the `--config` option that takes a JSON configuration file path, the tool now takes a single positional argument that can either be a JSON configuration file path or a Directions API or Map Matching API request URL. When you pass in a request URL, the URL is round-tripped through a DirectionsOptions object back to a URL, exercising the entire typical code path (and then some).

But if you do decide to pass in a JSON configuration file instead of a URL, the command line tool now respects the `MAPBOX_HOST` environment variable alongside the `MAPBOX_ACCESS_TOKEN` environment variable. An access token is scoped to a single API endpoint, so this configuration option makes it possible to run the tool against a proxy or staging server for testing purposes.

`RouteOptions(url:)` now returns `nil` if given a Mapbox Map Matching API request URL, and `MatchOptions(url:)` returns `nil` if given a Mapbox Directions API request URL. Previously, you could mix and match URLs to convert between RouteOptions and MatchOptions. Now, you’ll need to use JSONEncoder and JSONDecoder as well: 

```swift
guard let url = URL(string: "https://api.mapbox.com/directions/v5/mapbox/driving-traffic/…"),
    let routeOptions = RouteOptions(url: url),
    let encodedOptions = try? JSONEncoder().encode(self) else { return }
try JSONDecoder().decode(MatchOptions.self, from: encodedOptions)
```

#564 would unblock a more convenient conversion workflow.

Some tangential changes along the way:

* When the `MAPBOX_ACCESS_TOKEN` environment variable is unset, the tool exits with an error code instead of crashing.
* When a JSON configuration file path is passed into the tool, any tilde is expanded during validation, fixing an issue where an otherwise valid file path got rejected.
* Removed mentions of the `--waypoints` and `--url` options from the command line tool’s documentation that probably came from #580.

Fixes #692 and fixes #724. Supersedes #580. Depends on #727.

/cc @mapbox/navigation-ios @purew

### DIFF
--- a/Sources/MapboxDirectionsCLI/CodingOperation.swift
+++ b/Sources/MapboxDirectionsCLI/CodingOperation.swift
@@ -118,13 +118,17 @@ class CodingOperation<ResponseType : Codable & DirectionsResultsProvider, Option
         return interpolatedCoordinates
     }
     
-    private func requestResponse(_ directionsOptions: OptionsType) -> (Data) {
+    private func response(fetching directionsOptions: OptionsType) -> (Data) {
+        let directions = Directions(credentials: credentials)
+        let url = directions.url(forCalculating: directionsOptions)
+        return response(fetching: url)
+    }
+    
+    private func response(fetching url: URL) -> Data {
         let semaphore = DispatchSemaphore(value: 0)
         
         var responseData: Data!
         
-        let directions = Directions(credentials: credentials)
-        let url = directions.url(forCalculating: directionsOptions)
         let urlSession = URLSession(configuration: .ephemeral)
 
         let task = urlSession.dataTask(with: url) { (data, response, error) in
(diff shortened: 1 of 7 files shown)


Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -27,16 +26,34 @@ struct ProcessingOptions: ParsableArguments {
...
 struct Command: ParsableCommand {
+    static var credentials: Credentials {
+        get throws {
+            guard let accessToken = ProcessInfo.processInfo.environment["MAPBOX_ACCESS_TOKEN"] ??
+                    UserDefaults.standard.string(forKey: "MBXAccessToken") else {
+                throw ValidationError("A Mapbox access token is required. Go to <https://account.mapbox.com/access-tokens/>, then set the MAPBOX_ACCESS_TOKEN environment variable to your access token.")
+            }
+            
+            let hostURL: URL?
+            if let host = ProcessInfo.processInfo.environment["MAPBOX_HOST"] ??
```
1ec5, line 37: The name of this environment variable is consistent with the [Mapbox Python SDK](https://github.com/mapbox/mapbox-sdk-py/blob/e56127b96eea150831c84b67a779378598790fd4/mapbox/services/base.py#L85), which similarly reads credentials from the environment. It differs from the [`MGLMapboxAPIBaseURL`](https://github.com/mapbox/mapbox-directions-swift/blob/2db7fda6029a7c9987cfb9fc7975061891820ca3/Sources/MapboxDirections/Credentials.swift#L8) Info.plist key that applications would set, but the naming conventions for environment variables are different than for Info.plist keys anyways.

Sources/MapboxDirectionsCLI/main.swift
```diff
@@ -4,14 +4,13 @@ import Foundation
...
 
-
 struct ProcessingOptions: ParsableArguments {
     
     @Option(name: [.short, .customLong("input")], help: "[Optional] Filepath to the input JSON. If no filepath provided - will fall back to Directions API request using locations in config file.")
     var inputPath: String?
     
-    @Option(name: [.short, .customLong("config")], help: "Filepath to the JSON, containing serialized Options data.")
-    var configPath: String
+    @Argument(help: "Path to a JSON file containing serialized RouteOptions or MatchOptions properties, or the full URL of a Mapbox Directions API or Mapbox Map Matching API request.")
```
1ec5, line 12: This is kind of unwieldy as an argument description. A possible future improvement would be to accept multiple configurations, perhaps a mix of files and URLs, and merge them somehow.

Sources/MapboxDirectionsCLI/CodingOperation.swift
```diff
@@ -146,28 +146,52 @@ class CodingOperation<ResponseType : Codable & DirectionsResultsProvider, Option
...
+        let requestURL: URL
+        if FileManager.default.fileExists(atPath: (options.config as NSString).expandingTildeInPath) {
+            // Assume the file is a configuration JSON file. Convert it to an options object.
+            let configData = FileManager.default.contents(atPath: (options.config as NSString).expandingTildeInPath)!
+            let decoder = JSONDecoder()
+            directions = Directions(credentials: credentials)
+            directionsOptions = try decoder.decode(OptionsType.self, from: configData)
+        } else if let url = URL(string: options.config) {
+            // Try to convert the URL to an options object.
+            if let parsedOptions = (RouteOptions(url: url) ?? MatchOptions(url: url)) as? OptionsType {
```
1ec5, line 169: This is why `DirectionsOptions(url:)` has to return `nil` when given a URL whose path conflicts with the `abridgedPath` property. If it doesn’t return `nil`, then either RouteOptions and MatchOptions would be equally likely to initialize based on the same URL, mooting the coalescing operator.

Maybe it would be cleaner to nix the `OptionsType` generic parameter in favor of an enumeration with associated values that gets passed into this method, since the two APIs aren’t all that different. But I’ll leave it for now in case another API needs this kind of flexibility.

jill-cardamon (approved):

Sources/MapboxDirectionsCLI/main.swift
jill-cardamon: will you document that possible future improvement?

1ec5: Let’s track it as tail work in #729.

PR #733, opened 2022-09-09
Support `current_geometry_index` in route refresh

chezzdev:


Sources/MapboxDirections/Directions.swift
```diff
@@ -439,8 +439,28 @@ open class Directions: NSObject {
...
      */
     @discardableResult open func refreshRoute(responseIdentifier: String, routeIndex: Int, fromLegAtIndex startLegIndex: Int = 0, completionHandler: @escaping RouteRefreshCompletionHandler) -> URLSessionDataTask? {
+        _refreshRoute(responseIdentifier: responseIdentifier,
+                      routeIndex: routeIndex,
+                      fromLegAtIndex: startLegIndex,
+                      currentRouteShapeIndex: nil,
+                      completionHandler: completionHandler)
+    }
+
+    @discardableResult open func refreshRoute(responseIdentifier: String, routeIndex: Int, fromLegAtIndex startLegIndex: Int = 0, currentRouteShapeIndex: Int, completionHandler: @escaping RouteRefreshCompletionHandler) -> URLSessionDataTask? {
```
1ec5, line 449: I assume we’re overloading the method instead of giving `currentRouteShapeIndex` a default value to avoid changing the method signature. We should document the new method and explain the difference between the two.

1ec5 (commented):
Remember to add a changelog entry for the new publicly documented methods and also add tests to ensure we don’t introduce off-by-one errors.

1ec5 (approved):

PR #734, opened 2022-09-12
v2.7.0-rc.1

ShanMa1991:
This Pr is to update MapboxDirections for Swift to v2.7.0-rc.1


MaximAlien (approved):

MaximAlien:
I think such releases should be considered to be part of release train as well. 
/cc @SevaZhukov

PR #736, opened 2022-09-13
Remove iOS 11 support.

MaximAlien:
PR increases minimum deployment targets to:
- iOS 12.0
- tvOS 12.0
- macOS 10.14
- watchOS 5.0

PR #737, opened 2022-09-14
Update dependency.

ShanMa1991:
This PR is to update the dependency as following:
Turf for Swift to v2.5.0, Polyline to v5.1.0


ShanMa1991:
Close because #736 works to update the minimum iOS requirement and update the dependency.

PR #738, opened 2022-09-16
Add documentation to `RouteResponse` and `MapMatchingResponse`

### LINKED ISSUE #732: RouteResponse and MapMatchingResponse are undocumented
RouteResponse and MapMatchingResponse aren’t documented, so they don’t show up at all in the jazzy-generated API reference:

https://github.com/mapbox/mapbox-directions-swift/blob/8ded1f608508f6e14bc6914f6cf2b64e3414940b/Sources/MapboxDirections/RouteResponse.swift#L12 https://github.com/mapbox/mapbox-directions-swift/blob/8ded1f608508f6e14bc6914f6cf2b64e3414940b/Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift#L7

/cc @Udumft

jill-cardamon:
This PR fixes #732 by adding documentation comments to `RouteResponse` and `MapMatchingResponse` that will show up in the jazzy-generated API reference.

### DIFF
--- a/Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
+++ b/Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
@@ -8,9 +8,6 @@ import Turf
  A `MapMatchingResponse` object is a structure that corresponds to a map matching response returned by the Mapbox Map Matching API.
  */
 public struct MapMatchingResponse: ForeignMemberContainer {
-    /**
-     Unrecognized properties that are documented as "beta" properties in the Mapbox Map Matching API response.
-     */
     public var foreignMembers: JSONObject = [:]
     
     /**
--- a/Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
+++ b/Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
@@ -4,15 +4,39 @@ import FoundationNetworking
 #endif
 import Turf
 
+/**
+ A `MapMatchingResponse` object is a structure that corresponds to a map matching response returned by the Mapbox Map Matching API.
+ */
 public struct MapMatchingResponse: ForeignMemberContainer {
+    /**
+     Unrecognized properties that are documented as "beta" properties in the Mapbox Map Matching API response.
+     */
     public var foreignMembers: JSONObject = [:]
     
+    /**
+     The raw HTTP response from the Map Matching API.
+     */
     public let httpResponse: HTTPURLResponse?
     
+    /**
+     An array of `Match` objects.
+     */
     public var matches : [Match]?
+    
+    /**
+     An array of `Tracepoint` objects that represent the location an input point was matched with, in the order in which they were matched.
+     This property will be `nil` if a trace point is omitted by the Map Matching API because it is an outlier.
+     */
     public var tracepoints: [Tracepoint?]?
     
+    /**
+     The criteria for the map matching response.
+     */
     public let options: MatchOptions
+    
+    /**
+     The credentials used to make the request.
+     */
     public let credentials: Credentials
     
     /**
--- a/Sources/MapboxDirections/RouteResponse.swift
+++ b/Sources/MapboxDirections/RouteResponse.swift
@@ -13,9 +13,6 @@ public enum ResponseOptions {
  A `RouteResponse` object is a structure that corresponds to a directions response returned by the Mapbox Directions API.
  */
 public struct RouteResponse: ForeignMemberContainer {
-    /**
-     Unrecognized properties that are documented as "beta" properties in the Mapbox Directions API response.
-     */
     public var foreignMembers: JSONObject = [:]
     
     /**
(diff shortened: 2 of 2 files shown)


Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -4,15 +4,39 @@ import FoundationNetworking
 #endif
 import Turf
 
+/**
+ A `MapMatchingResponse` object is a structure that corresponds to a map matching response returned by the Mapbox Map Matching API.
+ */
 public struct MapMatchingResponse: ForeignMemberContainer {
+    /**
+     Unrecognized properties that are documented as "beta" properties in the Mapbox Map Matching API response.
+     */
```
1ec5, line 13: This property is already documented by way of `ForeignMemberContainer`, so unless there’s something unique about foreign members in the context of a top-level response object, we can omit the documentation comment for this particular property. (Same for `RouteResponse.foreignMembers`.)

https://github.com/mapbox/mapbox-directions-swift/blob/3e0d1ac37aca0489965028e4521623531a02cb7d/Sources/MapboxDirections/Extensions/ForeignMemberContainer.swift#L57-L61

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
```diff
@@ -4,15 +4,39 @@ import FoundationNetworking
...
     
+    /**
+     The raw HTTP response from the Map Matching API.
+     */
     public let httpResponse: HTTPURLResponse?
     
+    /**
+     An array of `Match` objects.
+     */
     public var matches : [Match]?
```
1ec5, line 24: I wonder under what circumstances the Map Matching API will return a response with `matchings` set to `null`, resulting in this property being set to `nil`. If we don’t know, then we can leave that case unmentioned.

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -9,20 +9,47 @@ public enum ResponseOptions {
...
     public let httpResponse: HTTPURLResponse?
     
+    /**
+     The response identifier used to request the route.
+     */
     public let identifier: String?
+    
+    /**
+     An array of `Route` objects ordered by descending recommendation rank. This property may contain at most two `Route`s.
+     */
```
1ec5, line 33:
```suggestion
    /**
     An array of `Route` objects sorted from most recommended to least recommended. A route may be highly recommended based on characteristics such as expected travel time or distance. This property contains a maximum of two `Route`s.
     */
```

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -9,20 +9,47 @@ public enum ResponseOptions {
...
     public var routes: [Route]? {
         didSet {
             updateRoadClassExclusionViolations()
         }
     }
+    
+    /**
+     An array of `Waypoint` objects in the order of the input coordinates. Each `Waypoint` is an input coordinate snapped to the road and path network.
+     */
```
1ec5, line 42:
```suggestion
    /**
     An array of `Waypoint` objects in the order of the input coordinates. Each `Waypoint` is an input coordinate snapped to the road and path network.
     
     This property omits the waypoint corresponding to any waypoint in `RouteOptions.waypoints` that has `Waypoint.separatesLegs` set to `true`.
     */
```

Sources/MapboxDirections/RouteResponse.swift
```diff
@@ -9,20 +9,47 @@ public enum ResponseOptions {
...
     public var foreignMembers: JSONObject = [:]
     
+    /**
+     The raw HTTP response from the Directions API.
+     */
     public let httpResponse: HTTPURLResponse?
     
+    /**
+     The response identifier used to request the route.
+     */
```
1ec5, line 28: I could imagine a scenario where the developer finds it convenient to set this property to an internally generated ID when converting a third-party routing engine’s response to a RouteResponse, but then the navigation SDK would reuse that ID during route refreshing, which could cause an error.

```suggestion
    /**
     The unique identifier that the Mapbox Directions API has assigned to this response.
     */
```

1ec5 (approved):

Sources/MapboxDirections/RouteResponse.swift
1ec5: The API intentionally includes an array instead of two separate properties so that it can expand to multiple routes in the future. I’m not sure if we should even say how many would be returned.

```suggestion
    /**
     An array of `Route` objects sorted from most recommended to least recommended.
     
     If `RouteOptions.includesAlternativeRoutes` is set to `true`, this property contains more than one `Route` object. A route may be highly recommended based on factors such as the expected travel time or distance.
     */
```

Sources/MapboxDirections/MapMatching/MapMatchingResponse.swift
jill-cardamon: I'll check if that's possible and update accordingly.

Sources/MapboxDirections/RouteResponse.swift
1ec5: mapbox/mapbox-navigation-ios#1372 demonstrates that the Directions API at one point returned up to three routes.

PR #745, opened 2022-09-21
Add `Waypoint.layer` property

jill-cardamon:
Added optional interger-typed layer property to `Waypoint` class corresponding to the [layers](https://docs.mapbox.com/api/navigation/directions/#retrieve-directions) URL query parameter.

### DIFF
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -435,7 +435,7 @@ open class RouteOptions: DirectionsOptions {
             params.append(URLQueryItem(name: CodingKeys.waypointTargets.stringValue, value: targetCoordinates))
         }
         
-        if waypoints.first(where: { $0.layer != nil} ) != nil {
+        if waypoints.contains(where: { $0.layer != nil }) {
             let layers = waypoints.map { $0.layer?.description ?? "" }.joined(separator: ";")
             params.append(URLQueryItem(name: CodingKeys.layers.stringValue, value: layers))
         }
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -92,13 +92,10 @@ open class RouteOptions: DirectionsOptions {
             self.maximumWeight = Measurement(value: doubleValue, unit: UnitMass.metricTons)
         }
         
-//        if waypoints.first(where: { $0.layer != nil} ) != nil {
-//            let layers = waypoints.map { $0.layer?.description ?? ";" }.joined()
-//            params.append(URLQueryItem(name: CodingKeys.layers.stringValue, value: layers))
-//        }
-        
         if let mappedValue = mappedQueryItems[CodingKeys.layers.stringValue] {
-            waypoints.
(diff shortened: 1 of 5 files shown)


Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -92,6 +92,11 @@ open class RouteOptions: DirectionsOptions {
             self.maximumWeight = Measurement(value: doubleValue, unit: UnitMass.metricTons)
         }
         
+        if let mappedValue = mappedQueryItems[CodingKeys.layers.stringValue] {
+            zip(waypoints, mappedValue.components(separatedBy: ";")).forEach {
+                $0.layer = Int($1) ?? nil
+            }
+        }
```
1ec5, line 99: This differs a bit from how other per-waypoint properties like `separatesLegs` (aka `waypoints` in the API request) are implemented:

https://github.com/mapbox/mapbox-directions-swift/blob/53f2d24edb78db8ea8cb12147be521b05d976a16/Sources/MapboxDirections/DirectionsOptions.swift#L172-L179

What you’ve got makes a bit more sense, though, so we should double-check that the `separatesLegs` code is correct.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -429,6 +435,11 @@ open class RouteOptions: DirectionsOptions {
             params.append(URLQueryItem(name: CodingKeys.waypointTargets.stringValue, value: targetCoordinates))
         }
         
+        if waypoints.first(where: { $0.layer != nil} ) != nil {
```
1ec5, line 438: Nit: `contains(where:_:)` is adequate since we aren’t using the return value for anything.

Sources/MapboxDirections/Waypoint.swift
```diff
@@ -176,6 +180,15 @@ public class Waypoint: Codable, ForeignMemberContainerClass {
      */
     public var snappedDistance: LocationDistance?
     
+    /**
+     The layer of road that the waypoint is positioned which is used to filter the road segment that the waypoint will be placed on in Z-order. It is useful for avoiding ambiguity in the case of multi-level roads (such as a tunnel under a road).
+     
+     This property corresponds to the ['layers'](https://docs.mapbox.com/api/navigation/directions/#optional-parameters) query parameter in the Mapbox Directions API. If a matching layer is not found, the Mapbox Directions API will choose a suitable layer according to the other procided parameters.
```
1ec5, line 186: Although there’s no guarantee that the Directions API returns data from OpenStreetMap in every situation, the [OSM Wiki’s `layer` key documentation](https://wiki.openstreetmap.org/wiki/Key:layer) is probably the best explanation available for what goes into this property, so we should mention it somehow in the documentation comment. Even if the developer using the navigation SDK should just plumb the value from a PassiveLocationManager into a NavigationRouteOptions, it helps to know what these values correspond to in terms of real-world phenomena. There are some nuances that are difficult to discover otherwise, such as building passages being 0 rather than −1.

1ec5 (approved):
Minor comments only; feel free to merge once you get a chance to address them.

Issue #718, opened 2022-07-19
Replace AttributeOptions with more extensible type


1ec5:
AttributeOptions has historically been an OptionSet for convenience and to avoid redundancy: https://github.com/mapbox/mapbox-directions-swift/pull/118#discussion_r109551839. However, after we made it an OptionSet, it became apparent that the Directions API could arbitrarily begin accepting or returning new unrecognized values.

OptionSet is essentially syntactic sugar and a more compact representation of Set composed with an enumeration. We should replace it with an explicit Set and an “extensible enumeration”. This could take the form of either a struct with constants or an enumeration with an `unrecognized(_:)` case (taking the raw string as an associated value).

This would be a backwards-incompatible change.

1ec5:
As a workaround until the next major version, #748 adds an extensibility hook to AttributeOptions.

Issue #662, opened 2022-03-11
Add option to avoid specific geometries


1ec5:
The Directions API recently overloaded the `exclude` query parameter with an option to specify a list of point geometries in [WKT format](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry). We should add support for this syntax to serve certain use cases that require arbitrary avoidance. This would also help us achieve platform parity with the Java SDK, which implemented the feature in mapbox/mapbox-java#1362. However, it should not be implemented until the new `exclude` syntax leaves beta and becomes a formal part of the API contract.

## Design

The overloading of `exclude` is particularly problematic for this library, which has always exposed `exclude` as an option set. But now it can hold arbitrary geometry data, which needs to be typed as a Geometry. Unfortunately, an OptionSet struct can’t store associated values, so there’s no way for us to support this syntax without breaking backwards compatibility. Besides, a geometry is semantically different than a geometry.

https://github.com/mapbox/mapbox-directions-swift/blob/2b14b8c276523a01dc730897593990ef83c71320/Sources/MapboxDirections/RouteOptions.swift#L57 https://github.com/mapbox/mapbox-directions-swift/blob/2b14b8c276523a01dc730897593990ef83c71320/Sources/MapboxDirections/RouteOptions.swift#L152-L157 https://github.com/mapbox/mapbox-directions-swift/blob/2b14b8c276523a01dc730897593990ef83c71320/Sources/MapboxDirections/RoadClasses.swift#L3-L6

Perhaps this conflation in the API can be revisited in light of the semantic awkwardness that becomes apparent in Swift. Regardless, we can implement a parallel property, `shapeToAvoid`, declared as type `Geometry?`, with the expectation of being set to a `MultiPoint` until more geometry types are implemented on the server side. RouteOptions’ Codable implementation can switch between `roadClassesToAvoid` and `shapeToAvoid` depending on the value. This feature depends on Turf adding support for converting between WKT and GeoJSON: mapbox/turf-swift#185.

## Workaround

Until we’re able to formally add support for this syntax, a developer can [hook into the beta parameter themselves](https://docs.mapbox.com/ios/navigation/examples/beta-query-parameters/):

```swift
/**
  Route options for avoiding known [pedestrian scrambles](https://en.wikipedia.org/wiki/Pedestrian_scramble).
 */
class UnscrambledRouteOptions: RouteOptions {
    /// The locations of some known pedestrian scrambles to avoid.
[...]

PR #748, opened 2022-09-22
Custom AttributeOptions

Udumft:
Added `AttributeOptions.customOptions` and manipulation methods to allow custom options to be processed.

This is an approach to allow customization while avoiding breaking changes. The safer option would be to refactor AttributeOptions into a struct.


1ec5:
> This is an approach to allow customization while avoiding breaking changes. The safer option would be to refactor AttributeOptions into a struct.

For reference, this is tracked in #718 as a future backwards-incompatible change.

Sources/MapboxDirections/AttributeOptions.swift
```diff
@@ -122,3 +135,101 @@ extension AttributeOptions: Codable {
         self = AttributeOptions(descriptions: descriptions)!
     }
 }
+
+extension AttributeOptions {
```
1ec5, line 139: This is a pretty extensive extension that makes #718 all the more desirable. That said, the `SetAlgebra` protocol that `OptionSet` conforms to requires [several other methods](https://developer.apple.com/documentation/swift/setalgebra#Conforming-to-the-SetAlgebra-Protocol) that `OptionSet` is currently implementing, now incompletely. The fundamental methods seem to be `==`, `intersection(_:)`, and `union(_:)`, with all the others being derived from them. This is a different set of methods than what this extension implements, so I’m not sure if there would be inconsistencies as a result.

Udumft: Right. What I wanted to discuss before fully committing to the impl. is if this approach is really worth the result. I am inclined to call it over-engineered and potentially confusing to use, but it gets the job done in a nonbreaking way.
So do we need to allow options customizing even in suboptimal way now or we are OK to wait till it is time for API breakage update?

1ec5: I guess my question was whether a different subset of methods (`==`, `intersection(_:)`, and `union(_:)`) could be overridden to ensure that all the `SetAlgebra` methods works, since we can’t restrict which of those methods a developer will use in their code. If not, I agree that having to reimplement all the `SetAlgebra` methods just for `AttributeOptions` would be overkill. But our current use of `OptionSet` for `RoadClasses` similarly blocks further support for the Directions API, such as #662, so if the solution is general enough to cover both types, I think the additional code would be worth the trouble.

Sources/MapboxDirections/CustomValueOptionSet.swift
```diff
@@ -0,0 +1,629 @@
...
+/// Describes how `customOptions` component is compared during logical operations in `CustomValueOptionSet`.
+public enum CustomOptionComparisonPolicy {
+    /// Custom options are equal if `customOptions` key-value pairs are strictly equal
+    ///
+    /// Example:
+    /// [1: "value1"] == [1: "value1"]
+    /// [1: "value1"] != [1: "value2"]
+    /// [1: "value1"] != [:]
+    /// [:] == [:]
+    case allowEqual
```
Udumft, line 12: Not sure about the naming at this point...

Sources/MapboxDirections/CustomValueOptionSet.swift
```diff
@@ -0,0 +1,629 @@
...
+    /// [:] == [:]
+    case allowEqualOrNull
+    /// Custom options are not compared. Only `rawValue` is taken into account when comparing `CustomStringOptionSet`s.
+    ///
+    /// Example:
+    /// [1: "value1"] == [1: "value1"]
+    /// [1: "value1"] == [1: "value2"]
+    /// [1: "value1"] == [:]
+    /// [:] == [:]
+    case allowUnequal
```
1ec5, line 28: `rawValueEqual`?

Sources/MapboxDirections/CustomValueOptionSet.swift
```diff
@@ -0,0 +1,629 @@
...
+}
+
+/// Option set implementation which allows each option to have custom string value attached.
+public protocol CustomValueOptionSet: OptionSet where RawValue: FixedWidthInteger, Element == Self {
+    associatedtype Element = Self
+    associatedtype CustomValue: Equatable
+    var rawValue: Self.RawValue { get set }
+    var customOptions: [RawValue: CustomValue] { get set }
```
1ec5, line 36: `customOptionsByRawValue`

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Other Changes
 
 * Added the `Waypoint.layer` property, which can ensure that the route begins on the correct road if it is above or below another road. ([#745](https://github.com/mapbox/mapbox-directions-swift/pull/745))
+* Expanded `AttributeOptions` to allow user options with custom values. See `CustomValueOptionSet` for reference. ([#748](https://github.com/mapbox/mapbox-directions-swift/pull/748))
```
1ec5, line 12: `CustomValueOptionSet` is probably more interesting to us internally; from a developer standpoint, the one to pay attention to is `AttributeOptions.customOptionsByRawValue`.

1ec5 (approved):
Some renames, but the tests look convincing. The design of mapping strings to integers is unfortunate, but I don’t know how better to accommodate the expectations built into `OptionSet`.

PR #749, opened 2022-09-28
Exposing toll fees

Udumft:
The PR adds `Route.tollCosts` and `RouteOptions.computeTollCost` for requesting and working with toll fees. Marking new properties and types with `:nodoc:` and skipping CHANGELOG update since it is a private API feature.

### DIFF
--- a/Sources/MapboxDirections/Route.swift
+++ b/Sources/MapboxDirections/Route.swift
@@ -7,6 +7,9 @@ import Turf
  Typically, you do not create instances of this class directly. Instead, you receive route objects when you request directions using the `Directions.calculate(_:completionHandler:)` or `Directions.calculateRoutes(matching:completionHandler:)` method. However, if you use the `Directions.url(forCalculating:)` method instead, you can use `JSONDecoder` to convert the HTTP response into a `RouteResponse` or `MapMatchingResponse` object and access the `RouteResponse.routes` or `MapMatchingResponse.routes` property.
  */
 open class Route: DirectionsResult {
+    private enum CodingKeys: String, CodingKey, CaseIterable {
+        case tollCosts = "toll_costs"
+    }
     
     /**
      Initializes a route.
@@ -28,8 +31,27 @@ open class Route: DirectionsResult {
      - parameter decoder: The decoder of JSON-formatted API response data or a previously encoded `Route` object.
      */
     public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        tollCosts = try container.decodeIfPresent([TollCost].self, forKey: .tollCosts)
+        
         try super.init(from: decoder)
+        try decodeForeignMembers(notKeyedBy: CodingKeys.self, with: decoder)
     }
+    
+    public override func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encodeIfPresent(tollCosts, forKey: .tollCosts)
+        
+        try super.encode(to: encoder)
+    }
+    
+    /**
+     :nodoc:
+     List of calculated toll costs for this route.
+     
+     See `TollCost`.
+     */
+    open var tollCosts: [TollCost]?
 }
 
 extension Route: Equatable {
--- a/Sources/MapboxDirections/RouteOptions.swift
+++ b/Sources/MapboxDirections/RouteOptions.swift
@@ -107,6 +107,9 @@ open class RouteOptions: DirectionsOptions {
            let arriveBy = formatter.date(from: mappedValue) {
             self.arriveBy = arriveBy
         }
+        if mappedQueryItems[CodingKeys.computeTollCost.stringValue] == "true" {
+            self.computeTollCost = true
+        }
     }
 
     #if canImport(CoreLocation)
@@ -155,6 +158,7 @@ open class RouteOptions: DirectionsOptions {
         case arriveBy = "arrive_by"
         case departAt = "depart_at"
         case layers = "layers"
+        case computeTollCost = "compute_toll_cost"
     }
     
     public override func encode(to encoder: Encoder) throws {
@@ -180,6 +184,10 @@ open class RouteOptions: DirectionsOptions {
         }
         if let departAt = departAt {
             try container.encode(formatter.string(from: departAt), forKey: .departAt)
+        }
+        
+        if computeTollCost {
+            try container.encode(computeTollCost, forKey: .computeTollCost)
         }
     }
     
@@ -224,6 +232,8 @@ open class RouteOptions: DirectionsOptions {
         if let dateString = try container.decodeIfPresent(String.self, forKey: .arriveBy) {
             arriveBy = formatter.date(from: dateString)
         }
+        
+        computeTollCost = try container.decodeIfPresent(Bool.self, forKey: .computeTollCost) ?? false
         
         try super.init(from: decoder)
     }
@@ -393,6 +403,14 @@ open class RouteOptions: DirectionsOptions {
     }
     private var _initialManeuverAvoidanceRadius: LocationDistance?
     
+    /**
+     :nodoc:
+     Toggle whether to return calculated toll cost for the route, if data is available.
+     
+     Default value is `false`.
+     */
+    open var computeTollCost = false
+    
     // MARK: Getting the Request URL
     
     override open var urlQueryItems: [URLQueryItem] {
@@ -472,6 +490,11 @@ open class RouteOptions: DirectionsOptions {
                 params.append(URLQueryItem(name: CodingKeys.arriveBy.stringValue,
                                            value: String(formatter.string(from: arriveBy))))
             }
+        }
+        
+        if computeTollCost {
+            params.append(URLQueryItem(name: CodingKeys.computeTollCost.stringValue,
+                                       value: String(computeTollCost)))
         }
         
         return params + super.urlQueryItems
--- a/Sources/MapboxDirections/TollCost.swift
+++ b/Sources/MapboxDirections/TollCost.swift
@@ -1,0 +1,190 @@
+import Foundation
+import Turf
+
+/**
+ :nodoc:
+ Toll cost information for the `Route`.
+ */
+public struct TollCost: Equatable,  Codable, ForeignMemberContainer {
+    public var foreignMembers: Turf.JSONObject = [:]
+    
+    private enum CodingKeys: String, CodingKey {
+        case currency
+        case paymentMethods = "payment_methods"
+    }
+
+    /**
+     Related currency string.
+     
+     Uses ISO 4217 format. Refers to values in `CostPerVehicleSize`.
+     */
+    public let currency: String
+    /**
+     Information about toll payment methods.
+     */
+    public let paymentMethods: PaymentMethods
+    
+    init(currency: String, paymentMethods: PaymentMethods) {
+        self.currency = currency
+        self.paymentMethods = paymentMethods
+    }
+        
+    /**
+     :nodoc:
+     Information about toll payment methods.
+     */
+    public struct PaymentMethods: Equatable, Codable, ForeignMemberContainer {
+        public var foreignMembers: Turf.JSONObject = [:]
+        
+        private enum CodingKeys: String, CodingKey {
+            case ETC = "etc"
+            case cash
+        }
+        
+        /**
+         Information about payment by ETC.
+         */
+        public let ETC: CostPerVehicleSize?
+        /**
+         Information about payment by cash.
+         */
+        public let cash: CostPerVehicleSize?
+        
+        init(ETC: CostPerVehicleSize? = nil, cash: CostPerVehicleSize? = nil) {
+            self.ETC = ETC
+            self.cash = cash
+        }
+    }
+    
+    /**
+     :nodoc:
+     `PaymentMethods` details about particular vehicle size.
+     
+     Vehicle sizes are [standartized](https://en.wikipedia.org/wiki/Expressways_of_Japan#Tolls).
+     */
+    public struct CostPerVehicleSize: Equatable, Codable, ForeignMemberContainer {
+        public var foreignMembers: Turf.JSONObject = [:]
+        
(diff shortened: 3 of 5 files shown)


Sources/MapboxDirections/TollCost.swift
```diff
@@ -0,0 +1,190 @@
...
+            case large
+            case jumbo
+        }
+        
+        /**
+         The toll cost for a small sized vehicle.
+         
+         A toll cost of `0` is valid and simply means that no toll costs are incurred for this route.
+         */
+        public let small: Double?
```
1ec5, line 81: Frameworks such as [Intents](https://developer.apple.com/documentation/sirikit/inridefarelineitem/1649903-price) and [StoreKit](https://developer.apple.com/documentation/storekit/product/3749586-price) type `price` as a `Decimal`, probably because if you do enough arithmetic on prices represented as floating-point numbers, [rounding error starts to add up](https://stackoverflow.com/a/3730040/4585461). It’s unlikely to be a problem in practice when calculating the price of tolls along a single route, but we should be consistent with software that does have to worry about this issue. Let’s try using `Decimal` in our own representation and convert back to `Double` when encoding as JSON.

Sources/MapboxDirections/Route.swift
```diff
@@ -28,8 +31,27 @@ open class Route: DirectionsResult {
...
+        try super.encode(to: encoder)
     }
+    
+    /**
+     :nodoc:
+     List of calculated toll costs for this route.
+     
+     See `TollPrice`.
+     */
+    open var tollsPrices: [TollPrice]?
```
1ec5, line 54: Typo: `tollPrices`.

Sources/MapboxDirections/Route.swift
```diff
@@ -28,8 +31,27 @@ open class Route: DirectionsResult {
...
+        try container.encodeIfPresent(tollsPrices, forKey: .tollCosts)
+        
+        try super.encode(to: encoder)
     }
+    
+    /**
+     :nodoc:
+     List of calculated toll costs for this route.
+     
+     See `TollPrice`.
```
1ec5, line 52: I realize this is currently `:nodoc:`, but in case we make the documentation public later, let’s mention that this property is set to `nil` unless `RouteOptions.includesTollPrices` is set to `true`.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -393,6 +403,14 @@ open class RouteOptions: DirectionsOptions {
     }
     private var _initialManeuverAvoidanceRadius: LocationDistance?
     
+    /**
+     :nodoc:
+     Toggle whether to return calculated toll cost for the route, if data is available.
+     
+     Default value is `false`.
```
1ec5, line 410: Mention that, if this property is set to `true`, each resulting route’s `Route.tollPrices` property is populated with toll prices.

Sources/MapboxDirections/TollPrice.swift
```diff
@@ -0,0 +1,136 @@
...
+     Payment is done by cash.
+     */
+    public static let cash = PaymentMethod(identifier: "cash")
+}
+
+/**
+ :nodoc:
+ Categories by which toll fees are divided.
+ */
+public struct Category: Hashable, Equatable {
```
1ec5, line 28: This is a very generic name prone to collisions with unrelated types in other libraries. Let’s call it `TollCategory`.

Sources/MapboxDirections/TollPrice.swift
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ :nodoc:
+ Information about toll payment method.
+ */
+public struct PaymentMethod: Hashable, Equatable {
+    /**
+     Method identifier.
+     */
+    public let identifier: String
```
1ec5, line 12: Interesting choice to make the identifier a member of this struct. If we were following the usual extensible enumeration pattern, this struct would only contain `static let`s of type `String`. The more structured approach you’ve taken is more flexible, but JSON doesn’t allow an object’s keys to be objects, so this flexibility would be unused.

Sources/MapboxDirections/TollPrice.swift
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     A large sized vehicle.
+     
+     Vehicle size is [standartized](https://en.wikipedia.org/wiki/Expressways_of_Japan#Tolls).
+     */
+    public static let large = Category(name: "large")
+    /**
+     A jumbo sized vehicle.
+     
+     Vehicle size is [standartized](https://en.wikipedia.org/wiki/Expressways_of_Japan#Tolls).
```
1ec5, line 61: Typo, and also we should probably mention for now that these categories are for Japan, since “jumbo” doesn’t exist as a category anywhere else.

> In Japan, this is a [standard vehicle size](https://en.wikipedia.org/wiki/Expressways_of_Japan#Tolls).

Sources/MapboxDirections/TollPrice.swift
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public static let jumbo = Category(name: "jumbo")
+}
+
+public typealias CategoriesTolls = [Category: Decimal]
+public typealias PaymentMethods = [PaymentMethod: CategoriesTolls]
```
1ec5, line 67: Do we need these type aliases? The interface might be clearer if we simply inline these types.

Sources/MapboxDirections/TollPrice.swift
```diff
@@ -0,0 +1,136 @@
...
+    
+    private enum CodingKeys: String, CodingKey {
+        case currency
+        case paymentMethods = "payment_methods"
+    }
+
+    /**
+     Related currency code string.
+     
+     Uses ISO 4217 format. Refers to values in `CategoriesTolls`. A toll cost of `0` is valid and simply means that no toll costs are incurred for this route.
```
1ec5, line 84: As a developer, I’d expect to be able to plug this `currencyCode` directly into `NumberFormatter.currencyCode`, but this special case would be surprising and easy to miss. If a road is free, either there should be no toll price (`nil` entry in the array), or if we need to distinguish this from an unknown value, then the currency code should be `nil` and any payment amount should be 0. We can always convert the `nil` to a `0` string when converting back to JSON.

(This would be a little less awkward if the tolls were structured as a dictionary containing structs containing currency codes and amounts, rather than a struct containing currency codes and dictionaries of amounts.)

Sources/MapboxDirections/TollPrice.swift
Udumft: I am not sure that this is easier to read:
```
public let paymentMethods: [[TollPaymentMethod: [TollCategory: Decimal]]]
```

Udumft: I'm not fond of embedding that much detail into a type or variable name, IMHO that looks synthetic. Probably it would be better to generalize the contents and name this typealias something like `PaymentDetails`? Thus it suggests that there are various parameters inside, which may or may not be grouped by category, payment method, etc.

Sources/MapboxDirections/TollPrice.swift
Udumft: Documentation is sparse on that point, but it never said that currency code may be empty. I expect that it will still be local code even if the route is free.

Sources/MapboxDirections/TollPrice.swift
1ec5: It is an awkward name, but probably also a sign that the raw structure exposed by the JSON response is not an ideal structure to expose to developers in Swift.

I hesitate to rabbit-hole on what’s still an undocumented part of the API, but maybe an alternative would be a flatter array of TollPrice structures, where TollPrice itself can contain a vehicle type condition and a payment method condition, both as properties. When encoding, we can assemble the elements of this array into a dictionary, throwing if there are duplicates. It reminds me of how a `URLComponents` has an array of `URLQueryItem`s even though typically one thinks of it as a key-value store.

I realize this goes against some of what I wrote in https://github.com/mapbox/mapbox-directions-swift/pull/749#discussion_r982701348, when I thought we could get away with a rawer structure. The downside is that we would shift some consistency checks to runtime, but the upside is that if the API introduces more axes in the future, we wouldn’t have to dramatically restructure the property and all its call sites.

If this doesn’t sound like a sound strategy, then at least we should rename the property to describe the whole type, not just the keys: `pricesByPaymentMethod`, `paymentsByMethod`, or… `userInfo`. 😉

Udumft: Could you please expand what you mean by the need to restructure properties if there are new entities added? As I see, current structure is extendable with new types of payment methods and categories as well as adding new properties on those levels.
Flattening the structure, on the other hand, creates troubles coding foreign keys which almost the same as addition of new properties to JSON. I agree that flat structure might be more readable for user, and `paymentMethods` should be renamed for the same reason, but I think that altering the hierarchy here may damage extensibility.
BTW I still think `PaymentMethods -> PaymentDetails` is a viable option :D

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -54,6 +54,10 @@
 		2B67F68227EDC2EF007BD6D2 /* RouteResponseWithForeignMembers.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B67F68127EDC2EF007BD6D2 /* RouteResponseWithForeignMembers.json */; };
 		2B67F68327EDC2EF007BD6D2 /* RouteResponseWithForeignMembers.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B67F68127EDC2EF007BD6D2 /* RouteResponseWithForeignMembers.json */; };
 		2B67F68427EDC2EF007BD6D2 /* RouteResponseWithForeignMembers.json in Resources */ = {isa = PBXBuildFile; fileRef = 2B67F68127EDC2EF007BD6D2 /* RouteResponseWithForeignMembers.json */; };
+		2B79D3E628E5EB3900B3F127 /* TollPrice.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B79D3E528E5EB3900B3F127 /* TollPrice.swift */; };
+		2B79D3E728E5EB3900B3F127 /* TollPrice.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B79D3E528E5EB3900B3F127 /* TollPrice.swift */; };
+		2B79D3E828E5EB3900B3F127 /* TollPrice.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B79D3E528E5EB3900B3F127 /* TollPrice.swift */; };
+		2B79D3E928E5EB3900B3F127 /* TollPrice.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B79D3E528E5EB3900B3F127 /* TollPrice.swift */; };
```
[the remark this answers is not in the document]
Udumft: 4 targets == 4 includes. That's expected.

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -474,6 +493,11 @@ open class RouteOptions: DirectionsOptions {
             }
         }
         
+        if includesTollPrices {
+            params.append(URLQueryItem(name: CodingKeys.includesTollPrices.stringValue,
+                                       value: String(includesTollPrices)))
```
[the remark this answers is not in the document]
Udumft: We've been relying on this conversion since the very beginning

Udumft:
So I've updated `TollPrice` to the flat representation. It should be easier to read and understand for developers and it still has some possibilities for `foreignMembers` coding. It is capable of detecting new types of payment methods and new categories. New parameters are handled on the toll price object itself, but it won't handle new properties anywhere else.

Sources/MapboxDirections/RouteOptions.swift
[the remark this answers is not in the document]
Udumft: Updated

Sources/MapboxDirections/RouteOptions.swift
```diff
@@ -474,11 +493,22 @@ open class RouteOptions: DirectionsOptions {
...
+            params.append(URLQueryItem(name: CodingKeys.includesTollPrices.stringValue,
+                                       value: includesTollPrices.queryString))
+        }
+        
         return params + super.urlQueryItems
     }
     
 }
 
+extension Bool {
```
kried, line 506: Should it be marked as private? Or should it be moved to Extensions directory instead?

1ec5: This could be moved to a separate file for consistency with Array.swift and String.swift. `queryString` should be internal by default.

1ec5:
> It is capable of detecting new types of payment methods and new categories. New parameters are handled on the toll price object itself, but it won't handle new properties anywhere else.

I think this is consistent with the JSON format. In case the API introduces new payment methods or vehicle categories that we don’t recognize, `TollPaymentMethod` and `TollCategory` are both extensible enumerations; the unrecognized values would just be strings without symbols.

Sources/MapboxDirections/RouteOptions.swift
Udumft: Didn't want to draw too much attention to this small extension, just like `DateFormatter` or `LocationDistance` extensions we have nearby, since it is used just in this one file. Should've probably `fileprivate` it...

Sources/MapboxDirections/Route.swift
```diff
@@ -7,6 +7,9 @@ import Turf
  Typically, you do not create instances of this class directly. Instead, you receive route objects when you request directions using the `Directions.calculate(_:completionHandler:)` or `Directions.calculateRoutes(matching:completionHandler:)` method. However, if you use the `Directions.url(forCalculating:)` method instead, you can use `JSONDecoder` to convert the HTTP response into a `RouteResponse` or `MapMatchingResponse` object and access the `RouteResponse.routes` or `MapMatchingResponse.routes` property.
  */
 open class Route: DirectionsResult {
+    private enum CodingKeys: String, CodingKey, CaseIterable {
+        case tollPrices = "toll_costs"
```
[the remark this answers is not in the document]
Udumft: It does not. Attempting to request a MM with toll prices results in unrecognized parameter error.

PR #755, opened 2022-10-21
RouteLeg.closures integration

Udumft:
This PR adds support for RouteLeg Closures. Added corresponding type, properties in RouteLeg and DirectionsOptions. Also updated refreshing methods for the new updated info. Refreshing also refined a bit by introducing single refresh method with flexible configuration which should cover possible variations and reduce amount of user code for refreshing routes.

### DIFF
--- a/Sources/MapboxDirections/RouteRefreshResponse.swift
+++ b/Sources/MapboxDirections/RouteRefreshResponse.swift
@@ -31,7 +31,7 @@ public struct RouteRefreshResponse: ForeignMemberContainer {
     /**
      A skeleton route that contains only the time-sensitive information that has been updated.
      
-     Use the `Route.refreshLegAttributes(from:)`, `Route.refreshLegAttributes(from:legIndex:legShapeIndex:)` and `Route.refreshLegIncidents(from:)`, `Route.refreshLegIncidents(from:legIndex:legShapeIndex:)` methods to merge this object with the original route to continue using the original route with updated information.
+     Use the `Route.refreshLegAttributes(from:)`, `Route.refreshLegAttributes(from:legIndex:legShapeIndex:)`, `Route.refreshLegIncidents(from:)`, `Route.refreshLegIncidents(from:legIndex:legShapeIndex:)`, `Route.refreshLegClosures(from:legIndex:legShapeIndex:)` or `Route.refresh(from:refreshParameters:)` methods to merge this object with the original route to continue using the original route with updated information.
      */
     public var route: RefreshedRoute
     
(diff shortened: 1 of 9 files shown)


swift-package-baseline/breakage-allowlist-path.txt
```diff
@@ -1 +1,2 @@
-
+API breakage: enumelement RouteLeg.CodingKeys.closures has been added as a new enum case
```
Udumft, line 1: Adding new annotation type is impossible without adding new enum element...

swift-package-baseline/breakage-allowlist-path.txt
```diff
@@ -1 +1,2 @@
...
+API breakage: var RouteLegRefreshSource.refreshedClosures has been added as a protocol requirement
```
Udumft, line 2: Mitigated by default extension implementation.

Sources/MapboxDirections/RouteRefreshResponse.swift
```diff
@@ -157,4 +247,25 @@ extension Route {
...
     }
+    
+    /**
+     Merges the closures of the given route’s legs into the receiver’s legs.
+
+     - parameter refreshedRoute: The route containing leg closures to merge into the receiver. If this route contains fewer legs than the receiver, this method skips legs from the beginning of the route to make up the difference, so that merging the closures from a one-leg route affects only the last leg of the receiver.
+     - parameter legIndex: The index of a leg, from which to start applying the refreshed closures.
+     - parameter legShapeIndex: Index of a geometry of the `legIndex` leg, where to start refreshing from.
+     */
+    public func refreshLegClosures(from refreshedRoute: RouteRefreshSource, legIndex: Int = 0, legShapeIndex: Int = 0) {
```
kried, line 258: `refreshLegClosures` method looks just like `refreshLegIncidents`, but contains a specific implementation with 1 parameter.  

Maybe it's better to use the same approach for all three methods here, either default parameters values or separate method. Using different solutions for the same situation might be confusing

Tests/MapboxDirectionsTests/RouteRefreshTests.swift
```diff
@@ -113,11 +113,11 @@ class RouteRefreshTests: XCTestCase {
                 }
                 
                 let route = routeResponse.routes?[routeIndex]
-                route?.refreshLegAttributes(from: refresh.route)
-                route?.refreshLegIncidents(from: refresh.route)
+                route?.refresh(from: refresh.route)
```
kried, line 116: Should we also add separate tests for `refreshLegClosures` only if there is separate public method `Route.refreshLegClosures(from:legIndex:legShapeIndex:)`?

kried (approved):

Tests/MapboxDirectionsTests/RouteRefreshTests.swift
Udumft: I don't think it will bring some value, since `refresh(from:)` is basically a convenience wrapper method

PR #766, opened 2022-11-18
Fix issues related to resolving dependencies for Mapbox Directions `v2.7.0`.

MaximAlien:
Fix issues related to resolving dependencies for Mapbox Directions `v2.7.0`.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
-github "raphaelmor/Polyline" == 5.0.3
-github "mapbox/turf-swift" ~> 2.4
+github "raphaelmor/Polyline" ~> 5.0.3
+github "mapbox/turf-swift" ~> 2.5.0
--- a/Cartfile
+++ b/Cartfile
@@ -1,2 +1,2 @@
-github "raphaelmor/Polyline" ~> 5.0.3
-github "mapbox/turf-swift" ~> 2.5.0
+github "raphaelmor/Polyline" == 5.0.3
+github "mapbox/turf-swift" == 2.5.0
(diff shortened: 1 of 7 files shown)


Cartfile
```diff
@@ -1,2 +1,2 @@
 github "raphaelmor/Polyline" == 5.0.3
-github "mapbox/turf-swift" ~> 2.4
```
MaximAlien, line 2: Unfortunately Carthage resolves both `~> 2.4` and `~> 2.4.0` to the latest `v2.6.1` version of Turf. Since Mapbox Directions brings iOS 11 removal only in `v2.8.0` we have to pin to the latest version of Turf that precedes `v2.6.1`.

MapboxDirections.xcodeproj/project.pbxproj
```diff
@@ -542,6 +542,10 @@
 		43F89F922350F952007B591E /* MapMatchingResponse.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MapMatchingResponse.swift; sourceTree = "<group>"; };
 		8A3B4C9A24EB55F60085DA64 /* RouteResponseTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = RouteResponseTests.swift; sourceTree = "<group>"; };
 		8A41B0FC24F5C2390021FFDC /* CHANGELOG.md */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = net.daringfireball.markdown; path = CHANGELOG.md; sourceTree = "<group>"; };
+		8A586D1B2927061B00679A54 /* Cartfile */ = {isa = PBXFileReference; lastKnownFileType = text; path = Cartfile; sourceTree = "<group>"; };
```
MaximAlien, line 545: Added for convenience.

Cartfile
1ec5: Correct, Carthage [intentionally does not support](https://github.com/Carthage/Carthage/blob/a91d086ceaffef65c4a4a761108f3f32c519940c/Documentation/Artifacts.md#version-requirement) an “up to next minor version” operator, because it considers that semantic to violate the Semantic Versioning standard. It only makes an exception for pre-v1.0 versions.

1ec5 (approved):
Suggested changelog entry:

* This release of MapboxDirections requires Turf v2.5._x_ to work around a dependency resolution failure. If you use Carthage, it requires Turf v2.5.0 exactly. ([#766](https://github.com/mapbox/mapbox-directions-swift/pull/766))

Package.swift
```diff
@@ -20,8 +20,8 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        .package(url: "https://github.com/raphaelmor/Polyline.git", from: "5.0.2"),
-        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.4.0"),
+        .package(url: "https://github.com/raphaelmor/Polyline.git", .upToNextMinor(from: "5.0.3")),
```
MaximAlien, line 23: `upToNextMinor` was deprecated in Swift 5.6. We can still use it here for now.

PR #778, opened 2022-12-08
Add `SubType` to `VisualInstruction.Component`.

MaximAlien:
Add `SubType` to `VisualInstruction.Component`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,5 @@
 # Changes to Mapbox Directions for Swift
 
-## v2.9.0
-
-* Added the `RestStop.amenities` property that describes useful and important facilities such as gas stations, restaurants, and ATMs. ([#780](https://github.com/mapbox/mapbox-directions-swift/pull/780))
-* Added the `VisualInstruction.Component.SubType` that provides more context about the component guidance view. ([#778](https://github.com/mapbox/mapbox-directions-swift/pull/778))
-
 ## v2.8.0
 
 ### Packaging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v2.9.0
 
 * Added the `RestStop.amenities` property that describes useful and important facilities such as gas stations, restaurants, and ATMs. ([#780](https://github.com/mapbox/mapbox-directions-swift/pull/780))
+* Added the `VisualInstruction.Component.SubType` that provides more context about the component guidance view. ([#778](https://github.com/mapbox/mapbox-directions-swift/pull/778))
 
 ## v2.8.0
 
--- a/Sources/MapboxDirections/VisualInstructionComponent.swift
+++ b/Sources/MapboxDirections/VisualInstructionComponent.swift
@@ -276,16 +276,71 @@ extension VisualInstruction.Component: Codable {
         case lane
     }
     
+    /**
+     Subtype that provides more context about the component guidance view that may help in visual
+     markup and display choices.
+     */
     public enum SubType: String, Codable {
+        
+        /**
+         Junction view. Bird’s-eye artist’s rendition view of the overhead signage and
+         preferred road lane arrow on motorways where the road bifurcates into 2 or
+         more motorway trunk roads.
+         */
         case jct = "jct"
+        
+        /**
+         Advanced 2D signboard (vendor enhanced detailed signboard).
+         */
         case signboard = "signboard"
+        
+        /**
+         Service area/parking area guide map. Vertical artist’s rendition guide map of an SAPA rest area
+         showing various facilities icons such as restaurants, restrooms and parking areas.
+         The scale is approximately 1 to 5K.
+         */
         case sapaGuideMap = "sapaguidemap"
+        
+        /**
+         Service area/parking area. Bird’s-eye artist’s rendition view of the overhead
+         signage and preferred road lane arrow at a rest area ramp where the route
+         leaves the main road.
+         */
         case sapa = "sapa"
+        
+        /**
+         Sign image after a toll gate. Used immediately after exiting a toll gate containing just the
+         overhead signboard. The preferred road (not the lane) arrow is highlighted on the signboard.
+         */
         case afterToll = "aftertoll"
+        
+        /**
+         3D city real. Bird’s-eye artist’s rendition view of a general road intersection
+         and preferred road lane arrow. There is no overhead signage.
+         */
         case cityReal = "cityreal"
+        
+        /**
+         Expressway entrance. Bird’s-eye artist’s rendition view of the overhead signage and
+         preferred road lane arrow at an entrance ramp onto an expressway.
+         */
         case expresswayEntrance = "ent"
+        
+        /**
+         Expressway exit. Bird’s-eye artist’s rendition view of the overhead signage and
+         preferred road lane arrow at an exit ramp from an expressway.
+         */
         case expresswayExit = "exit"
+        
+        /**
+         Branched image after a toll gate. Bird’s-eye artist’s rendition view of the overhead
+         signage and preferred road lane arrow immediately after exiting a toll gate.
+         */
         case tollBranch = "tollbranch"
+        
+        /**
+         Direction signboard guidance view.
+         */
         case directionBoard = "directionboard"
     }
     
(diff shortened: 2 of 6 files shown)


Tests/MapboxDirectionsTests/VisualInstructionTests.swift
```diff
@@ -29,16 +29,17 @@ class VisualInstructionsTests: XCTestCase {
...
-                  "text": "CA01610_1_E",
-                  "type": "guidance-view",
-                  "imageURL": "https://www.mapbox.com/navigation"
+                "text": "CA01610_1_E",
+                "components": [
+                    [
+                        "text": "CA01610_1_E",
+                        "type": "guidance-view",
+                        "imageURL": "https://www.mapbox.com/navigation",
+                        "subType": "cityreal"
```
jill-cardamon, line 38: For tests, it would be great to check the other subtypes and their encodings to verify they all work as expected.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -254,6 +254,7 @@ public struct GuidanceViewImageRepresentation: Equatable {
 extension VisualInstruction.Component: Codable {
     private enum CodingKeys: String, CodingKey {
         case kind = "type"
+        case subType = "subType"
```
jill-cardamon, line 257: Is the string literal needed? Since the case name is the same as the literal.

MaximAlien: Still unclear if we should name it as `subType`, maybe `subKind` will align better? I'll remove custom coding key when we decide how to name it.

jill-cardamon: How I see it:
- `subType` works well because it is the same as what is seen in the response, and this naming is clear.
- `subKind` aligns with the other coding key, so it's good for consistency. However, this naming is a bit clunky (imo), and it might be confusing that this `subKind` doesn't work for all `kind`s.

Ideally, we would use `type` and `subType` here imo, but since that would be a breaking change, I would say `subKind` is better for consistency's sake. The documentation for `SubType` below is clear that this is solely for guidance views.

MaximAlien: Will leave `subType` as undocumented for now.

Tests/MapboxDirectionsTests/VisualInstructionTests.swift
MaximAlien: Tests for encoding and decoding were added.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
         case lane
     }
     
+    /**
+     :nodoc:
```
1ec5, line 280: If this is supposed to be undocumented, then it shouldn’t be in the changelog.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
...
+     Subtype that provides more context about the component guidance view that may help in visual
+     markup and display choices.
+     */
+    public enum SubType: String, Codable, CaseIterable {
```
1ec5, line 284: We’re avoiding the term “Type” in favor of “Kind”, as in `VisualInstruction.Component.Kind`, so “SubType” is a bit incongruous. If only a `guidanceView` can have a subtype, then call it `GuidanceViewKind`. Ideally, we’d make this an associated value of `Kind.guidanceView`, but that’s difficult because `Kind` is string-backed.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
...
+        
+        /**
+         Junction view. Bird’s-eye artist’s rendition view of the overhead signage and
+         preferred road lane arrow on motorways where the road bifurcates into 2 or
+         more motorway trunk roads.
+         */
+        case jct = "jct"
```
1ec5, line 291: Swift symbol names should [avoid abbreviation](https://www.swift.org/documentation/api-design-guidelines/#avoid-abbreviations), so this would be `junction`. However, from the documentation, it sounds like this is actually about a fork in the road, not any junction. In that case, we could call it `fork` for consistency with `ManeuverType.reachFork`.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
...
+         The scale is approximately 1 to 5K.
+         */
+        case sapaGuideMap = "sapaguidemap"
+        
+        /**
+         Service area/parking area. Bird’s-eye artist’s rendition view of the overhead
+         signage and preferred road lane arrow at a rest area ramp where the route
+         leaves the main road.
+         */
+        case sapa = "sapa"
```
1ec5, line 310: The “SAPA” abbreviation is specific to Japan. Consider renaming to `serviceArea` for discoverability and consistency with `RestStop.StopType.serviceArea`.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
...
+         Sign image after a toll gate. Used immediately after exiting a toll gate containing just the
+         overhead signboard. The preferred road (not the lane) arrow is highlighted on the signboard.
+         */
+        case afterToll = "aftertoll"
+        
+        /**
+         3D city real. Bird’s-eye artist’s rendition view of a general road intersection
+         and preferred road lane arrow. There is no overhead signage.
+         */
+        case cityReal = "cityreal"
```
1ec5, line 322: This name doesn’t make much sense – “real” isn’t a noun. A more intuitive name would be `realisticUrbanIntersection`.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
...
+        
+        /**
+         Expressway entrance. Bird’s-eye artist’s rendition view of the overhead signage and
+         preferred road lane arrow at an entrance ramp onto an expressway.
+         */
+        case expresswayEntrance = "entrance"
```
1ec5, line 328: “Expressway” is one translation of the Japanese term for what this library otherwise calls a `motorway`. Consider renaming to `motorway` for consistency with `RoadClasses.motorway` and `MapboxStreetsRoadClass.motorway`, and to avoid confusion with OSM’s [`expressway`](https://wiki.openstreetmap.org/wiki/Key:expressway) key, which is mutually exclusive of motorways.

1ec5 (commented):
Some naming-related feedback. (@abhishek1508, not sure if this is coming too late for the corresponding APIs on Android.)

Sources/MapboxDirections/VisualInstructionComponent.swift
MaximAlien: It's hard to find the best option here because this option is not yet public. At the same time it seems that `SubType` is only related to guidance views, so I'll rename it to `GuidanceViewKind`.

CHANGELOG.md
```diff
@@ -1,9 +1,5 @@
 # Changes to Mapbox Directions for Swift
 
-## v2.9.0
```
MaximAlien, line 3: This entry should not be public just yet.

Sources/MapboxDirections/VisualInstructionComponent.swift
```diff
@@ -275,6 +276,75 @@ extension VisualInstruction.Component: Codable {
...
+     markup and display choices.
+     */
+    public enum GuidanceViewKind: String, Codable, CaseIterable {
+        
+        /**
+         Junction view. Bird’s-eye artist’s rendition view of the overhead signage and
+         preferred road lane arrow on motorways where the road bifurcates into 2 or
+         more motorway trunk roads.
+         */
+        case fork = "jct"
```
jill-cardamon, line 291: nit: imo, we should make these doc comments more clear since the naming has changed. Some of @1ec5's suggestions provide some clarity as to what each of these `GuidanceViewKind`s mean, and it would be great to let developers know.

jill-cardamon (approved):

Sources/MapboxDirections/VisualInstructionComponent.swift
MaximAlien: It's a bit hard to provide more clarity at this point, maybe in future we'll be able to extend these docs. Current info regarding sub types was taken from private docs and unfortunately they do not have enough info either.

PR #818, opened 2024-02-29
Turf bump

Udumft:
Bumped Turf to v2.7

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         .package(url: "https://github.com/raphaelmor/Polyline.git", from: "5.0.2"),
-        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.7.0"),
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -21,7 +21,7 @@ let package = Package(
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         .package(url: "https://github.com/raphaelmor/Polyline.git", from: "5.0.2"),
-        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.6.1"),
+        .package(name: "Turf", url: "https://github.com/mapbox/turf-swift.git", from: "2.7.0"),
```
kried, line 24: in SPM it allows to use 2.8.0

From the SPM docs
```
    /// The following example allows the Swift Package Manager to select a version
    /// like a  `1.2.3`, `1.2.4`, or `1.3.0`, but not `2.0.0`.
    ///
    ///.package(url: "https://example.com/example-package.git", from: "1.2.3"),
```

kried (approved):

PR #819, opened 2024-02-29
Turf bump (#818)

Udumft:
#818 Cherry-pick


Cartfile
```diff
@@ -1,2 +1,2 @@
 github "raphaelmor/Polyline" ~> 5.0
-github "mapbox/turf-swift" ~> 2.6.1
+github "mapbox/turf-swift" ~> 2.7.0
```
azarovalex, line 2: Should be fine to bump to 2.8?

azarovalex (approved):

Cartfile
Udumft: There are Maps versions which use exactly 2.7.0, so we need a compatible version.

PR #827, opened 2024-06-11
Route refreshing TTL

Udumft:
Added route response refresh TTL field and a `DirectionsError` to signal refreshing has expired

### DIFF
--- a/Sources/MapboxDirections/DirectionsError.swift
+++ b/Sources/MapboxDirections/DirectionsError.swift
@@ -8,7 +8,11 @@ import FoundationNetworking
  */
 public enum DirectionsError: LocalizedError {
     
-    public init(code: String?, message: String?, response: URLResponse?, underlyingError error: Error?) {
+    public init(code: String?, message: String?, response: URLResponse?, underlyingError error: Error?, refreshTTL: Int? = nil) {
+        guard refreshTTL == nil || refreshTTL != 0 else {
+            self = .refreshExpired
+            return
+        }
         if let response = response as? HTTPURLResponse {
             switch (response.statusCode, code ?? "") {
             case (200, "NoRoute"):
@@ -106,6 +110,13 @@ public enum DirectionsError: LocalizedError {
     case rateLimited(rateLimitInterval: TimeInterval?, rateLimit: UInt?, resetTime: Date?)
     
     /**
+     Current route may no longer be refreshed.
+     
+     Route's refresh TTL has expired and it is not possible to refresh it anymore. Try requesting a new one instead.
+     */
+    case refreshExpired
+
+    /**
      Unknown error case. Look at associated values for more details.
      */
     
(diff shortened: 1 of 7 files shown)


Udumft:
List of accepted breaking changes:
-     💔 API breakage: enumelement DirectionsError.refreshExpired has been added as a new enum case
- PR adds handling of a new type of an error, I don't see a way how to avoid adding a new enum case, other than abusing `.unknown`.
-     💔 API breakage: constructor DirectionsError.init(code:message:response:underlyingError:) has been removed
- New argument added to this constructor with a default value -> won't require code changes on client's side
-     💔 API breakage: constructor RouteResponse.init(httpResponse:identifier:routes:waypoints:options:credentials:) has been removed
- New argument added to this constructor with a default value -> won't require code changes on client's side

Sources/MapboxDirections/DirectionsError.swift
```diff
@@ -105,6 +109,13 @@ public enum DirectionsError: LocalizedError {
      */
     case rateLimited(rateLimitInterval: TimeInterval?, rateLimit: UInt?, resetTime: Date?)
     
+    /**
+     Current route may no longer be refreshed.
```
kried, line 113:
```suggestion
     The current route may no longer be refreshed.
```

Sources/MapboxDirections/DirectionsError.swift
```diff
@@ -146,6 +157,8 @@ public enum DirectionsError: LocalizedError {
             #endif
             let formattedCount = NumberFormatter.localizedString(from: NSNumber(value: limit), number: .decimal)
             return "More than \(formattedCount) requests have been made with this access token within a period of \(formattedInterval)."
+        case .refreshExpired:
+            return "Current routes can no longer be refreshed due to it's TTL expiration."
```
kried, line 161:
```suggestion
            return "Current routes can no longer be refreshed due to their TTL expiration."
```

Tests/MapboxDirectionsTests/RouteResponseTests.swift
```diff
@@ -96,4 +98,18 @@ class RouteResponseTests: XCTestCase {
...
+    
+    func testRefreshTTL() {
+        let options = RouteOptions(coordinates: [])
+        let refreshTTL: TimeInterval = 60
+        let response = RouteResponse(httpResponse: nil,
+                                     options: .route(options),
+                                     credentials: BogusCredentials,
+                                     refreshTTL: refreshTTL)
+        
+        XCTAssertNotNil(response.refreshTTL)
```
kried, line 110: Why not check for equality?

Sources/MapboxDirections/DirectionsError.swift
```diff
@@ -8,7 +8,11 @@ import FoundationNetworking
  */
 public enum DirectionsError: LocalizedError {
     
-    public init(code: String?, message: String?, response: URLResponse?, underlyingError error: Error?) {
+    public init(code: String?, message: String?, response: URLResponse?, underlyingError error: Error?, refreshTTL: Int? = nil) {
+        guard refreshTTL == nil || refreshTTL != 0 else {
```
kried, line 12: Are we sure that in the case when `refreshTTL  == 0` and e.g. `response.statusCode == 404` the error should be `. refreshExpired`? 
As for me, it's better first to check the non-200 response code and only then check `refreshTTL`. 
Btw, how does the Android platform handle this case?

Udumft: `refresh_ttl` in the refresh response will be appended and =0 only in case of the refresh invalidation error. Android also does [the check](https://github.com/mapbox/mapbox-navigation-android/pull/7318/files#diff-0fd6e55eb977c007ca6298a2d0610352b533bd1166f2c0e6eb5571b7cb4cb8ecR164) for this parameter to be 0

kried (approved):

PR #829, opened 2024-07-29
Fix CI builds after CircleCI deprecated support for all Intel-based macOS resources

kried:
* Fixes CI builds
* Fixes failing DirectionsTests.swift

### DIFF
--- a/Tests/MapboxDirectionsTests/DirectionsTests.swift
+++ b/Tests/MapboxDirectionsTests/DirectionsTests.swift
@@ -56,7 +56,7 @@ class DirectionsTests: XCTestCase {
         XCTAssertEqual(directions.credentials, BogusCredentials)
     }
     
-    let maximumCoordinateCount = 794
+    let maximumCoordinateCount = 991
     
     func testGETRequest() {
         // Bumps right up against MaximumURLLength
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,9 @@
 version: 2.1
 
+orbs:
+  macos: circleci/macos@2.4.1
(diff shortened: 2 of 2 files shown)


Tests/MapboxDirectionsTests/DirectionsTests.swift
```diff
@@ -56,7 +56,7 @@ class DirectionsTests: XCTestCase {
         XCTAssertEqual(directions.credentials, BogusCredentials)
     }
     
-    let maximumCoordinateCount = 794
+    let maximumCoordinateCount = 991
```
Udumft, line 59: Did Directions increase this limit?

Udumft (approved):

Tests/MapboxDirectionsTests/DirectionsTests.swift
kried: No. The limit is actually was and is set in the SDK code, [MaximumURLLength](https://github.com/mapbox/mapbox-directions-swift/blob/6512320ee7f0091a4c55abe2bc6414f078da88c8/Sources/MapboxDirections/DirectionsOptions.swift#L11)

The diff is caused by the Xcode and iOS version update in tests. https://developer.apple.com/documentation/foundation/url/3126806-init
> Now, URL automatically percent- and IDNA-encodes invalid characters to help create a valid URL.

Previously, with iOS 16.4 `;` was presented in url as url encoded `%3B`, e.g. coordinates were represented as `0.0,0.0%3B0.0,0.0`. With iOS 17 now the url contains `0.0,0.0;0.0,0.0`

PR #831, opened 2025-02-05
Remove codecov support

kried:
* Adds `lcov` gem to generate html coverage report
* Adds test result reports to CircleCI, [which accepts only xml reports](https://circleci.com/docs/collect-test-data/#xcode). 

Here is an example of how the build looks like: [build](https://app.circleci.com/pipelines/gh/mapbox/mapbox-directions-swift/2176/workflows/e1debb86-e075-48d4-a8ad-5f8981a0ed00/jobs/10789/tests), the coverage can be open via `artifacts/coverage-report/index.html`


.circleci/config.yml
```diff
@@ -258,18 +263,14 @@ jobs:
...
+          command: xcodebuild -sdk iphonesimulator -project MapboxDirections.xcodeproj -scheme 'MapboxDirections iOS' -destination 'platform=iOS Simulator,OS=<< parameters.iOS >>,name=<< parameters.device >>' clean build <<# parameters.test >>test<</ parameters.test >><<# parameters.codecoverage >> -enableCodeCoverage "YES" | xcpretty --report junit --output test-results/junit.xml<</ parameters.codecoverage >>
       - when:
           condition: << parameters.codecoverage >>
           steps:
-            - publish-codecov
+            - publish-coverage
       - run:
           name: tvOS
           command: xcodebuild -project MapboxDirections.xcodeproj -scheme 'MapboxDirections tvOS' -destination 'platform=tvOS Simulator,name=Apple TV,OS=<< parameters.tvOS >>' clean build <<# parameters.test >>test <</ parameters.test >> <<# parameters.codecoverage >>-enableCodeCoverage YES<</ parameters.codecoverage >>
-      - when:
```
volkdmitri, line 269: Could you please explain this change?

README.md
```diff
@@ -4,7 +4,6 @@
 [![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage)
 [![CocoaPods](https://img.shields.io/cocoapods/v/MapboxDirections.svg)](https://cocoapods.org/pods/MapboxDirections/)
 [![SPM compatible](https://img.shields.io/badge/SPM-compatible-4BC51D.svg?style=flat)](https://swift.org/package-manager/)
-[![codecov](https://codecov.io/gh/mapbox/mapbox-directions-swift/branch/main/graph/badge.svg?token=bjr4ASuXuO)](https://codecov.io/gh/mapbox/mapbox-directions-swift)
```
volkdmitri, line 7: Don't we want to mention `lcov`?

kried: This section displayed labels on the top of the repo, to show the coverage. Unfortunately, `lcov` is just a local tool to build the report, not to provide the actual coverage status and coverage history.

Screenshot 2025-02-06 at 11 26 26

.circleci/config.yml
kried: With codecov, it was possible to report multiple coverage report per one job `build-job`. With the coverage display as artifact in CircleCI, only one test run will be reported, so I reported the coverage only for this iOS run.

Screenshot 2025-02-06 at 12 13 54

volkdmitri: But what about other platforms?

kried: Currently, coverage is collected only per one platform. Without this pull request, it will be collected per 0 platforms.

Currently, without codecov we don't report the coverage report to the third-party tool. I think it's convenient to see artifacts only per 1 platform because CircleCI artifacts are not designed to display lots of html. If you believe it's critical to collect MacOS and TV OS coverage as well, let's discuss our options.

volkdmitri: Does it make sense to disable code coverage for all other platforms (currently it is set to`-enableCodeCoverage YES`) to speed up build?

PR #834, opened 2025-07-04
Update version 2.14.1

kried:


volkdmitri (approved):
I believe the destination branch should be `NAVIOS-2276`.

PR #847, opened 2025-12-11
CI documentation workflow

maru-biz:
CI documentation workflow


.circleci/config.yml
```diff
@@ -28,6 +28,7 @@ commands:
 
     install-carthage:
         steps:
+            - macos/install-rosetta
```
kried, line 31: The same step was added to the `install-carthage` command.
Does it make sense to add a separate command to `install-rosetta`?

kried (approved):

PR #853, opened 2026-08-20
valid_indications parsing to allow multiple or no values

Udumft:
Refactored lane indication parsing to handle multiple preferred lanes or no preferred lanes.

### DIFF
--- a/Sources/MapboxDirections/Intersection.swift
+++ b/Sources/MapboxDirections/Intersection.swift
@@ -9,7 +9,111 @@ public struct Intersection: ForeignMemberContainer {
     public var lanesForeignMembers: [JSONObject] = []
     
     // MARK: Creating an Intersection
-    
+
+    /**
+     Initializes an intersection.
+
+     - parameter location: The geographic coordinates at the center of the intersection.
+     - parameter headings: The absolute headings of the roads that meet at the intersection.
+     - parameter approachIndex: The index of the item in `headings` that corresponds to the road used to approach the intersection.
+     - parameter outletIndex: The index of the item in `headings` that corresponds to the road used to leave the intersection.
+     - parameter outletIndexes: The indices of the items in `headings` that correspond to the roads that may be used to leave the intersection.
+     - parameter approachLanes: All the lanes of the road used to approach the intersection.
+     - parameter usableApproachLanes: The indices of the items in `approachLanes` that correspond to the lanes that may be used to execute the maneuver.
+     - parameter preferredApproachLanes: The indices of the items in `approachLanes` that correspond to the lanes that are preferred to execute the maneuver.
+     - parameter laneValidIndications: For each item in `approachLanes`, the indication that is applicable to the current route, or `nil` for a lane that has no applicable indication.
+     - parameter outletRoadClasses: The road classes of the road used to leave the intersection.
+     - parameter tollCollection: The toll collection point at the intersection.
+     - parameter tunnelName: The name of the tunnel that the intersection is a part of.
+     - parameter restStop: The rest stop at the intersection.
+     - parameter isUrban: Whether the intersection lies within the bounds of an urban zone.
+     - parameter regionCode: A 2-letter region code identifying the country that the intersection lies in.
+     - parameter outletMapboxStreetsRoadClass: The Mapbox Streets road class of the road used to leave the intersection.
+     - parameter railroadCrossing: Whether there is a railroad crossing at the intersection.
+     - parameter trafficSignal: Whether there is a traffic signal at the intersection.
+     - parameter stopSign: Whether there is a stop sign at the intersection.
+     - parameter yieldSign: Whether there is a yield sign at the intersection.
+     - parameter interchange: Information about routing and passing an interchange along the route.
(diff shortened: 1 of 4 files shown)


Sources/MapboxDirections/Lane.swift
```diff
@@ -59,7 +59,17 @@ extension Lane: Codable {
         indications = try container.decode(LaneIndication.self, forKey: .indications)
         isValid = try container.decode(Bool.self, forKey: .valid)
         isActive = try container.decodeIfPresent(Bool.self, forKey: .active)
-        validIndication = try container.decodeIfPresent(ManeuverDirection.self, forKey: .preferred)
+        // The Directions API may report an empty `valid_indication` for a lane that has no applicable
+        // indication, which is equivalent to omitting the field.
+        if let validIndicationString = try container.decodeIfPresent(String.self, forKey: .preferred),
+           !validIndicationString.isEmpty {
+            guard let validIndication = ManeuverDirection(rawValue: validIndicationString) else {
+                throw DecodingError.dataCorruptedError(forKey: .preferred, in: container, debugDescription: "Cannot initialize ManeuverDirection from invalid String value \(validIndicationString)")
```
maxkattner: Just wondering if an unknown indication string could ever happen intentionally in the future by some API change, and if falling back to nil in this case would be better than throwing an error?

Udumft: Good point. Though unlikely a new type of direction would be invented for guidance, but technically it would be correct to just ignore the unknown value rather than blocking the navigation route at all (per agreement with the nav team). I'll update this

Udumft: Updated!

Sources/MapboxDirections/Intersection.swift
```diff
@@ -175,12 +300,28 @@ public struct Intersection: ForeignMemberContainer {
...
     public let preferredApproachLanes: IndexSet?
     
+    /**
+     For each item in the `approachLanes` array, which of its `LaneIndication`s is applicable to the current route when there is more than one.
+
+     A lane’s entry is `nil` if the lane has no applicable indication. Lanes at the same intersection may have differing entries: for example, one lane may indicate a straight maneuver while an adjacent lane indicates a slight turn, if both lanes can be used to execute the maneuver.
+
+     If no lane information is available for the intersection, this property’s value is `nil`.
+     */
+    public let laneValidIndications: [ManeuverDirection?]?
```
kried, line 310: Should this type be double optional? Can it be `[ManeuverDirection]?`?

Sources/MapboxDirections/Lane.swift
```diff
@@ -59,7 +59,15 @@ extension Lane: Codable {
         indications = try container.decode(LaneIndication.self, forKey: .indications)
         isValid = try container.decode(Bool.self, forKey: .valid)
         isActive = try container.decodeIfPresent(Bool.self, forKey: .active)
-        validIndication = try container.decodeIfPresent(ManeuverDirection.self, forKey: .preferred)
+        // The Directions API may report an empty `valid_indication` for a lane that has no applicable
+        // indication, which is equivalent to omitting the field.
+        if let validIndicationString = try container.decodeIfPresent(String.self, forKey: .preferred),
```
kried, line 64: Should we add separate LaneTests?

Tests/MapboxDirectionsTests/IntersectionTests.swift
```diff
@@ -159,4 +157,277 @@ class IntersectionTests: XCTestCase {
...
+     “slight right”, must decode without throwing. This is spec-compliant data per the Directions API lane
+     object documentation, not malformed data:
+     https://docs.mapbox.com/api/navigation/directions/#lane-object
+     */
+    func testDifferingValidIndicationsDoNotThrow() {
+        let lanes = [
+            LaneJSON(indications: ["straight"], validIndication: "straight", isValid: true, isActive: false),
+            LaneJSON(indications: ["right", "slight right"], validIndication: "slight right", isValid: true, isActive: true),
+        ]
+        XCTAssertNoThrow(try decodeIntersectionWithLanes(lanes))
```
kried, line 218: Can we actually test the result values here? What laneValidIndications would they have?

kried (approved):

maxkattner (approved):

Sources/MapboxDirections/Lane.swift
Udumft: These lines are tested in scope of Intersection decoding tests, so this would be a duplication

Sources/MapboxDirections/Intersection.swift
Udumft: That is to distinguish missing lanes vs. empty lanes. It doesn't necessarily lead to a different business logic decisions on the SDK's side, but it's a more precise representation of the original JSON

Tests/MapboxDirectionsTests/IntersectionTests.swift
Udumft: The actual results are tested in the next one. Might be an excessive test, but it identifies the potential failure more precisely

